PR #8, opened 2024-04-10
add rpc to send cross chain payments via lnd

doitian:
This PR added two cross-chain RPC methods:

- `send_btc`: submit a BTC invoice and create a SendBTCOrder. CKB user can pay the cross-chain hub with CKB, and the hub will pay the target BTC invoice.
- `test_pay_btc`: simulate that the hub has received corresponding payment from CKB lightning channel.

What this PR lacks of:

- The logic to get preimage of the BTC payment and use it to settle the CKB lightning channel payment.


contrun:
Since PR #11 contains all the commits in this PR, can we close this PR in favor of #11 ?

PR #9, opened 2024-04-11
cch: settle ckb payment using accepted btc payment

doitian:
Ths PR depends on #8

PR #10, opened 2024-04-12
c2h: Add RPC to create order for receiving BTC

doitian:
Depends on #8 and #9


contrun:
Since PR #11 contains all the commits in this PR, can we close this PR in favor of #11 ?

PR #11, opened 2024-04-15
cch: subscribe lnd invoice to settle ReceiveBTCOrder

doitian:
Depends on #8, #9, #10


src/cch/command.rs
```diff
@@ -20,3 +26,24 @@ impl CchCommand {
 pub struct SendBTC {
     pub btc_pay_req: String,
 }
+
+/// Test fulfilling the SendBTC order.
+#[serde_as]
+#[derive(Clone, Debug, Deserialize)]
+pub struct TestPayBTC {
+    pub payment_hash: String,
```
contrun, line 34: Just a side note, we may use the more refined type Hash256 https://github.com/contrun/ckb-pcn-node/blob/b84257b36b97122b88becbf16151caa3449b3e52/src/ckb/types.rs#L82 , so that we don't have to validate whether the String payment_hash is indeed a valid hash. Since Hash256 is currently defined in another branch. This is ok for me. Just a side note.

src/cch/command.rs
```diff
@@ -20,3 +26,24 @@ impl CchCommand {
...
+}
+
+#[serde_as]
+#[derive(Clone, Debug, Deserialize)]
+pub struct ReceiveBTC {
+    /// Payment hash for the HTLC for both CKB and BTC.
+    pub payment_hash: String,
+
+    /// Identity of the payee CKB lightning node.
+    pub payee_pubkey: String,
```
contrun, line 44: Ditto.

src/cch/error.rs
```diff
@@ -2,8 +2,11 @@ use thiserror::Error;
 
 #[derive(Error, Debug)]
 pub enum CchDbError {
-    #[error("Duplicated SendBTCOrder with the same payment hash: {0}")]
-    DuplicatedSendBTCOrder(String),
+    #[error("Inserting duplicated key: {0}")]
```
contrun, line 5: "Inserting duplicated key" is perfectly fine from a developer's perspective. But it can be confusing for a user error message. Users may not know the context of "Inserting duplicated key" without looking at the source code. I suggest that we keep the original description "Duplicated SendBTCOrder with the same payment hash: {0}". I guess your intention was to make the description more general with respect to both sending and receiving BTC orders. Maybe we can just use "Duplicated BTC order with the same payment hash: {0}"

src/cch/error.rs
```diff
@@ -2,8 +2,11 @@ use thiserror::Error;
...
+    Duplicated(String),
+
+    #[error("Key not found: {0}")]
```
contrun, line 8: Ditto.

src/ckb/config.rs
```diff
@@ -48,7 +48,7 @@ impl CkbConfig {
 
     pub fn create_base_dir(&self) -> Result<()> {
         if !self.base_dir().exists() {
-            fs::create_dir_all(&self.base_dir()).map_err(Into::into)
```
contrun, line 51: Why don't we just use `fs::create_dir_all(self.base_dir())?`. To my understanding, `?` will automatically convert error types for us.

PR #16, opened 2024-04-29
Funding tx management (CKB Only)

doitian:
The sepc256k1 private key must be saved as a file using the path "ckb_chain/key". The file must encode the private key as hex without the "0x" prefix.

Example config:

```yaml
ckb_chain:
  # sighash_all
  funding_source_lock_script_code_hash: "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8"
  funding_source_lock_script_hash_type: "type"

  # funding lock
  funding_cell_lock_script_code_hash: "0x8090ce20be9976e2407511502acebf74ac1cfed10d7b35b7f33f56c9bd0daec6"
  funding_cell_lock_script_hash_type: "type"

services:
  - ckb_chain
```

### DIFF
--- a/src/ckb_chain/config.rs
+++ b/src/ckb_chain/config.rs
@@ -1,0 +1,126 @@
+use ckb_types::{packed, prelude::*};
+use clap_serde_derive::ClapSerde;
+use molecule::prelude::*;
+use secp256k1::SecretKey;
+use std::{
+    io::{ErrorKind, Read},
+    path::PathBuf,
+};
+
+use crate::ckb::types::Hash256;
+
+pub const DEFAULT_CKB_CHAIN_BASE_DIR_NAME: &str = "ckb-chain";
+const DEFAULT_CKB_CHAIN_NODE_RPC_URL: &str = "http://127.0.0.1:8114";
+
+#[derive(ClapSerde, Debug, Clone)]
+pub struct CkbChainConfig {
+    /// ckb base directory
+    #[arg(
+        name = "CKB_CHAIN_BASE_DIR",
+        long = "ckb-chain-base-dir",
+        env,
+        help = format!("base directory for ckb chain actor [default: $BASE_DIR/{}]", DEFAULT_CKB_CHAIN_BASE_DIR_NAME)
+    )]
+    pub base_dir: Option<PathBuf>,
+
+    #[default(DEFAULT_CKB_CHAIN_NODE_RPC_URL.to_string())]
+    #[arg(
+        name = "CKB_CHAIN_NODE_RPC_URL",
+        long = "ckb-chain-node-rpc-url",
+        env,
+        help = "rpc url to connect the ckb node [default: http://127.0.0.1:8114]"
+    )]
+    pub rpc_url: String,
+
+    #[arg(skip)]
+    pub funding_source_lock_script_code_hash: Hash256,
+
+    #[default(ckb_jsonrpc_types::ScriptHashType::Type)]
+    #[arg(skip)]
+    pub funding_source_lock_script_hash_type: ckb_jsonrpc_types::ScriptHashType,
+
+    #[arg(skip)]
+    pub funding_cell_lock_script_code_hash: Hash256,
+
+    #[default(ckb_jsonrpc_types::ScriptHashType::Type)]
+    #[arg(skip)]
+    pub funding_cell_lock_script_hash_type: ckb_jsonrpc_types::ScriptHashType,
+}
+
+impl CkbChainConfig {
+    pub fn base_dir(&self) -> &PathBuf {
+        self.base_dir.as_ref().expect("have set base dir")
+    }
+
+    pub fn create_base_dir(&self) -> crate::Result<()> {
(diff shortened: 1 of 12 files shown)


src/ckb_chain/config.rs
```diff
@@ -0,0 +1,154 @@
...
+                })
+            })
+            .map_err(Into::into)
+    }
+
+    pub fn build_funding_source_lock_script(&self, args: packed::Bytes) -> packed::Script {
+        packed::Script::new_builder()
+            .code_hash(self.funding_source_lock_script_code_hash.into())
+            .hash_type(packed::Byte::new(
+                Into::<ckb_types::core::ScriptHashType>::into(
```
contrun: Using `ckb_types::core::ScriptHashType::from` seems to be more natural.

src/ckb_chain/config.rs
```diff
@@ -0,0 +1,154 @@
...
+        packed::CellDep::new_builder()
+            .out_point(
+                packed::OutPoint::new_builder()
+                    .tx_hash(self.funding_source_lock_script_tx_hash.into())
+                    .index(self.funding_source_lock_script_tx_index.pack())
+                    .build(),
+            )
+            .dep_type(packed::Byte::new(
+                Into::<ckb_types::core::DepType>::into(
+                    self.funding_source_lock_script_dep_type.clone(),
```
contrun: Ditto

src/ckb_chain/config.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[arg(
+        name = "CKB_CHAIN_NODE_RPC_URL",
+        long = "ckb-chain-node-rpc-url",
+        env,
+        help = "rpc url to connect the ckb node [default: http://127.0.0.1:8114]"
+    )]
+    pub rpc_url: String,
+
+    #[arg(skip)]
+    pub funding_source_lock_script_tx_hash: Hash256,
```
contrun: Do you intend to create an new actor for every channel funding? User may want to open multiple channels to multiple users. It is my understanding that create an CkbChainActor to handle multiple channel opening would be better.

src/ckb_chain/funding/funding_tx.rs
```diff
@@ -0,0 +1,253 @@
...
+            .as_ref()
+            .map(|tx| !tx.outputs().is_empty())
+            .unwrap_or(false);
+
+        match self.request.udt_info {
+            Some(ref udt_info) => {
+                let mut udt_amount = self.request.local_amount as u128;
+                let mut ckb_amount = udt_info.local_ckb_amount;
+
+                // If outputs is not empty, assume that the remote party has already funded.
```
contrun: It would be better if we can put the comment here to line 110 where the variable `remote_funded` is defined. BTW, can we use something like `has_remote_funded_channel` (may be there is a better name) to make the meaning of this variable more clear?

src/ckb_chain/funding/funding_tx.rs
```diff
@@ -0,0 +1,253 @@
...
+            .set_outputs_data(outputs_data)
+            .build();
+
+        Ok(tx)
+    }
+}
+
+impl FundingTxBuilder {
+    fn build_funding_cell(&self) -> Result<(packed::CellOutput, packed::Bytes), FundingError> {
+        let remote_funded = self
```
contrun, line 113: I think at this stage, we can safely assume that both parties have already decided which live cells to use to fund the channel. This is because both parties must have gone through a lengthy transaction collaboration process and they have already sent a `tx_complete` message. So it seems to me `FundingTxBuilder` must have known all the inputs and outputs of this funding transaction. It is only the witness of the other party that is missing and we need to set our own witness. If my understanding is correct, then a lot of things in the `FundingTxBuilder` can be simplified.

src/ckb_chain/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod actor;
```
contrun, line 1: The functionality of the mod `ckb_chain` is indispensable for the `ckb` payment channel network. I think this mod would suit itself better as an sub module to `ckb`. How do you think of that? One benefit of using `ckb_chain` as a separate service is that in this case we can run a standalone daemon remotely that does nothing but signing funding transactions. I don't know if this is a real use case.

src/ckb_chain/config.rs
doitian: This is for cell dep

src/ckb_chain/funding/funding_tx.rs
doitian: FundingTxBuilder is responsoble to collect live cells during negotiation.

src/ckb_chain/funding/funding_tx.rs
doitian: Agree

src/ckb_chain/mod.rs
doitian: I have no strong opinion on this. However, I think we'd better refactor it into a submodule later after the demo.

PR #22, opened 2024-05-16
Add demo e2e

chenyukang:


src/ckb/channel.rs
```diff
@@ -317,11 +324,13 @@ impl ChannelActor {
                     }),
                 };
                 if let RemoveTlcReason::RemoveTlcFulfill(fulfill) = command.reason {
-                    if tlc.payment_hash != blake2b_256(fulfill.payment_preimage).into() {
+                    let filled_payment_preimage: Hash256 =
```
contrun: It would be better if we can rename `filled_payment_preimage` to `filled_payment_hash` as technically the result of a hash function is no longer a preimage.

src/ckb/network.rs
```diff
@@ -66,6 +66,11 @@ pub struct AcceptChannelResponse {
     pub new_channel_id: Hash256,
 }
 
+#[derive(Copy, Clone, Debug, Serialize)]
+pub enum ControlChannelResponse {
```
contrun: I think your intention here is to encompass all the responses to`ControlPcnChannel` messages to this enum? In my opinion, return a concrete type like `AcceptChannelResponse` is better here, i.e. instead of

```
    AddTlc(
        AddTlcCommand,
        #[serde(skip)] Option<RpcReplyPort<Result<ControlChannelResponse, RpcError>>>,
    ),
```

we can define

```
    AddTlc(
        AddTlcCommand,
        #[serde(skip)] Option<RpcReplyPort<Result<AddTlcResponse, RpcError>>>,
    ),
```

because the client can always expect a response to `AddTlc` to be a concrete response type which contains the tlc id. So making a enum for all responses seems to have no real benefits. If we are to use the same enum for all responses, then the client has to pattern match results for a specific response type. This seems strange to me.

tests/bruno/e2e/3-nodes-transfer/12-node1-add-tlc.bru
```diff
@@ -0,0 +1,50 @@
+meta {
```
contrun, line 1: There are two ways to process the invoice here. The first is to make the wallet parse the invoice and send `AddTlc` to the ckb-pcn-node, the second is let the ckb-pcn-node directly process invoice. Since we already have `NewInvoice` for ckb-pcn-node, maybe we should also make ckb-pcn-node process invoices from another node.

src/ckb/network.rs
chenyukang: Yes, my first version is encompassing all the responses to `ControlPcnChannel`, then I changed direction and forgot on this.

PR #25, opened 2024-05-24
fix: should stop and remove channel when peer disconnected

quake:


src/ckb/network.rs
```diff
@@ -499,7 +498,8 @@ pub struct NetworkActorState {
     // This immutable attribute is placed here because we need to create it in
     // the pre_start function.
     control: ServiceAsyncControl,
-    peers: HashMap<PeerId, HashSet<SessionId>>,
+    peer_session_map: HashMap<PeerId, SessionId>,
+    session_channels_map: HashMap<SessionId, HashSet<Hash256>>,
```
contrun, line 502: A hash map that maps peer id to channels to this map seems to be more sensible. What is the reason for using `SessionId` instead of `PeerId` here?

quake: SessionId is usize and PeerId is vec, just reduce some memory usage, no special reason.

PR #39, opened 2024-06-04
Add unit test to check the validitiy of commitment transaction

### LINKED ISSUE #26: Add unit test for commitment transaction
Currently we use the code https://github.com/contrun/ckb-pcn-node/blob/e62bf994c2b0c7abfc368f3662172f97ee12f17e/src/ckb/channel.rs#L2513-L2515 to generate a valid funding tx, and the code  https://github.com/contrun/ckb-pcn-node/blob/e62bf994c2b0c7abfc368f3662172f97ee12f17e/src/ckb/channel.rs#L2148-L2150 to create a revocation transaction from an old commitment transaction. We verify the validity of the resulting revocation transaction in https://github.com/contrun/ckb-pcn-node/blob/e62bf994c2b0c7abfc368f3662172f97ee12f17e/src/ckb/channel.rs#L2258-L2262 . These lines of code should be converted into a unit test. We need to first complete to implementation of mock ckb_chain actor https://github.com/contrun/ckb-pcn-node/blob/e62bf994c2b0c7abfc368f3662172f97ee12f17e/src/ckb_chain/actor.rs#L216-L242 so that it can respond to `Fund`, `Sign` and `TraceTx` requests in https://github.com/contrun/ckb-pcn-node/blob/e62bf994c2b0c7abfc368f3662172f97ee12f17e/src/ckb_chain/actor.rs#L30-L40 .

contrun:
This PR partly resolves https://github.com/contrun/ckb-pcn-node/issues/26. It tests that we can broadcast the commitment transaction to the network (i.e. commitment transaction is valid). We need also to create a test which consumes the comitment tx output (e.g.with revocation key).

The notable changes to this PR are:

1. A mock chain actor is created to verify transaction validity. Every transaction send to this actor via `SendTx` will be verified. If this transaction is invalid then the response to `TraceTx` request will return `Status::Rejected`. This is very useful to quickly check the validity of any transactions. In this PR, I used this infrastructure to test the commitment transaction can indeed consume the funding cell.
2. There is now a unit test to check the validity of the commitment transaction (in the sense that it can be posted to the network). As a side effect, the bug of not correctly setting witnesses to the commitment transaction is now fixed.
3. We now have a few more events that will be made visible to outside, including `CommitmentSignaturePending` which tells the owner it is his turn to sign a commitment transaction (this makes it quite easy to solve https://github.com/contrun/ckb-pcn-node/issues/27,. `RemoteCommitmentSigned` to signify a new commitment is signed by the remote.
4. Move the module chain in ckb to ckb_chain. ckb/chain was previously used to obtain contract (like commitment lock outpoint) information. It seems that only using ckb_chain to interact with the chain is more appropriate.


src/ckb_chain/actor.rs
```diff
@@ -214,29 +217,361 @@ impl CkbChainState {
...
+                    let outputs = match outputs.get(0) {
+                        Some(output) => {
+                            if output.lock() != request.script {
+                                error!(
+                                        "funding request script ({:?}) does not match the first output lock script ({:?})", request.script, output.lock()
+                                    );
+                                return Ok(());
+                            }
+                            let current_capacity: u64 = output.capacity().unpack();
+                            let capacity = request.local_amount + current_capacity;
```
chenyukang, line 300: emm, we need to care about UDT on this part.

chenyukang (approved):

src/ckb_chain/actor.rs
contrun: Yeah. We should also take care of UDT here, which would depend on your work. So, for now, we'd better leave this as a known issue.

quake (approved):

PR #46, opened 2024-06-07
simple cross-chain hub demo

doitian:
Specification: #85

Future Works: #88


tests/nodes/3/config.yml
```diff
@@ -5,19 +5,8 @@ rpc:
...
-ckb_chain:
-  # sighash_all
-  funding_source_lock_script_code_hash: "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8"
-  funding_source_lock_script_hash_type: "type"
-
-  # funding lock
-  funding_cell_lock_script_code_hash: "0x8090ce20be9976e2407511502acebf74ac1cfed10d7b35b7f33f56c9bd0daec6"
-  funding_cell_lock_script_hash_type: "type"
+  wrapped_btc_type_script_args: "0x32e555f3ff8e135cece1351a6a2971518392c1e30375c1e006ad0ce8eac07947"
+  lnd_cert_path: ../../../deploy/lnd-init/lnd-ingrid/tls.cert
```
chenyukang: why nodes/1 and nodes/2 config.yml don't need this?
now in main, `config.yml` will ignored by git, they will be generated from `deployer/config.yml`.

src/rpc/cch.rs
```diff
@@ -1,40 +1,234 @@
...
     pub btc_pay_req: String,
+    pub currency: Currency,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SendBTCResponse {
+    // Seconds since epoch when the order is created
+    #[serde_as(as = "U64Hex")]
+    pub timestamp: u64,
```
contrun, line 29: The names `created_at` and `expires_after` seem to more informative than `timestamp` and `expiry` for me. What do you say?

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+                    ChannelCommandWithId {
+                        channel_id: order.channel_id,
+                        command: ChannelCommand::AddTlc(
+                            AddTlcCommand {
+                                amount: order.amount_sats - order.fee_sats,
+                                preimage: None,
+                                payment_hash: Some(
+                                    Hash256::from_str(&order.payment_hash).expect("parse Hash256"),
+                                ),
+                                expiry: LockTime::new(self.config.ckb_final_tlc_expiry),
```
contrun: LockTime::new is actually expecting a parameter in the unit blocks instead of seconds. The root of this problem is that in the early days I thought TLC lock time is a relative number https://github.com/nervosnetwork/cfn-node/issues/29

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+            self.cert.as_deref(),
+            self.macaroon.as_deref(),
+        )
+        .await
+    }
+
+    async fn create_invoices_client(
+        &self,
+    ) -> Result<InvoicesClient, lnd_grpc_tonic_client::channel::Error> {
+        create_invoices_client(
```
contrun, line 119: It seems that we need to create a new gRPC client for each tlc settled. It would be better if we can somehow reuse the same connection. I don't know if tonic can automatically do that for us. This is not an urgent or important feature for now. But I think we can document that.

src/cch/config.rs
```diff
@@ -46,21 +67,21 @@ pub struct CchConfig {
...
 
     #[default(1)]
     #[arg(
-        name = "CCH_FEE_RATE_PER_MILLION_SHANNONS",
-        long = "cch-fee-rate-per-million-shannons",
+        name = "CCH_FEE_RATE_PER_MILLION_SATS",
+        long = "cch-fee-rate-per-million-sats",
         env,
-        help = "The proportional fee charged per million shannons based on the cross-chain order value, default is 1"
+        help = "The proportional fee charged per million satoshis based on the cross-chain order value, default is 1"
```
contrun, line 82: I think you meant to refer to millisatoshi (msat for short in bolts). `million` seems to be a typo for me, as `million` is 1,000,000 and milli is a 1/1000.

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+        &self,
+        myself: ActorRef<Self::Msg>,
+        message: Self::Msg,
+        state: &mut Self::State,
+    ) -> Result<(), ActorProcessingErr> {
+        match message {
+            CchMessage::SendBTC(send_btc, port) => {
+                let result = self.send_btc(state, send_btc).await;
+                if !port.is_closed() {
+                    port.send(result).expect("send reply");
```
contrun: There may be a TOCTOU bug here, i.e. when we are actually using the `port`, we may find out `port` is closed, but this `port` is not closed when we run `port.is_closed`. Since we didn't take any useful measures when the port is closed. Can we just change the code to

```
let _ = port.send(result);
```
? To be clear, we have already write this kind of code in current code base.

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+
+#[derive(Debug)]
+pub struct SettleReceiveBTCOrderEvent {
+    payment_hash: String,
+    preimage: Option<String>,
+    status: CchOrderStatus,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct SendBTC {
```
contrun, line 64: The name `SendBTC` is a little bit confusing to me. In the context of cch, this is not just a send operation. Instead, this is an exchange operation. We sent out BTC and are now expecting wBTC in return. Can we rename it to something like `ExchangeBTCforWBTC` (this name is debatable).

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+    pub btc_pay_req: String,
+    pub currency: Currency,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct ReceiveBTC {
```
contrun, line 70: Ditto.

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+    }
+}
+
+struct LndPaymentsTracker {
+    cch_actor: ActorRef<CchMessage>,
+    lnd_connection: LndConnectionInfo,
+    token: CancellationToken,
+}
+
+impl LndPaymentsTracker {
```
contrun, line 627: I think make this as an actor, spawn this actor from cch parent actor and make ractor manage the life cycle of `LndPaymentsTracker` will simplify below code a lot (especially the boilerplate of making `run` check for whether we should exit the running and letting `run_inner` take care of the business logic.

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+///
+/// <https://github.com/lightningnetwork/lnd/blob/07b6af41dbe2a5a1c85e5c46cc41019b64640d90/invoices/invoiceregistry.go#L292-L293>
+struct LndInvoiceTracker {
+    cch_actor: ActorRef<CchMessage>,
+    payment_hash: String,
+    lnd_connection: LndConnectionInfo,
+    token: CancellationToken,
+}
+
+impl LndInvoiceTracker {
```
contrun, line 730: Ditto.

src/cch/config.rs
```diff
@@ -72,13 +93,35 @@ pub struct CchConfig {
...
     #[arg(
-        name = "CCH_CKB_FINAL_TLC_EXPIRY",
-        long = "cch-ckb-final-tlc-expiry",
+        name = "CCH_CKB_FINAL_TLC_EXPIRY_BLOCKS",
+        long = "cch-ckb-final-tlc-expiry-blocks",
         env,
-        help = format!("final tlc expiry time in seconds for CKB network, default is {}", DEFAULT_CKB_FINAL_TLC_EXPIRY_TIME),
+        help = format!("final tlc expiry time in blocks for CKB network, default is {}", DEFAULT_CKB_FINAL_TLC_EXPIRY_BLOCKS),
     )]
     pub ckb_final_tlc_expiry: u64,
```
contrun: The unit (blocks) here is inconsistent with the unit of `ckb_final_tlc_expiry` in order.rs file (which uses seconds as unit)

src/cch/error.rs
```diff
@@ -1,21 +1,64 @@
...
+    #[error("ReceiveBTC received payment amount is too small")]
+    ReceiveBTCReceivedAmountTooSmall,
+    #[error("ReceiveBTC expected preimage but missing")]
+    ReceiveBTCMissingPreimage,
+    #[error("System time error: {0}")]
+    SystemTimeError(#[from] SystemTimeError),
+    #[error("JSON serialization error: {0}")]
+    JSONSerializationError(#[from] serde_json::Error),
+    #[error("Hex decoding error")]
+    HexDecodingError,
```
contrun: I personally always prefer to tell the user more context about the error. E.g. here we should tell the user which original string we are not able to decode. This is because some times it is not obvious as to which inputs caused the `HexDecodingError`. This may be costly as the original string to decode may be two large. What do you say?

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+        let req = routerrpc::SendPaymentRequest {
+            payment_request: order.btc_pay_req.clone(),
+            timeout_seconds: BTC_PAYMENT_TIMEOUT_SECONDS,
+            ..Default::default()
+        };
+        log::debug!("[inbounding tlc] SendPaymentRequest: {:?}", req);
+
+        let mut client = state.lnd_connection.create_router_client().await?;
+        // TODO: set a fee
+        let mut stream = client.send_payment_v2(req).await?.into_inner();
```
contrun, line 370: This will block the actor from processing other messages while issuing gRPC request. I think we can spawn a subtask and make this non-blocking. The problem is that after received gRPC response, we need to update the state. We can do that by add a message to the cch actor which can then process state update by handling this message.

src/cch/actor.rs
```diff
@@ -0,0 +1,818 @@
...
+
+        log::debug!("[settled tlc] preimage: {:#x}", preimage);
+
+        // settle the lnd invoice
+        let req = invoicesrpc::SettleInvoiceMsg {
+            preimage: preimage.as_ref().to_vec(),
+        };
+        log::debug!("[settled tlc] SettleInvoiceMsg: {:?}", req);
+
+        let mut client = state.lnd_connection.create_invoices_client().await?;
```
contrun, line 420: ditto.

tests/bruno/e2e/cross-chain-hub/07-node1-add-tlc.bru
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+assert {
+  res.body.error: isUndefined
+  res.body.result.tlc_id: isDefined
+}
+
+script:pre-request {
+  console.log("sending node1 AddTlc:", bru.getVar("payment_hash"));
```
contrun: I didn't find the definitions of `payment_hash` and `payment_amount`.

src/cch/orders_db.rs
```diff
@@ -1,18 +1,71 @@
...
 // TODO: persist orders
 #[derive(Default)]
 pub struct CchOrdersDb {
     /// SendBTCOrder map by payment hash
     send_btc_orders: HashMap<String, SendBTCOrder>,
+    receive_btc_orders: HashMap<String, ReceiveBTCOrder>,
 }
 
 impl CchOrdersDb {
     pub async fn insert_send_btc_order(&mut self, order: SendBTCOrder) -> Result<(), CchDbError> {
```
contrun, line 14: Since the key is just the payment hash here, I am wondering if an adversary can create a fake `AddTlc` order to nullify a valid send_btc request.

contrun (commented):
This PR is a little bit centered at BTC<->wBTC (an xDUT token with 1-1 pegging to BTC). I am wondering how easy it is to tweak code in this PR to support other assets. To be clear, we currently are only interested in BTC<->wBTC exchange. But the code can be make more generic.

src/cch/config.rs
doitian: `fee_rate_per_million_sats = 1` means user has to pay 1 sat for 1,000,000 sats order. See `send_btc` and `receive_btc` method in cch actor.

src/cch/actor.rs
doitian: This function is aync. Awaiting on a inner future will not block the actor.

doitian:
> This PR is a little bit centered at BTC<->wBTC (an xDUT token with 1-1 pegging to BTC). I am wondering how easy it is to tweak code in this PR to support other assets. To be clear, we currently are only interested in BTC<->wBTC exchange. But the code can be make more generic.

The code base before this PR supports CKB <-> BTC by configuring a ratio. A similar change can be applied to support other xUDT.

src/cch/orders_db.rs
doitian: Overwriting is not allowed. Real user can retry a new payment hash.

src/cch/actor.rs
contrun: Async function means that this code snippet will not block current thread, but this will still make the actor unable to process any other messages unrelated to this one, e.g. other SendBtc messages. This is because the function handle of this actor has still not returned, thus we have no chance to handle any other messages queued in the mailbox.

src/cch/error.rs
doitian: Resolved in https://github.com/nervosnetwork/cfn-node/pull/46/commits/a7736b33e702f6fed3b9144d4e62802a41a217f5

contrun:
While trying to demo the functionality of cross chain hub, I also added a few requests to close the channels in both lightning network and fiber network. Can you also add these to current PR? I think it would make the e2e test more complete. Here is my additional bruno files. https://github.com/nervosnetwork/cfn-node/tree/afbfc1e5a1c2dc2f900dae57f00ba1ac86f4927c/tests/bruno/e2e/cross-chain-hub

doitian:
> While trying to demo the functionality of cross chain hub, I also added a few requests to close the channels in both lightning network and fiber network. Can you also add these to current PR? I think it would make the e2e test more complete. Here is my additional bruno files. https://github.com/nervosnetwork/cfn-node/tree/afbfc1e5a1c2dc2f900dae57f00ba1ac86f4927c/tests/bruno/e2e/cross-chain-hub

You can send a PR to https://github.com/doitian/cfn-node/tree/feature/cch

doitian:
> > While trying to demo the functionality of cross chain hub, I also added a few requests to close the channels in both lightning network and fiber network. Can you also add these to current PR? I think it would make the e2e test more complete. Here is my additional bruno files. https://github.com/nervosnetwork/cfn-node/tree/afbfc1e5a1c2dc2f900dae57f00ba1ac86f4927c/tests/bruno/e2e/cross-chain-hub
> 
> You can send a PR to https://github.com/doitian/cfn-node/tree/feature/cch

https://github.com/doitian/cfn-node/pull/2

doitian:
Blocking, see https://github.com/nervosnetwork/cfn-node/pull/123

PR #57, opened 2024-06-10
Fix inputs of transactions that consume funding cell

contrun:


src/ckb/channel.rs
```diff
@@ -1854,22 +1861,21 @@ impl ChannelActorState {
             partial_signatures,
         )?;
 
-        let witness = self.create_witness_for_funding_cell(signature, None);
```
contrun, line 1857: The fix is essentially here and the lines below. Other code changes are to make it easier to debug.

quake: suggest to change the `version` parameter from `Option<u64>` to `u64` since we need to pass a non-empty version in all case.

src/ckb_chain/actor.rs
```diff
@@ -104,29 +104,37 @@ impl Actor for CkbChainActor {
...
-                                );
-                            }
-                            _ => {
-                                log::error!(
-                                    "[{}] send transaction failed: {:?}",
-                                    myself.get_name().unwrap_or_default(),
-                                    err
-                                );
+                    let result = match ckb_client.send_transaction(tx.data().into(), None) {
+                        Ok(_) => Ok(()),
```
chenyukang, line 112: Do we need to wait the shutdown transaction is confirmed?

contrun: I don't know. On the one hand, the fee may be so low that miners are never willing to pick up this closing transaction, thus this transaction may fail permanently. On the other hand, we have to wait indefinitely (due to network jitters etc) if we really want the transaction to be confirmed

chenyukang: If the transaction is not picked by low fee, any partner can not have a chance to retry it by raising fee if we closed the channel, since the transaction is signed by two partners?

contrun: I am not really familiar with the transaction fee bumping techniques used in ckb. But the closing transaction created here is a valid transaction. Thus I think methods like https://bitcoinops.org/en/topics/cpfp/ can create enough incentives  for miner. So my previous conclusion of this transaction may fail permanently may be incorrect.

PR #58, opened 2024-06-11
feat: auto accept channel

quake:


src/ckb/config.rs
```diff
@@ -46,6 +46,35 @@ pub struct CkbConfig {
     /// name of the network to use (can be any of `mocknet`/`mainnet`/`testnet`/`staging`/`dev`)
     #[arg(name = "CKB_NETWORK", long = "ckb-network", env)]
     pub network: Option<CkbNetwork>,
+
+    /// minimum ckb funding amount for open channel requests, unit: shannons [default: 16100000000 shannons]
+    #[arg(
+        name = "CKB_OPEN_CHANNEL_MIN_CKB_FUNDING_AMOUNT",
+        long = "ckb-open-channel-min-ckb-funding-amount",
+        env,
+        help = "minimum ckb funding amount for open channel requests, unit: shannons [default: 16100000000 shannons]"
```
contrun, line 55: Can we define a constant for the default value `16100000000` so that we can keep various references to this value in sync?

src/ckb/channel.rs
```diff
@@ -1061,6 +1061,11 @@ enum CommitmentSignedFlags {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(
```
contrun, line 1064: What is the purpose of this change?

src/store.rs
```diff
@@ -141,6 +141,26 @@ impl ChannelActorStateStore for Store {
         })
         .collect()
     }
+
+    fn get_channel_states(&self, peer_id: Option<PeerId>) -> Vec<(PeerId, Hash256, ChannelState)> {
+        let prefix = match peer_id {
+            Some(peer_id) => [&[64], peer_id.as_bytes()].concat(),
```
contrun, line 147: Can use define a constant for the prefix 64 and refer to that constant instead of directly using the literal 64?

contrun (approved):
Generally LTGM. Just some nits.

src/ckb/config.rs
quake: let's do it in another PR

src/store.rs
quake: let's do it in another PR

src/ckb/channel.rs
quake: > What is the purpose of this change?

to unify the json format of the output, for example:
```json
"state": {
  "state_name": "SIGNING_COMMITMENT",
  "state_flags": "OUR_COMMITMENT_SIGNED_SENT | THEIR_COMMITMENT_SIGNED_SENT"
}
```

PR #60, opened 2024-06-11
Find the correspending cell deps for script

chenyukang:


src/ckb_chain/contracts.rs
```diff
@@ -394,12 +392,35 @@ impl ContractsContext {
...
-    pub(crate) fn get_cell_deps(&self, _contracts: Vec<Contract>) -> CellDepVec {
-        match self {
+    pub(crate) fn get_cell_deps(&self, contracts: Vec<Contract>) -> CellDepVec {
+        let (script_cell_deps, contracts) = match self {
             #[cfg(test)]
-            Self::Mock(mock) => mock.contracts_context.cell_deps.clone(),
-            Self::Real(real) => real.cell_deps.clone(),
+            Self::Mock(mock) => {
+                //FIXME(yukang): why we need this
+                let mut contracts = contracts;
```
chenyukang, line 400: @contrun the mock actor need this?
without `CkbAuth` contract, an error will occur:

```console
[2024-06-11T16:35:47Z ERROR ckb_pcn_node::ckb_chain::actor::test_utils] submit tx failed: Other(Failed to verify transaction: TransactionView { data: Transaction(0xf70000000c000000f3000000e70000001c00000020000000490000004d0000007d000000db0000000000000001000000638e80b91a42f2f91d2c3d99f581dc925cd349b8c4a90b0e04239503014aca1f00000000000000000001000000000000000000000027ee683a2346c95a787e23ae253ffca1442602d066cefb943388240857901ace000000005e000000080000005600000010000000180000005600000064000000000000003e000000100000003000000031000000c904c21aeb7d475391f0b5c1bad70568d013071bef6ea143af9bfbded738388801090000007768617465766572320c000000080000000000000004000000), hash: Byte32(0x046211772acd7e3bbfa57bf0cf94fe91da5c303cd3c1a4a0272dab98936e466f), witness_hash: Byte32(0xfbee34611c421183f4a35da363f5ca86127833301f4637dcf38baa1bd76507eb) }, error: Error { kind: Script, inner: TransactionScriptError { source: Inputs[0].Lock, cause: ValidationFailure: see error code 1 on page https://nervosnetwork.github.io/ckb-script-error-codes/by-type-hash/7867c96e827775ef0057dff0b4398d2bf6939d4d0860552b5987e11296d60af5.html#1 } })
```

in test `ckb::channel::tests::test_create_channel`.

contrun: Yes. `ckb-testtool` will also need to set cell deps in order to find out which script binary to run. I think any dependencies on commitment lock and funding lock also require CkbAuth for "real" block chain environment.

src/ckb/channel.rs
```diff
@@ -2519,7 +2519,10 @@ impl ChannelActorState {
             }
         };
         let tx_builder = TransactionBuilder::default()
-            .cell_deps(get_cell_deps_by_contracts(vec![Contract::CommitmentLock]))
+            .cell_deps(get_cell_deps_by_contracts(vec![
+                Contract::CommitmentLock,
```
contrun: I made an error here previously. We actually need only the `FundingLock` because we are building a transaction to consume the funding cell. `CommitmentLock` is not required.

src/ckb/channel.rs
```diff
@@ -2568,8 +2571,9 @@ impl ChannelActorState {
     // so as to consume the funding cell.
     pub fn build_commitment_tx(&self, local: bool) -> (TransactionView, [u8; 32]) {
         let funding_out_point = self.get_funding_transaction_outpoint();
+        let contracts = vec![Contract::CommitmentLock, Contract::FundingLock];
```
contrun: Ditto.

src/ckb_chain/contracts.rs
```diff
@@ -394,12 +392,35 @@ impl ContractsContext {
...
+                (&mock.contracts_context.script_cell_deps, contracts)
+            }
+            Self::Real(real) => (&real.script_cell_deps, contracts),
+        };
+
+        let cell_deps_vec = contracts
+            .into_iter()
+            .map(|contract| {
+                script_cell_deps
+                    .get(&contract)
```
contrun, line 411: We can't simply find cell deps by lock up the hash map, as some contracts actually require a few more dependencies. E.g. funding lock and commitment lock both require ckb auth binary. Commitment lock may additionally require the sudt binary. We can create a dep group for each contracts. But that seems to be too much of a trouble for me.

src/ckb_chain/contracts.rs
chenyukang: I think we need to add CkbAuth in `deploy.sh` to initialize it?

src/ckb_chain/contracts.rs
contrun: Currently, in dev net, funding lock and commitment lock is already configured as a dep group. https://github.com/nervosnetwork/cfn-node/blob/main/tests/deploy/migrations/templates/funding-lock.toml . We need to make testnet and mainnet contracts as dep groups.

PR #68, opened 2024-06-13
Add min_fee_rate option commitment transaction

chenyukang:


chenyukang:
will be merged in https://github.com/nervosnetwork/cfn-node/pull/69

PR #72, opened 2024-06-14
Refactor tlc saving in channel state and fix a few bugs

contrun:
This PR refactored TLC fields in channel state. We will save all the tlcs in a unified tlcs field. This make it more convenient to see when a tlc is added/removed. When refactoring this I also found a few bugs. Below are the changes in this PR.

1. A notification for local commitment signed `LocalCommitmentSigned` is added. This may be used to notify watch tower that some transactions are now broadcastable by the peer.
2. We now have complete information for when a tlc is added/removed.
3. Previously commitment number is updated on `RevokeAndAck` from the peer received. This effectively makes local/remote commitment number always the same. We should instead update commitment numbers on `CommitmentSigned` messages sent or received. This PR fixes that.
4. The commitment number used to obtain remote commitment point should always be local commitment number (instead of remote commitment number). This asymmetry takes some time to understand, and previously we implemented in the wrong way. A remote commitment point is only obtained when `RevokeAndAck` is received from the remote party. And `RevokeAndAck` is sent only when the remote party receives a `CommitmentSigned` message. This is exactly when the local commitment transaction is updated.

### DIFF
--- a/src/ckb/channel.rs
+++ b/src/ckb/channel.rs
@@ -513,6 +513,10 @@ impl<S> ChannelActor<S> {
         match flags {
             CommitmentSignedFlags::SigningCommitment(flags) => {
                 let flags = flags | SigningCommitmentFlags::OUR_COMMITMENT_SIGNED_SENT;
+                // Normally commitment number will be incremented after received a RevokeAndAck message.
+                // But here channel has not been etablished yet, so we will not receive RevokeAndAck message.
+                // We increment the commitment number here instead.
+                state.increment_local_commitment_number();
                 state.update_state(ChannelState::SigningCommitment(flags));
                 state.maybe_transition_to_tx_signatures(flags, &self.network)?;
             }
@@ -524,7 +528,6 @@ impl<S> ChannelActor<S> {
                 state.maybe_transition_to_shutdown(&self.network)?;
             }
         }
-        state.increment_local_commitment_number();
         Ok(())
     }
 
@@ -1153,6 +1156,13 @@ impl CommitmentNumbers {
     pub fn increment_remote(&mut self) {
         self.remote += 1;
     }
+
+    pub fn flip(&self) -> Self {
+        Self {
+            local: self.remote,
+            remote: self.local,
+        }
+    }
 }
 
 #[derive(Copy, Clone, Debug, Serialize, Deserialize)]
@@ -1607,6 +1617,12 @@ impl ChannelActorState {
             self.total_amount = self.to_local_amount + self.to_remote_amount;
         }
 
+        if is_received {
+            self.increment_local_commitment_number();
+        } else {
+            self.increment_remote_commitment_number();
+        }
+
         // If this revoke_and_ack message is received from the counterparty,
         // then we should be operating on remote commitment numbers.
         let commitment_numbers = self.get_current_commitment_numbers();
@@ -1629,7 +1645,11 @@ impl ChannelActorState {
                     "Setting local_committed_at for tlc {:?} to commitment number {:?}",
                     tlc.tlc.id, commitment_numbers
                 );
-                tlc.creation_confirmed_at = Some(commitment_numbers);
+                if is_received {
+                    tlc.creation_confirmed_at = Some(commitment_numbers);
+                } else {
+                    tlc.creation_confirmed_at = Some(commitment_numbers);
+                };
                 // We offered a tlc, so we should decrease our balance.
                 if tlc.tlc.is_offered() {
                     to_local_amount -= amount;
@@ -1904,8 +1924,8 @@ impl ChannelActorState {
         let index = commitment_number as usize;
         let commitment_point = self.remote_commitment_points[index];
         debug!(
-            "Obtained remote commitment point #{} (counting from 1) of {}: {:?}",
-            index + 1,
+            "Obtained remote commitment point #{} (counting from 0) out of total {} commitment points: {:?}",
+            index,
             self.remote_commitment_points.len(),
             commitment_point
         );
@@ -1994,7 +2014,7 @@ impl ChannelActorState {
                         return am_i_sending_remove_tlc_message;
                     }
                     (_, Some(_), None) => {
-                        panic!("TLC {:?} is removed but not confirmed yet", tlc.id);
+                        panic!("TLC {:?} is removed but not confirmed yet", info);
                     }
                     (_, _, Some(n)) => {
                         debug!("Including TLC {:?} to commitment transaction because tlc confirmed at {:?}", tlc.id, n);
@@ -2041,6 +2061,10 @@ impl ChannelActorState {
             local: local_commitment_number,
             remote: remote_commitment_number,
         } = tlc.get_commitment_numbers(local);
+        debug!(
+            "Local commitment number: {}, remote commitment number: {}",
+            local_commitment_number, remote_commitment_number
+        );
         let local_pubkey = derive_tlc_pubkey(
             &self.get_local_channel_parameters().pubkeys.tlc_base_key,
             &self.get_local_commitment_point(remote_commitment_number),
@@ -2621,6 +2645,9 @@ impl ChannelActorState {
         self.remote_nonce = Some(commitment_signed.next_local_nonce);
         match flags {
             CommitmentSignedFlags::SigningCommitment(flags) => {
+                // Normally commitment number will be incremented after sent a RevokeAndAck message.
+                // But here channel has not been etablished yet, so we will not send RevokeAndAck message.
+                // We increment the commitment number here instead.
                 self.increment_remote_commitment_number();
                 let flags = flags | SigningCommitmentFlags::THEIR_COMMITMENT_SIGNED_SENT;
                 self.update_state(ChannelState::SigningCommitment(flags));
@@ -2630,7 +2657,9 @@ impl ChannelActorState {
                 // Now we should revoke previous transation by revealing preimage.
                 let old_number = self.get_remote_commitment_number();
                 let secret = self.signer.get_commitment_secret(old_number);
-                self.increment_remote_commitment_number();
+
+                self.update_state_on_raa_msg(false);
+
                 let new_number = self.get_remote_commitment_number();
                 let point = self.get_local_commitment_point(new_number);
 
@@ -2640,7 +2669,6 @@ impl ChannelActorState {
                 );
                 debug!("Sending new commitment point #{}: {:?}", new_number, &point);
 
-                self.update_state_on_raa_msg(true);
                 network
                     .send_message(NetworkActorMessage::new_command(
                         NetworkActorCommand::SendCFNMessage(CFNMessageWithPeerId {
@@ -2777,13 +2805,13 @@ impl ChannelActorState {
 
     pub fn append_remote_commitment_point(&mut self, commitment_point: Pubkey) {
         debug!(
-            "Appending remote commitment point #{}: {:?}",
-            self.remote_commitment_points.len() + 1,
+            "Setting remote commitment point #{} (counting from 0)): {:?}",
+            self.remote_commitment_points.len(),
             commitment_point
         );
-        debug_assert_eq!(
+        assert_eq!(
             self.remote_commitment_points.len() as u64,
-            self.get_current_commitment_number(true)
+            self.get_local_commitment_number()
         );
         self.remote_commitment_points.push(commitment_point);
     }
@@ -2797,7 +2825,7 @@ impl ChannelActorState {
             per_commitment_secret,
             next_per_commitment_point,
         } = revoke_and_ack;
-        let commitment_number = self.get_local_commitment_number() - 1;
+        let commitment_number = self.get_local_commitment_number();
         debug!(
             "Checking commitment secret and point for revocation #{}",
             commitment_number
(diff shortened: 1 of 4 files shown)


src/ckb/channel.rs
```diff
@@ -1590,34 +1645,183 @@ impl ChannelActorState {
...
+                        if removed_at == commitment_numbers && reason == remove_reason =>
+                    {
+                        debug!(
+                            "Skipping removing of tlc {:?} as it is already removed at {:?} with the same reason {:?}"
+                            , current, removed_at, reason
+                        );
+                        return Ok(current.clone());
+                    }
+                    Some((removed_at, remove_reason)) => {
+                        error!("Trying to remove an already removed tlc {:?} (removed at {:?}) with different reason: {:?} (old reason {:?})", current, removed_at, reason, remove_reason);
```
chenyukang: this branch seems we have `removed_at != commitment_numbers` or  `reason != remove_reason`,
but the error message implies only the different `reason`.

src/ckb/channel.rs
```diff
@@ -1590,34 +1645,183 @@ impl ChannelActorState {
...
+                let should_increase_our_balance = match reason {
+                    RemoveTlcReason::RemoveTlcFulfill(_) => tlc_id.is_received(),
+                    RemoveTlcReason::RemoveTlcFail(_) => tlc_id.is_offered(),
+                };
+                if should_increase_our_balance {
+                    self.to_local_amount += current.tlc.amount;
+                } else {
+                    self.to_remote_amount += current.tlc.amount;
+                }
+                current.removal_info = Some((commitment_numbers, reason));
```
chenyukang: is it better we remove the current entry from `self.tlcs`?
otherwise `tlcs` will keep growing, and we need to serialize and store this field.

src/ckb/channel.rs
```diff
@@ -1590,34 +1645,183 @@ impl ChannelActorState {
...
+            Entry::Vacant(_) => {
+                return Err(ProcessingChannelError::InvalidParameter(format!(
+                    "Trying to remove non-existing tlc with id {:?}",
+                    tlc_id
+                )));
+            }
+        }
+    }
+
+    pub fn get_channel_parameters(&self, local: bool) -> &ChannelParametersOneParty {
```
chenyukang: seems unused function here.

src/ckb/channel.rs
```diff
@@ -1590,34 +1645,183 @@ impl ChannelActorState {
...
+            self.to_remote_amount
+        );
+        let detailed_tlc = DetailedTLCInfo {
+            tlc,
+            committed_at: self.get_current_commitment_numbers(),
+            removal_info: None,
+        };
+        self.tlcs.insert(tlc.id, detailed_tlc);
+        if tlc.is_offered() {
+            self.to_local_amount -= tlc.amount;
```
chenyukang: can we add an `blance_check` function, which will make sure `to_local_amount + to_remote_amount + pending tlcs amount == initial channel amount`, so that we can catch error whenever unexpected thing happened.

src/ckb/channel.rs
```diff
@@ -1751,25 +1925,163 @@ impl ChannelActorState {
...
-            "Current pending offered tlcs: {:?}",
-            self.pending_offered_tlcs
+    pub fn get_active_received_tlcs(&self) -> impl Iterator<Item = &DetailedTLCInfo> {
+        if let ChannelState::ChannelReady(flags) = self.state {
+            if flags.contains(ChannelReadyFlags::AWAITING_REMOTE_REVOKE) {
+                // Remote has not confirmed our latest commitment transaction.
+                // We may have inconsistent view of the active tlcs.
+                // This should not happen.
+                panic!(
+                    "Trying to obtain list of active received tlcs while remote has not confirmed our latest commitment transaction, should have verified we are not in this state before calling this function.",
```
chenyukang: it's better to use multiple lines.

src/ckb/channel.rs
```diff
@@ -2751,8 +3055,12 @@ impl ChannelActorState {
...
     // The function returns a tuple, the first element is the commitment transaction itself,
     // and the second element is the message to be signed by the each party,
-    // so as to consume the funding cell.
-    pub fn build_commitment_tx(&self, local: bool) -> (TransactionView, [u8; 32]) {
+    // so as to consume the funding cell. The last element is the witnesses for the
+    // commitment transaction.
+    pub fn build_commitment_tx(&self, local: bool) -> (TransactionView, [u8; 32], Vec<u8>) {
+        let version = self.get_current_commitment_number(local);
+        debug!("Building {} commitment transaction #{} with local commtiment number {} and remote commitment number {}", if local { "local" } else { "remote" }, version,
+    self.get_local_commitment_number(), self.get_remote_commitment_number());
```
chenyukang: bad indentation.

src/ckb/channel.rs
contrun: We need this historical tlc information to construct valid revocation transaction. Imagine that the adversary broadcasts an older commitment transaction, and we want to revoke this. In order to do so, we will need to fill in all the tlcs that are committed commitment transaction in the witness. This is the reason why we need to keep tlcs forever. Of course, we can also save the witnesses for commitment transaction (just hook into this event https://github.com/nervosnetwork/cfn-node/blob/c1c4557a15d6457b23008328e1fc5cdf82e9611b/src/ckb/network.rs#L152C5-L152C27). I personally prefer the latter method. But I haven't think carefully about the tradeoffs. I think it is not such so important to optimize this yet.

src/ckb/channel.rs
contrun: Added in https://github.com/nervosnetwork/cfn-node/pull/72/commits/19d43491f9c5ccf660131c106060b9117adb09d3

src/ckb/channel.rs
```diff
@@ -1751,25 +1945,163 @@ impl ChannelActorState {
...
+        }
+        self.tlcs
+            .values()
+            .filter(|detailed_tlc| detailed_tlc.tlc.is_received())
+            .filter_map(|detailed_tlc| match detailed_tlc.removal_info {
+                // Remote has acknowledged the removal of this tlc.
+                Some(_) => None,
+                // This tlc is still active.
+                None => Some(detailed_tlc),
+            })
```
quake: suggest to add an `is_removed` fn to DetailedTlcInfo and change the code to:
```suggestion
            .filter(|detailed_tlc| !detailed_tlc.is_removed() && detailed_tlc.tlc.is_received() )
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1945,163 @@ impl ChannelActorState {
...
+            .values()
+            .filter(|detailed_tlc| detailed_tlc.tlc.is_offered())
+            .filter_map(|detailed_tlc| match detailed_tlc.removal_info {
+                // This tlc is removed only by the remote party initiating the removal.
+                // If the remote party has sent this removal message, we should consider
+                // it as inactive.
+                Some(_) => None,
+                // This tlc is still active.
+                None => Some(detailed_tlc),
+            })
```
quake: suggest to add an `is_removed` fn to DetailedTlcInfo and change the code to:
```suggestion
            .filter(|detailed_tlc| !detailed_tlc.is_removed() && detailed_tlc.tlc.is_offered() )
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1945,163 @@ impl ChannelActorState {
...
+    }
+
+    pub fn get_active_tlcs(&self) -> impl Iterator<Item = &DetailedTLCInfo> {
+        self.get_active_offered_tlcs()
+            .into_iter()
+            .chain(self.get_active_received_tlcs())
```
quake:
```suggestion
        self.tlcs
            .values()
            .filter(|detailed_tlc| !detailed_tlc.is_removed())
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1945,163 @@ impl ChannelActorState {
...
+    }
+
+    pub fn get_active_tlcs_value(&self) -> u128 {
+        self.get_active_tlcs()
+            .into_iter()
+            .map(|tlc| tlc.tlc.amount)
+            .sum::<u128>()
```
quake:
```suggestion
        self.get_active_tlcs()
            .map(|tlc| tlc.tlc.amount)
            .sum()
```

src/ckb/channel.rs
```diff
@@ -1202,6 +1128,144 @@ impl From<Transaction> for FundingTxInput {
...
+}
+
+impl Default for CommitmentNumbers {
+    fn default() -> Self {
+        Self::initial()
+    }
+}
+
+impl CommitmentNumbers {
+    pub fn new(local: u64, remote: u64) -> Self {
```
quake: I think this fn is useless because there are no scenarios need to set these 2 values directly.

src/ckb/channel.rs
```diff
@@ -1202,6 +1128,144 @@ impl From<Transaction> for FundingTxInput {
...
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord)]
+pub enum TLCId {
+    Offered(u64),
+    Received(u64),
+}
+
+// Required for using TLCId as key in BTreeMap.
```
quake: I don't think BTreeMap has the `Hash` trait bound requirement for the key, could you double check it?

src/ckb/channel.rs
```diff
@@ -1751,25 +1925,163 @@ impl ChannelActorState {
...
+            (local_pubkey, remote_pubkey)
+        } else {
+            (remote_pubkey, local_pubkey)
+        }
+    }
+
+    pub fn get_active_received_tlc_with_pubkeys(
+        &self,
+    ) -> impl Iterator<Item = (&DetailedTLCInfo, Pubkey, Pubkey)> {
+        self.get_active_received_tlcs().into_iter().map(move |tlc| {
```
quake: just use map:
```suggestion
        self.get_active_received_tlcs().map(|tlc| {
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1925,163 @@ impl ChannelActorState {
...
+            let (k1, k2) = self.get_tlc_pubkeys(tlc);
+            (tlc, k1, k2)
+        })
+    }
+
+    pub fn get_active_offered_tlc_with_pubkeys(
+        &self,
+    ) -> impl Iterator<Item = (&DetailedTLCInfo, Pubkey, Pubkey)> {
+        self.get_active_offered_tlcs().into_iter().map(move |tlc| {
```
quake: just use map:
```suggestion
        self.get_active_offered_tlcs().map(|tlc| {
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1925,163 @@ impl ChannelActorState {
...
+        let tlcs = {
+            let (mut received_tlcs, mut offered_tlcs) = (
+                self.get_active_received_tlc_with_pubkeys()
+                    .into_iter()
+                    .map(|(tlc, local, remote)| (tlc.clone(), local, remote))
+                    .collect::<Vec<_>>(),
+                self.get_active_offered_tlc_with_pubkeys()
+                    .into_iter()
+                    .map(|(tlc, local, remote)| (tlc.clone(), local, remote))
+                    .collect::<Vec<_>>(),
```
quake:
```suggestion
                self.get_active_received_tlc_with_pubkeys()
                    .map(|(tlc, local, remote)| (tlc.clone(), local, remote))
                    .collect::<Vec<_>>(),
                self.get_active_offered_tlc_with_pubkeys()
                    .map(|(tlc, local, remote)| (tlc.clone(), local, remote))
                    .collect::<Vec<_>>(),
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1925,163 @@ impl ChannelActorState {
...
+            );
+            debug!("Received tlcs: {:?}", &received_tlcs);
+            debug!("Offered tlcs: {:?}", &offered_tlcs);
+            let (mut a, mut b) = if local {
+                (received_tlcs, offered_tlcs)
+            } else {
+                for tlc in received_tlcs.iter_mut().chain(offered_tlcs.iter_mut()) {
+                    // Need to flip these fields for the counterparty.
+                    tlc.0.tlc.flip_mut();
+                }
```
quake:
```suggestion
                for (tlc, _, _) in received_tlcs.iter_mut().chain(offered_tlcs.iter_mut()) {
                    // Need to flip these fields for the counterparty.
                    tlc.tlc.flip_mut();
                }
```

src/ckb/channel.rs
```diff
@@ -1751,25 +1925,163 @@ impl ChannelActorState {
...
+                (offered_tlcs, received_tlcs)
+            };
+            a.sort_by(|x, y| u64::from(x.0.tlc.id).cmp(&u64::from(y.0.tlc.id)));
+            b.sort_by(|x, y| u64::from(x.0.tlc.id).cmp(&u64::from(y.0.tlc.id)));
```
quake, line 2194: BTreeMap is sorted, I don't think we need sort again here. btw, we need unit test for this fn.

src/ckb/channel.rs
quake: @chenyukang a reminder,  we need to check `to_local_amount / to_remote_amount - tlc.amount > min_amount` here before inserting.

quake (changes requested):

src/ckb/channel.rs
contrun: Yes. You're right. I removed it.

PR #91, opened 2024-06-20
 Fix using the wrong commitment point for revocation

contrun:
There is currently a bug can be reproduced by repeatedly add tlcs from one party. It will trigger index out of bound panic. This is because we are using the wrong commitment number to obtain remote commitment point (which we have not kept in an array). This PR add the repro to the e2e test. Admittedly this is not enough, we need a "real" test to verify the created commitment transaction is valid (e.g. can be revoked when a revocation secret is revealed).

### DIFF
--- a/src/ckb/channel.rs
+++ b/src/ckb/channel.rs
@@ -3243,50 +3243,60 @@ impl ChannelActorState {
     }
 
     fn build_current_commitment_transaction_witnesses(&self, local: bool) -> Vec<u8> {
-        let commitment_number = self.get_current_commitment_number(local);
+        let local_commitment_number = self.get_local_commitment_number();
+        let remote_commitment_number = self.get_remote_commitment_number();
+        let commitment_number = if local {
+            local_commitment_number
+        } else {
+            remote_commitment_number
+        };
         debug!(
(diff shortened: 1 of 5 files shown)


src/ckb/channel.rs
```diff
@@ -3243,50 +3243,60 @@ impl ChannelActorState {
...
+                // The two fields below are used to derive a pubkey for the revocation.
+                // Unlike delayed_payment_commitment_point above, the revocation key is held
+                // by the counter-signatory until this commitment transaction is revoked.
+                revocation_commitment_point,
+                revocation_base_key,
+            ) = if local {
                 (
                     self.get_local_channel_parameters().selected_contest_delay,
-                    self.get_local_commitment_point(commitment_number),
+                    self.get_local_commitment_point(remote_commitment_number),
```
chenyukang, line 3276: it's a little bit werid `get_local_commitment_point` receive a `remote_*` parameter, and `get_remote_commitment_point` receive a `local_*` paramter.

contrun: I agree this is weird. There are currently a few more places with this kind of weird things in the code base. This is a known problem. I will clarify what it means to be local and clean up these things in a following PR.

PR #96, opened 2024-06-20
Derive secrets from commitment points

### LINKED ISSUE #63: Actually derive new secrets for each commitment transaction
Currently we didn't actually use the per commitment point to derive new secret keys.

https://github.com/contrun/ckb-pcn-node/blob/7de2f8ce969da5ca6cdc7a59c8ca16ac296b8db6/src/ckb/channel.rs#L3191-L3195

Besides the obvious security implication, it may make some bugs go unnoticed. For example, one bug in https://github.com/contrun/ckb-pcn-node/pull/62 should be discovered much earlier if we are actually using different keys for different commitment transaction.

> The second bug is that we are using the wrong per commitment point to construct commitment transaction. We should use the current commitment number (and obtain the current commitment point) to construct commitment transaction. Previously, we used previous commitment number. This bug goes undiscovered because we didn't actually use commitment point to derive new public keys. We just copy the old keys. Thus commitment transaction are still the same for both parties.

contrun:
Fixes #63

### DIFF
--- a/src/ckb/channel.rs
+++ b/src/ckb/channel.rs
@@ -1509,10 +1509,10 @@ impl ChannelActorState {
         second_commitment_point: Pubkey,
     ) -> Self {
         let signer = InMemorySigner::generate_from_seed(seed);
-        let local_pubkeys = signer.to_channel_public_keys(INITIAL_COMMITMENT_NUMBER);
+        let local_base_pubkeys = signer.get_base_public_keys();
 
         let channel_id = derive_channel_id_from_revocation_keys(
-            &local_pubkeys.revocation_base_key,
+            &local_base_pubkeys.revocation_base_key,
             &remote_pubkeys.revocation_base_key,
         );
 
@@ -1534,7 +1534,7 @@ impl ChannelActorState {
             to_remote_amount: remote_value,
             local_shutdown_script: None,
             local_channel_parameters: ChannelParametersOneParty {
-                pubkeys: local_pubkeys,
+                pubkeys: local_base_pubkeys,
                 selected_contest_delay: remote_delay,
             },
             signer,
@@ -1566,7 +1566,7 @@ impl ChannelActorState {
         to_local_delay: LockTime,
     ) -> Self {
         let signer = InMemorySigner::generate_from_seed(seed);
-        let local_pubkeys = signer.to_channel_public_keys(INITIAL_COMMITMENT_NUMBER);
+        let local_pubkeys = signer.get_base_public_keys();
         let temp_channel_id =
             derive_temp_channel_id_from_revocation_key(&local_pubkeys.revocation_base_key);
         Self {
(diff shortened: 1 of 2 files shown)


src/ckb/types.rs
```diff
@@ -285,6 +297,14 @@ impl Privkey {
     pub fn pubkey(&self) -> Pubkey {
         Pubkey::from(self.0.public_key(secp256k1_instance()))
     }
+
+    pub fn tweak<I: Into<[u8; 32]>>(&self, scalar: I) -> Self {
```
doitian, line 301: Since you only need ref of `[u8; 32]`, `AsRef` is better than `Into` here.

src/ckb/channel.rs
```diff
@@ -504,10 +509,6 @@ impl<S> ChannelActor<S> {
         match flags {
             CommitmentSignedFlags::SigningCommitment(flags) => {
                 let flags = flags | SigningCommitmentFlags::OUR_COMMITMENT_SIGNED_SENT;
-                // Normally commitment number will be incremented after received a RevokeAndAck message.
-                // But here channel has not been etablished yet, so we will not receive RevokeAndAck message.
-                // We increment the commitment number here instead.
-                state.increment_local_commitment_number();
```
doitian, line 510: Could you explain why this is no longer needed?

src/ckb/channel.rs
```diff
@@ -2698,14 +2705,14 @@ impl ChannelActorState {
...
+            self.get_remote_nonce(),
+            &commitment_signed.next_local_nonce
+        );
         self.remote_nonce = Some(commitment_signed.next_local_nonce);
         match flags {
             CommitmentSignedFlags::SigningCommitment(flags) => {
-                // Normally commitment number will be incremented after sent a RevokeAndAck message.
-                // But here channel has not been etablished yet, so we will not send RevokeAndAck message.
-                // We increment the commitment number here instead.
-                self.increment_remote_commitment_number();
```
doitian, line 2719: Could you explain why this is no longer needed?

doitian (commented):
Skimmed

src/ckb/channel.rs
contrun: This commit has relevant information about this change. https://github.com/nervosnetwork/cfn-node/pull/96/commits/11f710bafc809af779441daf1eb73b1b603e185c. TLDR is that I moved the code to `on_channel_ready` function. It has to be delayed to there because incrementing these numbers too soon may cause problems.

src/ckb/channel.rs
contrun: Ditto

src/ckb/types.rs
contrun: The problem I encountered is that https://doc.rust-lang.org/std/convert/trait.AsRef.html#tymethod.as_ref `AsRef<[u8; 32]>` is not automatically implemented for `[u8; 32]`. The automatic implementation of the rust compiler only works without the length 32. https://doc.rust-lang.org/std/convert/trait.AsRef.html#impl-AsRef%3C%5BT%5D%3E-for-%5BT%5D

PR #103, opened 2024-06-26
Add auto shutdown when shutdown tx fee is ok

chenyukang:


src/ckb_chain/actor.rs
```diff
@@ -363,7 +363,9 @@ mod test_utils {
                         request, &tx, &fulfilled_tx
                     );
 
-                    if let Err(e) = reply_port.send(Ok(fulfilled_tx)) {
+                    let dummy_lock = request.script.clone();
```
quake: suggesting rename `dummy_lock` to `funding_source_lock_script`

quake (approved):

PR #105, opened 2024-06-27
feat: allow sha256 for HTLC

doitian:
Allow use sha256 in TLC.

See contract PR https://github.com/nervosnetwork/cfn-scripts/pull/6

- Added a field `hash_algorithm` when adding TLC
- Save the `hash_algorithm` in the internal TLC struct
- Check hash against preimage using the saved `hash_algorithm`
- Added `hash_algorithm` in the invoice as a new attribute

The sha256 TLC is covered in the e2e test case 3-nodes-transfer with extra args `--env-var HASH_ALGORITHM=sha256`


doitian:
Need time to add tests to cover SHA256 HTLC, but the code is ready for review.

src/ckb/hash_algorithm.rs
```diff
@@ -0,0 +1,79 @@
+use bitcoin::hashes::{sha256::Hash as Sha256, Hash as _};
+use ckb_hash::blake2b_256;
+use ckb_types::packed;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+#[repr(u8)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
```
contrun, line 9: Side note: there are many other places where we want to deserialize a string into a enum. Some of they are case insensitive (e.g. https://github.com/nervosnetwork/cfn-node/blob/c69f8121f9a542ddeae4d38e467234024ec92c8f/src/ckb_chain/config.rs#L108-L114 ), some of they are case sensitive, some of they just use the default serde config (e.g. https://github.com/nervosnetwork/cfn-node/blob/c69f8121f9a542ddeae4d38e467234024ec92c8f/src/ckb/config.rs#L127-L129). This can create confusion. We may want to unify all these serialization/deserialization procedures.

quake (approved):

contrun (commented):
Generally LGTM. We may also need to update the binaries in the folder `tests/deploy/contracts` so that we can use the latest contracts. There are some tests in this repo that directly run these contract to see if the generated transactions are valid.

PR #107, opened 2024-06-27
chore: remove unused struct and tweak log (too many logs)

quake:


src/ckb/network.rs
```diff
@@ -1431,36 +1392,31 @@ impl ServiceProtocol for Handle {
...
-            );
+        match context.session.remote_pubkey.as_ref() {
+            Some(pubkey) => {
+                let peer_id = PeerId::from_public_key(pubkey);
+                self.send_actor_message(NetworkActorMessage::new_event(
+                    NetworkActorEvent::PeerDisconnected(peer_id, context.session.clone()),
+                ));
+            }
+            None => {
+                unreachable!("Received message without remote pubkey");
```
contrun, line 1415: AFAIK, executing the code `unreachable!` will result in an undefined behavior. I don't know if it is possible for the adversary to fake a tentacle connection without providing any remote pubkey. If this is a legal operation for tentacle, then this will make the host panic.

src/ckb/network.rs
```diff
@@ -1431,36 +1392,31 @@ impl ServiceProtocol for Handle {
...
-            );
+        match context.session.remote_pubkey.as_ref() {
+            Some(pubkey) => {
+                let peer_id = PeerId::from_public_key(pubkey);
+                self.send_actor_message(NetworkActorMessage::new_event(
+                    NetworkActorEvent::PeerMessage(peer_id, msg),
+                ));
+            }
+            None => {
+                unreachable!("Received message without remote pubkey");
```
contrun, line 1430: Ditto.

src/ckb/channel.rs
```diff
@@ -3706,24 +3689,14 @@ impl ChannelActorState {
         // for xudt compatibility issue,
         // refer to: https://github.com/nervosnetwork/cfn-scripts/pull/5
         let empty_witness_args: [u8; 16] = [16, 0, 0, 0, 16, 0, 0, 0, 16, 0, 0, 0, 16, 0, 0, 0];
-        let witnesses: Vec<u8> = [
```
contrun, line 3709: The content of witness is quite useful while debugging. Can we add a `trace!` here?

src/ckb/network.rs
```diff
@@ -424,10 +419,6 @@ where
                 }
             }
             NetworkActorEvent::ChannelReady(channel_id, peer_id) => {
-                info!(
```
contrun: I added these messages as `info!` because I thought it would be both useful to developer and users. Users sometimes want to learn that some channel is actually closed on their command.

src/ckb/network.rs
quake: a connected session always have the remote pubkey, it's guarded by tentacle

src/ckb/channel.rs
quake: there is a debug level log in the fn build_commitment_transaction_parameters already, it's redundant

```rust
        let witnesses: Vec<u8> = self.build_current_commitment_transaction_witnesses(local);

        let hash = blake2b_256(&witnesses);
        let script_arg: &[u8] = &hash[..20];
        debug!(
            "Building {} commitment transaction with witnesses {:?} and hash {:?} with local commitment number {} and remote commitment number {}",
            if local { "local" } else {"remote"},
            hex::encode(&witnesses),

```

src/ckb/network.rs
contrun: Then, this is fine for me.

src/ckb/network.rs
quake: reverted, please review again, thanks.

chenyukang:
Suggest only print tx_hash at here (two peers will send the same tx content)
https://github.com/nervosnetwork/cfn-node/blob/e7e210f061a1191305a78ec8f4e1d3550e63f187/src/ckb/network.rs#L468-L472


since we will also print the whole shutdown tx at:
https://github.com/nervosnetwork/cfn-node/blob/e7e210f061a1191305a78ec8f4e1d3550e63f187/src/ckb/network.rs#L457

contrun (approved):

quake:
> Suggest only print tx_hash at here (two peers will send the same tx content)
> 
> https://github.com/nervosnetwork/cfn-node/blob/e7e210f061a1191305a78ec8f4e1d3550e63f187/src/ckb/network.rs#L468-L472
> 
> since we will also print the whole shutdown tx at:
> 
> https://github.com/nervosnetwork/cfn-node/blob/e7e210f061a1191305a78ec8f4e1d3550e63f187/src/ckb/network.rs#L457

resolved in #108

PR #114, opened 2024-06-28
Enable clippy check

### LINKED ISSUE #110: Clippy should not fail in CI
Add `#[allow]` to ignore the errors in generated files.

chenyukang:
Fixes #110


Makefile
```diff
@@ -0,0 +1,15 @@
+CLIPPY_OPTIONS := -- -D warnings -D clippy::redundant_clone -D clippy::enum_glob_use \
```
contrun: Using a config file is more preferable https://doc.rust-lang.org/stable/clippy/configuration.html.

src/ckb/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::all)]
 pub mod config;
```
contrun: Is this added unintentionally? This file is not automatically generated.

src/ckb/schema/gen.sh
```diff
@@ -12,4 +12,9 @@ for file in "${files[@]}"; do
     f="$schema_dir/$file"
     output_file="$gen_dir/$(basename "${f%%.mol}.rs")"
     "$MOLC" --language rust --schema-file "$f" | rustfmt > "$output_file"
+
+    ## ignore them in clippy
+    DIRECTIVE="#![allow(clippy::all)]"
+    echo -e "$DIRECTIVE\n$(cat "$output_file")" > "$output_file"
```
contrun: Reading and writing the same file in a single shell pipeline may cause problems like incomplete read. It is normally recommended to use sponge https://man.archlinux.org/man/sponge.1 . But here we can just replace the cat command with the above molc command.

src/ckb/channel.rs
```diff
@@ -1873,6 +1867,8 @@ impl ChannelActorState {
                     "Setting local_committed_at for tlc {:?} to commitment number {:?}",
                     tlc.tlc.id, commitment_numbers
                 );
+                // FIXME: need to confirm with @contrun
+                #[allow(clippy::if_same_then_else)]
                 if is_received {
                     tlc.creation_confirmed_at = Some(commitment_numbers);
                 } else {
```
chenyukang: @contrun please confirm this, is this unintentionally?

src/ckb/schema/gen.sh
chenyukang: `sponge` is not a default tool on macos, fixed it with `mktemp`.

src/ckb/channel.rs
contrun: The is_received check should be deleted.

contrun (approved):

src/ckb/schema/gen.sh
quake: please add `#[allow(clippy::all)]` to the gen/mod.rs thus we can keep the generated file unchanged.
```
#[allow(clippy::all)]
pub mod cfn;
```

src/ckb/gen/cfn.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::all)]
```
doitian: https://github.com/nervosnetwork/cfn-node/blob/d1cdb4f2464beab34e962e6b1b8aeac85b40d74b/src/ckb/gen/mod.rs#L2 does not work?

doitian (approved):

src/ckb/gen/cfn.rs
chenyukang: yes, it works to add another `#[allow(clippy::all)]`.

PR #116, opened 2024-06-29
Add created_at datetime for channel

### LINKED ISSUE #90: Channels returned from `list_channels` RPC should have timestamp and can be sorted by time
We should persist channel creation time as it is quite useful to users. Another good-to-have feature is to sort returned channels by creation time by default. This is also useful for developers. Some of our tests (e.g. https://github.com/nervosnetwork/cfn-node/blob/main/tests/bruno/e2e/open-use-close-a-channel/03-get-auto-accepted-channel.bru ) use the first result of `list_channels` as the channel id of the newly created channel. This can create problems when we have many local channels, because the order of returned channel can be somewhat arbitrary.

chenyukang:
Fixes #90

### DIFF
--- a/src/rpc/channel.rs
+++ b/src/rpc/channel.rs
@@ -10,6 +10,7 @@ use crate::ckb::{
     NetworkActorCommand, NetworkActorMessage,
 };
 use crate::{handle_actor_call, handle_actor_cast, log_and_error};
+use chrono::{DateTime, Utc};
 use ckb_jsonrpc_types::Script;
 use ckb_types::core::FeeRate;
 use jsonrpsee::{
@@ -84,6 +85,7 @@ pub struct Channel {
     pub sent_tlc_balance: u128,
     pub remote_balance: u128,
     pub received_tlc_balance: u128,
+    pub created_at: DateTime<Utc>,
 }
 
 #[serde_as]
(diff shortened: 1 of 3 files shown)


src/rpc/channel.rs
```diff
@@ -235,25 +237,28 @@ where
...
+                        remote_balance: state.get_remote_balance(),
+                        sent_tlc_balance: state.get_sent_tlc_balance(),
+                        received_tlc_balance: state.get_received_tlc_balance(),
+                        created_at: state.created_at,
+                    })
+            })
+            .collect();
+        // Sort by created_at in descending order
+        channels.sort_by(|a, b| b.created_at.cmp(&a.created_at));
+        Ok(ListChannelsResult { channels })
```
contrun, line 267: Can we also update https://github.com/nervosnetwork/cfn-node/blob/1297afde388981ef14c1e498cf34e4ad1baef7d0/src/rpc/channel.rs#L78 to include the `created_at` field?

chenyukang: it's done in the PR?

contrun: Oh, sorry. I overlooked that. Can you also give an example of the output format of the DateTime<Utc>? Since this will be shown to the users. I think we need to make this unambiguous and easily readable. I don't know if the default serde rule is good enough.

chenyukang: looks like this:
```console
created_at: '2024-06-29T15:14:44.765077Z'
```

contrun: So, this is ISO 8601. I think we'd better make this serialization rule more explicit lest upstream changes break down stream applications. Some documentation for the date time format will also be useful to downstream application developers. Since we don't have rpc documentation yet. We can just add a Todo in the code.

quake: > looks like this:
> 
> ```
> created_at: '2024-06-29T15:14:44.765077Z'
> ```

please follow ckb's rpc convention, use the epoch timestamp in rpc field and displayed as hex string: https://github.com/nervosnetwork/ckb/tree/develop/rpc#type-timestamp

src/rpc/channel.rs
```diff
@@ -235,25 +237,28 @@ where
...
+                        local_balance: state.get_local_balance(),
+                        remote_balance: state.get_remote_balance(),
+                        sent_tlc_balance: state.get_sent_tlc_balance(),
+                        received_tlc_balance: state.get_received_tlc_balance(),
+                        created_at: state.created_at,
+                    })
+            })
+            .collect();
+        // Sort by created_at in descending order
+        channels.sort_by(|a, b| b.created_at.cmp(&a.created_at));
```
quake:
```suggestion
        channels.sort_by_key(|channel| Reverse(channel.created_at));
```

doitian (approved):

PR #120, opened 2024-07-01
Add node log prefix for logs

chenyukang:
I heavily rely on the node log prefix, there is no option to config this for tracing because of `info_span` does not work for tasks spawned by actor, this change works for tracing.


src/cch/service.rs
```diff
@@ -37,23 +37,23 @@ impl CchService {
...
                         }
                         Some(command) => {
                             let command_name = command.name();
-                            tracing::info!("Process cch command {}", command_name);
+                            crate::info!("Process cch command {}", command_name);
 
                             match self.process_command(command).await {
                                 Ok(_) => {}
                                 Err(err) => {
-                                    tracing::error!("Error processing command {}: {:?}", command_name, err);
```
contrun, line 56: Since tracing support structured logging, we don't have to set format specifier in the first string passed to the macro.

```
crate::error!("Error processing command", command_name, err);
```

seems to be more elegant for me.

src/cch/service.rs
```diff
@@ -37,23 +37,23 @@ impl CchService {
...
                     match command {
                         None => {
-                            tracing::debug!("Command receiver completed, shutting down tentacle service");
+                            crate::debug!("Command receiver completed, shutting down tentacle service");
                             break;
                         }
                         Some(command) => {
                             let command_name = command.name();
-                            tracing::info!("Process cch command {}", command_name);
+                            crate::info!("Process cch command {}", command_name);
```
contrun, line 51: Just use `crate::info!("Process cch command", command_name);`

src/cch/service.rs
```diff
@@ -64,7 +64,7 @@ impl CchService {
     }
 
     async fn process_command(&mut self, command: CchCommand) -> Result<()> {
-        tracing::debug!("CchCommand received: {:?}", command);
+        crate::debug!("CchCommand received: {:?}", command);
```
contrun, line 67: ditto

src/cch/service.rs
```diff
@@ -74,7 +74,7 @@ impl CchService {
         let duration_since_epoch = SystemTime::now().duration_since(UNIX_EPOCH)?;
 
         let invoice = Bolt11Invoice::from_str(&send_btc.btc_pay_req)?;
-        tracing::debug!("BTC invoice: {:?}", invoice);
+        crate::debug!("BTC invoice: {:?}", invoice);
```
contrun, line 77: ditto

contrun (approved):
Just a side note. We should use structured logging (instead of setting the format specifier in the first argument) in the future.

PR #121, opened 2024-07-01
Add commitment tx fee

### LINKED ISSUE #87: Take transaction fee into consideration while building commitment transaction
https://github.com/nervosnetwork/cfn-node/blob/f9e79b060f7329309eff752b7485bfacc91a629f/src/ckb/channel.rs#L3315

Currently, while building commitment transaction outputs, we didn't consider transaction fee. This will effectively make commitment transaction unbroadcastable.

chenyukang:
Fixes #87

### DIFF
--- a/src/ckb/channel.rs
+++ b/src/ckb/channel.rs
@@ -30,7 +30,7 @@ use std::{borrow::Borrow, collections::BTreeMap};
 use crate::{
     ckb::{
         config::{DEFAULT_UDT_MINIMAL_CKB_AMOUNT, MIN_OCCUPIED_CAPACITY},
-        fee::calculate_commitment_tx_fee,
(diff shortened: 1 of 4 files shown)


src/ckb/channel.rs
```diff
@@ -3827,6 +3884,9 @@ impl ChannelActorState {
                 )
             };
 
+            // FIXME(yukang): how do we make sure time_locked_ckb_amount >= commitment_tx_fee ?
+            time_locked_ckb_amount -= commitment_tx_fee;
```
chenyukang, line 3853: TODO: seems we need to add a check to make sure `reserved ckb amount >= commitment_tx_fee`, but when we're opening a channel, which side is the `to_local`? @quake

quake (approved):

PR #123, opened 2024-07-04
Lock bruno/cli verison

contrun:
The hack in https://github.com/nervosnetwork/cfn-node/blob/f740c4b54570162dd509f98ef38c53b895725732/tests/bruno/e2e/open-use-close-a-channel/06-generate-a-few-blocks.bru no longer works. We will have a

> Too many jumps, possible infinite loop

error in the newest version (1.20.1) of bruno. For now, we just lock bruno to an older version. 1.20.0 is the newest working version.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -68,7 +68,7 @@ jobs:
             fi
           done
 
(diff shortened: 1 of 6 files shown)


quake:
we may use [generate_epochs](https://github.com/nervosnetwork/ckb/tree/develop/rpc#integration_test-generate_epochs)

PR #130, opened 2024-07-10
Don't reply shutdown message automatically when it is sent

contrun:


src/ckb/channel.rs
```diff
@@ -387,7 +387,11 @@ impl<S: ChannelActorStateStore> ChannelActor<S> {
 
                 let mut flags = flags | ShuttingDownFlags::THEIR_SHUTDOWN_SENT;
 
-                if state.check_valid_to_auto_accept_shutdown() {
+                // Don't auto accept shutdown message if we have already sent our shutdown message.
+                // This may create problem when the peer and us send shutdown message at the same time.
+                if state.check_valid_to_auto_accept_shutdown()
+                    && !flags.contains(ShuttingDownFlags::OUR_SHUTDOWN_SENT)
```
chenyukang: prefer moving this flag check to `check_valid_to_auto_accept_shutdown`.

And we'd better check `matches!(state.status, ChannelState::ShuttingDown(_))` then not to auto accept shutdown?

contrun: > prefer moving this flag check to check_valid_to_auto_accept_shutdown.

Agreed.

> And we'd better check matches!(state.status, ChannelState::ShuttingDown(_)) then not to auto accept shutdown?

Do you mean only check `matches!(state.status, ChannelState::ShuttingDown(_))` instead of `matches!(state.status, ChannelState::ShuttingDown(flags) if flags.contains(ShuttingDownFlags::OUR_SHUTDOWN_SENT))`? Why?

chenyukang: why not, state will be in `ShuttingDown` when:
1. handled shutdown command
2. handled shutdown peer message
we don't want to go to auto accept in any these two cases?

contrun: In case 2, it is possible that they send their `Shutdown` message, and we have not successfully handle it and then we crashed. We didn't successfully send our shutdown message. So we need to send this message again.

chenyukang: I just reconsidered it, with the force shutdown, there is a new shutdown flags: `WAITING_COMMITMENT_CONFIRMATION`, if we have started force shutdown, we should also don't want to reply auto shutdown.

seems we can only reply auto shutdown when the flags is: `THEIR_SHUTDOWN_SENT`, means the other side is waiting reply from me.

contrun: Just reconsidered point 1. I think we should still have two functions. Because the ShutdownFlags here are obtained from the code above. We my have not update the state yet. If we are checking state in `check_valid_to_auto_accept_shutdown`. Then we will have to update the state earlier. It can be done this way, but it is not so ideal to me.

chenyukang: Ok, let's don't move the ckeck into `check_valid_to_auto_accept_shutdown`, the current change seems great.

PR #135, opened 2024-07-12
Remove funding source lock script

chenyukang:
Construct lock script with `pubkeys.funding_pubkey`

### DIFF
--- a/src/ckb/channel.rs
+++ b/src/ckb/channel.rs
@@ -150,7 +150,6 @@ pub const DEFAULT_TO_LOCAL_DELAY_BLOCKS: u64 = 10;
 #[derive(Debug)]
 pub struct TxUpdateCommand {
     pub transaction: Transaction,
-    pub funding_source_lock_script: Script,
 }
 
 pub struct OpenChannelParameter {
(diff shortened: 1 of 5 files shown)


src/ckb/channel.rs
```diff
@@ -2441,6 +2433,18 @@ impl ChannelActorState {
         KeyAggContext::new(keys).expect("Valid pubkeys")
     }
 
+    pub fn get_local_funding_script(&self) -> Script {
```
contrun: Can we rename it to `get_default_local_funding_script`? To some unfamiliar eyes, the name `get_local_funding_script` implies that we will get the intended funding script.

src/ckb/fee.rs
```diff
@@ -94,18 +94,14 @@ pub(crate) fn calculate_commitment_tx_fee(fee_rate: u64, udt_type_script: &Optio
...
-                .clone()
-                .expect("no funding source lock script"),
-        ),
-        state.local_shutdown_script.clone().unwrap_or(
-            state
-                .funding_source_lock_script
-                .clone()
-                .expect("no funding source lock script"),
-        ),
+        state
```
contrun, line 97: Can we make a function `get_remote_funding_script` (if above function is renamed, see comment above) or `get_remote_funding_script_or_default` or this?

PR #151, opened 2024-09-09
feat: refactor commitment lock to using daric protocol

quake:
ref: https://github.com/nervosnetwork/fiber-scripts/pull/10


src/fiber/channel.rs
```diff
@@ -3259,15 +3324,12 @@ impl ChannelActorState {
             }
         };
 
-        let tx = self.verify_and_complete_tx(commitment_signed.partial_signature)?;
+        let tx = self.verify_and_complete_tx(
+            commitment_signed.funding_tx_partial_signature,
+            commitment_signed.commitment_tx_partial_signature,
+        )?;
         // This is the commitment transaction that both parties signed,
         // can be broadcasted to the network if necessary
```
chenyukang, line 3332: can we keep this assertion?

```rust
assert_eq!(
            tx.data().serialized_size_in_block(),
            commitment_tx_size(&self.funding_udt_type_script, None)
        );
```

I remember it help me avoid a bug before.

src/fiber/channel.rs
```diff
@@ -3930,156 +4017,52 @@ impl ChannelActorState {
     // and the second element is the message to be signed by the each party,
     // so as to consume the funding cell. The last element is the witnesses for the
```
chenyukang, line 4018: The code comments for this function need to be updated.

src/fiber/channel.rs
```diff
@@ -4088,182 +4071,231 @@ impl ChannelActorState {
...
+        let message = blake2b_256(
+            [
+                to_local_output.as_slice(),
+                to_local_output_data.as_slice(),
+                to_remote_output.as_slice(),
+                to_remote_output_data.as_slice(),
+                &args[0..36],
+            ]
+            .concat(),
         );
```
chenyukang, line 4277: this code logic seems duplicated with the code in `build_and_verify_commitment_tx`, a helper function `message_from_settlement_tx(..)` may be help.

src/fiber/channel.rs
```diff
@@ -5580,35 +5574,32 @@ mod tests {
             Status::Committed
         );
 
-        dbg!(hex::encode(privkey.as_ref()));
-        dbg!(&commitment_tx);
+        println!("commitment_tx: {:?}", commitment_tx);
```
chenyukang, line 5577: clean up or use info!?

chenyukang (approved):

PR #152, opened 2024-09-11
Add network graph, broadcasting, route path and multi-hop payment

### LINKED ISSUE #149: auto close tx should use original pubkey as lock script
Currently, we're using drived pubkey for lock script when auto close:

https://github.com/nervosnetwork/fiber/blob/0e09af24726bb3d0031b5a9ac26bcb7dc375c2b9/src/fiber/channel.rs#L2526

we should use the original node pubkey, which will be passed into ChannelActorState in future change:

https://github.com/nervosnetwork/fiber/commit/517443b25f237feb215f474fc42c463227ed37a5#diff-69ae8d1e23722c7deb83f0097b5e72b6120e824443741bf08c3c6b8e902979f4L221-R238

feature: 
- User may also specify the `close_script` when open channel.

chenyukang:
This PR add the MVP implementation of multi-hop payment.

## broadcasting
- `DEFAULT_ANNOUNCE_NODE_INTERVAL_SECONDS` controls the interval to reannounce NodeAnnouncement, and `DEFAULT_AUTO_ANNOUNCE_NODE` controls whether a node will announce when starting up.

## network graph
- use rocksdb as storage layer
- network actor and channel actor will broadcasting proper message to add nodes or edges in graph

## Routing algorithm
- A Dijkstra variant from tagert to source, with several considerations
  - Fee and timeLockDelta, converted to weight
  - Then combined with the estimated success rate to convert to the distance dist of the evaluation function
  - Feerate
- Each hop uses its own fee rate to calculate the forwarding fee. This parameter can be specified when OpenChannel. If not specified, the default `inbounding_tlc_fee_proportional_millionths` is used.

## Multi-hop payment
- Now we use a temporary `OnionPacket` for forwarding payment in multiple hops, it is a mock one and will be replaced after the our own onion packet implementation finished.
- The last hop will automatically `remove_tlc` if the proper invoice preimage is found

Also fixes #149


src/fiber/graph.rs
```diff
@@ -0,0 +1,1174 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct ChannelInfo {
+    pub funding_tx_block_number: u64,
+    pub funding_tx_index: u32,
+    pub announcement_msg: ChannelAnnouncement,
+    // The last update (with the timestamp) for the channel received from the network.
+    pub one_to_two: Option<(ChannelUpdateInfo, u64)>,
+    pub two_to_one: Option<(ChannelUpdateInfo, u64)>,
```
chenyukang: It's better in readability to add a timestamp field in `ChannelUpdateInfo`?

chenyukang: @contrun

contrun: Yes. I added that.

PR #159, opened 2024-09-18
Add keysend mode in send_payment

chenyukang:
[How to allow the recipient to receive payments without generating an invoice](https://discord.com/channels/674518700518408213/1105706051602501683/1284442027030745119). In real-world scenarios, there is a similar need where the sender wants to initiate a payment to the recipient.

The keysend mode has been added for this purpose. For more information, see:

- https://docs.lightning.engineering/lightning-network-tools/lnd/send-messages-with-keysend
- https://docs.lightning.engineering/lightning-network-tools/lnd/amp

- In keysend mode, we randomly generate a preimage, then use it to generate a payment hash, and embed the preimage into the last hop of onion routing. This ensures that only the recipient can see the preimage and can use it to directly remove the TLC when settling the TLC.


quake (approved):

quake:
please note that this feature requires `Blinded Paths`, this PR is only to test the feasibility of the process, there are security and privacy concerns, waiting for https://github.com/cryptape/fiber-sphinx integration.

PR #182, opened 2024-09-23
Add network graph PRC

chenyukang:
Add two rpc for query nodes and channels, currently only support simple pagination.
Exact query mode maybe needed in future.

### DIFF
--- a/src/fiber/graph.rs
+++ b/src/fiber/graph.rs
@@ -76,14 +76,10 @@ impl ChannelInfo {
     }
 
     pub fn channel_last_update_time(&self) -> Option<u64> {
-        let node1_to_node2_time = self.node1_to_node2.as_ref().map(|x| x.timestamp);
-        let node2_to_node1_time = self.node2_to_node1.as_ref().map(|x| x.timestamp);
-        match (node1_to_node2_time, node2_to_node1_time) {
-            (Some(time1), Some(time2)) => Some(time1.max(time2)),
-            (Some(time1), None) => Some(time1),
-            (None, Some(time2)) => Some(time2),
-            (None, None) => None,
-        }
+        self.node1_to_node2
+            .as_ref()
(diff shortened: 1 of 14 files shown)


src/fiber/graph.rs
```diff
@@ -74,6 +75,17 @@ impl ChannelInfo {
...
 
+    pub fn channel_last_update_time(&self) -> Option<u64> {
+        let node1_to_node2_time = self.node1_to_node2.as_ref().map(|x| x.timestamp);
+        let node2_to_node1_time = self.node2_to_node1.as_ref().map(|x| x.timestamp);
+        match (node1_to_node2_time, node2_to_node1_time) {
+            (Some(time1), Some(time2)) => Some(time1.max(time2)),
+            (Some(time1), None) => Some(time1),
+            (None, Some(time2)) => Some(time2),
+            (None, None) => None,
+        }
```
quake:
```suggestion
    self.node1_to_node2.as_ref().map(|n| n.timestamp).max(self.node2_to_node1.as_ref().map(|n| n.timestamp))
```

src/rpc/graph.rs
```diff
@@ -0,0 +1,165 @@
...
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Clone)]
+pub struct ChannelInfo {
+    #[serde_as(as = "EntityHex")]
+    pub channel_outpoint: OutPoint,
+    #[serde_as(as = "U64Hex")]
+    pub funding_tx_block_number: u64,
+    #[serde_as(as = "U32Hex")]
+    pub funidng_tx_index: u32,
```
quake: typo
```suggestion
    pub funding_tx_index: u32,
```

quake (changes requested):

quake (approved):

PR #188, opened 2024-09-24
refactor: last_update_message should not be none

quake:


contrun (approved):
```
    /// Mostly redundant with the data we store in fields explicitly.
    /// Everything else is useful only for sending out for initial routing sync.
    /// Not stored if contains excess data to prevent DoS.
    pub last_update_message: ChannelUpdate,
```

The comment here shows that we copied some code from rust-lightning.

https://github.com/lightningdevkit/rust-lightning/blob/a0d0f02ed06f667cab29950c5019b1bb830b98f0/lightning/src/ln/msgs.rs#L1248-L1251

```
	/// Excess data which was signed as a part of the message which we do not (yet) understand how
	/// to decode.
	///
	/// This is stored to ensure forward-compatibility as new fields are added to the lightning gossip protocol.
	pub excess_data: Vec<u8>,
```
I didn't see bolt 07 mentioned the excess data here. I think this may be a feature of lightning network that any later field of the message are appended to the wire message. So the author of rust-lightning tried to make this message future-proof by saving the unknown fields to excess data. Since we have no notion of excess data. I am in favor of removing this. But we also need to update the comment.

PR #191, opened 2024-09-24
Payment session and error handling on onion packet forwarding

chenyukang:
This PR contains these changes:
1. Add payment session and payment status, and `get_payment` rpc will query the payment status.
2. Add error handling on sending onion packet and TLC forwarding (without onion encryption right now)
3. Some code clean up and refactor


doitian (approved):

src/fiber/channel.rs
```diff
@@ -677,16 +675,69 @@ where
...
+            }
+            ProcessingChannelError::TlcAmountIsTooLow => TlcErrorCode::AmountBelowMinimum,
+            ProcessingChannelError::TlcNumberExceedLimit
+            | ProcessingChannelError::TlcValueInflightExceedLimit => {
+                TlcErrorCode::TemporaryChannelFailure
+            }
+            ProcessingChannelError::InvalidState(_) => match state.state {
+                ChannelState::Closed(_) => TlcErrorCode::PermanentChannelFailure,
+                // ShuttingDown is a temporary state
+                ChannelState::ShuttingDown(_) => TlcErrorCode::TemporaryChannelFailure,
```
contrun: Why is this a temporary failure?

src/fiber/channel.rs
```diff
@@ -458,88 +465,69 @@ where
...
-                        peeled_packet_bytes = Some(peeled_packet.serialize());
+                        Ok(())
+                    }
+                    Err(e) => {
+                        // we assume that TLC was not inserted into our state,
+                        // so we can safely send RemoveTlc message to the peer
+                        // note we can not use get_received_tlc_by_id here, because this new add_tlc may be
+                        // trying to add a duplicate tlc, so we use tlc count to make sure no new tlc was added
+                        // and only send RemoveTlc message to peer if the TLC is not in our state
+                        assert!(tlc_count == state.tlcs.len());
```
contrun: I don't fully understand the comment above yet. But, if the assertion fail, there is not to much useful information for debugging here.

src/fiber/channel.rs
```diff
@@ -702,6 +753,158 @@ where
...
+                add_tlc.amount, forward_amount
+            );
+            if forward_to_next_hop {
+                assert!(received_amount >= forward_amount);
+                let forward_fee = received_amount.saturating_sub(forward_amount);
+                let fee_rate: u128 = state
+                    .public_channel_info
+                    .as_ref()
+                    .map(|info| info.tlc_fee_proportional_millionths.unwrap_or_default())
+                    .unwrap_or_default();
```
contrun, line 841: Getting a default fee value here effectively makes a non-public channel forwarding TLCs? Is my assumption correct? If it is, is this kind of behavior desirable?

src/fiber/channel.rs
```diff
@@ -702,6 +753,158 @@ where
...
+    ) -> Result<(TLCId, Option<Vec<u8>>), ProcessingChannelError> {
+        state.check_for_tlc_update(Some(add_tlc.amount))?;
+
+        // check the onion_packet is valid or not, if not, we should return an error.
+        // If there is a next hop, we should send the AddTlc message to the next hop.
+        // If this is the last hop, we should check the payment hash and amount and then
+        // try to fulfill the payment, find the corresponding payment preimage from payment hash.
+        let mut preimage = None;
+        let mut peeled_packet_bytes: Option<Vec<u8>> = None;
+        let mut forward_to_next_hop = false;
```
contrun: I personally prefer to use immutable variables. I think we can just assign the values in different branches below (instead of initializing and mutating).

src/fiber/graph.rs
```diff
@@ -732,7 +732,7 @@ where
                         continue;
                     }
                 }
-                // info!("weight: {:?} dist: {:?} fee: {:?}", weight, dist, fee);
+                // info!("weight: {:?} dist: {:?} fee: {:?}", weight, distance, fee);
```
contrun: Just delete it or use `debug` or `trace`?

src/fiber/channel.rs
chenyukang: I removed `forward_to_next_hop`, `preimage` and `peeled_packet_bytes` need to returned at the end of this function, seems don't using `mut` make the code harder.

src/fiber/channel.rs
chenyukang: I move the error log line to the previous line position of the assertion.

src/fiber/channel.rs
chenyukang: if this node broadcasting a fee rate 100, then a update on fee rate is done, but the latest fee rate haven't been received by the graph before a payment is sent. 

now there are two scenarios:
1. the new fee is lower than previous one, but we don't report error here [maybe need to discussed here]
2. the new fee is higher than previous one, we need to report error here.

if there is no configuration for the fee(means the node removed fee rate), I think it's same with scenario #1.

How can we find out a route with this node if this node is non-public? At least our current code will not do this. maybe we should add a check that only public node will forward a TLC right now.

src/fiber/channel.rs
chenyukang: I'm not sure about this, can `ShuttingDown` revert back to `Ready`?

chenyukang: Seems can not convert back right now? If ti's true, maybe we can change it to a Permanent error.

contrun: Yeah. I don't think we can revert the shutdown. It should be a permanent error.

src/fiber/channel.rs
contrun: The `public_channel_info` is set while creating the channel, we deem every channel with this information as public channel https://github.com/contrun/cfn-node/blob/4033caaae18933ed922fa05543b6d686407628c0/src/fiber/channel.rs#L2224 .

chenyukang: I resolve this issue with https://github.com/nervosnetwork/fiber/pull/191/commits/9fc0a35bccf95427a82337cb793b4ad84732f9ea

src/fiber/network.rs
```diff
@@ -2061,30 +2083,273 @@ where
...
+        match self.store.get_payment_session(*payment_hash) {
+            Some(payment_session) => Ok(payment_session.into()),
+            None => Err(Error::InvalidParameter(format!(
+                "Payment session not found: {:?}",
+                payment_hash
+            ))),
+        }
+    }
+
+    async fn try_payment_session(
```
quake, line 2253: suggest to extract two helper functions `build_route` and `send_onion_packet` to simplifying the main function.

and setting final error to `None`` outside the loop and then assigning it within the loop lead to unnecessary state management and logic complexity.

```
async fn try_payment_session(
    &self,
    state: &mut NetworkActorState<S>,
    mut payment_session: PaymentSession,
) -> Result<PaymentSession, Error> {
    let payment_data = payment_session.request.clone();
    let payment_hash = payment_data.payment_hash;

    while payment_session.can_retry() {
        payment_session.retried_times += 1;

        let hops_infos = self.build_route(payment_data.clone()).await?;
        let first_channel_outpoint = hops_infos[0]
            .channel_outpoint
            .clone()
            .expect("first hop channel outpoint");

        let tlc_id = self.send_onion_packet(state, hops_infos).await?;

        payment_session.set_status(PaymentSessionStatus::Inflight);
        payment_session.set_first_hop_info(first_channel_outpoint, tlc_id);
        self.store.insert_payment_session(payment_session.clone());

        return Ok(payment_session);
    }

    let error_msg = format!("Exhausted retries for payment: {:?}", payment_hash);
    payment_session.set_status(PaymentSessionStatus::Failed);
    payment_session.set_failed_status(&error_msg);
    self.store.insert_payment_session(payment_session);
    Err(Error::SendPaymentError(error_msg))
}
```

src/fiber/network.rs
```diff
@@ -3082,11 +3348,27 @@ where
...
+                if v == channel_id {
+                    Some(k.clone())
+                } else {
+                    None
+                }
+            }) {
+                self.outpoint_channel_map.remove(&outpoint);
+            }
+            channel
+        })
```
quake:
```suggestion
self.channels.remove(channel_id).inspect(|_| self.outpoint_channel_map.retain(|_, v| v != channel_id))
```

src/fiber/network.rs
chenyukang: When there is any error happened in `build_route` or `send_onion_packet`, instead of just throwing an exception to upper level, we need to set payment session status and store into db. 

let me think how to make the code simpler.

chenyukang:
Will refactor on PR #228, merge now.

PR #205, opened 2024-09-27
Add some unit tests for network syncing

contrun:


src/fiber/types.rs
```diff
@@ -363,6 +363,29 @@ impl Privkey {
...
     }
+
+    pub fn x_only_pub_key(&self) -> XOnlyPublicKey {
+        let secp256k1_instance = secp256k1_instance();
+        let secret_key = self.0;
+        let keypair = secp256k1::Keypair::from_secret_key(secp256k1_instance, &secret_key);
+        XOnlyPublicKey::from_keypair(&keypair).0
+    }
+
+    pub fn sign_schnorr(&self, message: [u8; 32]) -> SchnorrSignature {
```
quake, line 374: I think these two fn are for testing only, please add conditional flag for these and remove the unnecessary trace log.

src/fiber/graph.rs
```diff
@@ -336,9 +337,11 @@ where
                     && channel.funding_tx_block_number < end_block
             }),
             end_block,
```
chenyukang, line 339: please add a TODO to note we need a real pagination here.

chenyukang (approved):

PR #208, opened 2024-09-27
Add a peer store to network actor

### LINKED ISSUE #207: Bookkeep all the connected peers information in the network actor (instead of network graph)
https://github.com/contrun/fiber/blob/93ae2af22fed9822bf4b9468beea1e7301d04972/src/fiber/graph.rs#L132-L133

It is better for network graph to do one thing and one thing only, finding the route. We should save all the peers information to the network actor. We need to create a store for the network actor (This will also be useful latter. We can, for example, persist malicious peers), and save the connected peers to the network actor store.

contrun:
Closes https://github.com/nervosnetwork/fiber/issues/207

### DIFF
--- a/src/fiber/graph.rs
+++ b/src/fiber/graph.rs
@@ -11,8 +11,6 @@ use ckb_types::packed::{OutPoint, Script};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use std::collections::HashMap;
-use tentacle::multiaddr::Multiaddr;
-use tentacle::secio::PeerId;
 use thiserror::Error;
 use tracing::log::error;
 use tracing::{debug, info, warn};
@@ -138,8 +136,6 @@ pub struct NetworkGraph<S> {
     // Similar to the best_height, this is the last update time of the network graph.
     // We assume that we have already synced the graph up to this time - ASSUME_MAX_MESSAGE_TIMESTAMP_GAP.
     last_update_timestamp: u64,
-    // when we restarting a node, we will reconnect to these peers
-    connected_peer_addresses: HashMap<PeerId, Multiaddr>,
     nodes: HashMap<Pubkey, NodeInfo>,
     store: S,
     chain_hash: Hash256,
@@ -172,7 +168,6 @@ where
             last_update_timestamp: 0,
             channels: HashMap::new(),
             nodes: HashMap::new(),
-            connected_peer_addresses: HashMap::new(),
             store,
             chain_hash: get_chain_hash(),
         };
@@ -211,9 +206,6 @@ where
                 self.last_update_timestamp = node.timestamp;
             }
             self.nodes.insert(node.node_id, node.clone());
-        }
-        for (peer, addr) in self.store.get_connected_peer(None) {
-            self.connected_peer_addresses.insert(peer, addr);
         }
     }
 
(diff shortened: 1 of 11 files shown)


src/fiber/graph.rs
```diff
@@ -988,41 +949,6 @@ mod tests {
...
-        let connected_peers = network_graph.get_connected_peers();
-        assert_eq!(connected_peers.len(), 1);
-        assert_eq!(connected_peers[0], (&peer_id, &address));
-
-        network_graph.remove_connected_peer(&peer_id);
-        let connected_peers = network_graph.get_connected_peers();
-        assert_eq!(connected_peers.len(), 0);
-
-        network_graph.load_from_store();
-        let connected_peers = network_graph.get_connected_peers();
```
chenyukang: can we change this unit test for `PersistentNetworkActorState`.
just to make sure the apis work as expected will be good.

src/fiber/network.rs
```diff
@@ -1576,6 +1567,10 @@ where
                         self.network_graph.write().await.add_node(node_info);
 
                         // TODO: bookkeeping how many nodes we have connected to. Stop connnecting once we surpass a threshold.
```
chenyukang: we may resolve this TODO in this PR?

src/fiber/network.rs
```diff
@@ -1576,6 +1567,10 @@ where
...
+                        let peer_id = node_announcement.peer_id();
+                        state
+                            .persistent_state
+                            .save_peer_addresses(peer_id, node_announcement.addresses.clone());
```
chenyukang: should we save the data in DB after insert new one?
from the current code, seems we only save all the data in rocksdb when `post_stop` called,
i'm thinking if there fiber node is killed forcely, we can not get the data back after restarting?
since `save_peer_addresses` only save it in memory.

src/fiber/network.rs
```diff
@@ -3086,17 +3125,34 @@ where
             debug!("Tentacle service shutdown");
         });
 
+        let persistent_state = self
+            .store
+            .get_network_actor_state(&my_peer_id)
+            .unwrap_or_default();
+        let mut peers_to_connect = persistent_state.get_peers_to_connect();
```
chenyukang: any special reason to put `persistent_state` in `NetworkActorState` but not in `NetworkActor`? since in other actors like the `ChannelActor`, we use a `store` for persistent storage and `ChannelActorState` don't have a field for `persistent_state`.

src/fiber/network.rs
```diff
@@ -3196,6 +3252,9 @@ where
         if let Err(err) = state.control.close().await {
             error!("Failed to close tentacle service: {}", err);
         }
+        debug!("Saving network actor state for {:?}", state.peer_id);
+        self.store
+            .insert_network_actor_state(&state.peer_id, state.persistent_state.clone());
```
chenyukang: `persistent_state` make it confusing since it actually only works like a memory storage, maybe we need to add a `store` field for `persistent_state`, 
and update the changes into `store`, so that we can make sure data won't be lost event process is killed.

src/fiber/network.rs
contrun: This is because most fields in the network actor state are volatile, I add a persistent state to make the fact that these fields should be persisted more explicit.

src/fiber/network.rs
contrun: If I am going to add a store to the persistent_state, then there will be a cyclic dependency chain. The definition of the trait `NetworkActorStateStore` requires a `PersistentNetworkActorState` because we have to define what we want to save to/load from the store (this is the struct `PersistentNetworkActorState`). And the definition of  `PersistentNetworkActorState` contains the field `NetworkActorStateStore`, hence a dependency cycle. We should do something to ensure that we will not forget to save the state to persistent storage. This is the reason why I wrapped the function `save_peer_addresses` of `PersistentNetworkActorState` (which we should not call directly, we can only enforce this by rust's field/methods visibility) to a function `save_peer_addresses` of `NetworkActorState` (which does the final persistence work, and we should use this when possible).

src/fiber/network.rs
contrun: I feel the name `PersistentNetworkActorState` kind of sucks. It is actually the `NetworkActorStateRequiringPersistence`.

src/fiber/graph.rs
contrun: I wanted to make a unit test for this. The problem now for me is that `NetworkActorState` is too complex. In order to create a `NetworkActorState`, we almost certainly need to create a `NetworkActor`.  The pr https://github.com/nervosnetwork/fiber/pull/205 actually has an easier way to create and restart a network actor. I think that would make the test easier. Admittedly, that is kind of like an integration test. And we still need to solve the problem of how to easily write unit tests for `NetworkActorState`.

src/fiber/network.rs
```diff
@@ -2247,6 +2303,104 @@ pub struct NetworkActorState<S> {
...
+            }
+        }
+    }
+
+    pub(crate) fn sample_n_peers_to_connect(&self, n: usize) -> HashMap<PeerId, Vec<Multiaddr>> {
+        let nodes = self
+            .saved_peer_addresses
+            .keys()
+            .into_iter()
+            .chain(self.saved_peer_addresses.keys().into_iter())
```
quake: why chained `saved_peer_addresses` twice here?

quake (changes requested):

PR #210, opened 2024-10-03
feat: add shutdown_script to rpc

quake:
this PR reverted some changes of https://github.com/nervosnetwork/fiber/pull/174 by reusing the `shutdown_script` field of ChannelActorState, and added an optional field to the rpc `open_channel` and `accept_channel`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,9 +1,11 @@
+use ckb_hash::blake2b_256;
 use fnn::cch::CchMessage;
-use fnn::ckb::contracts::init_contracts_context;
+use fnn::ckb::contracts::{get_script_by_contract, init_contracts_context, Contract};
 use fnn::fiber::graph::NetworkGraph;
 use fnn::store::Store;
 use fnn::watchtower::{WatchtowerActor, WatchtowerMessage};
 use ractor::Actor;
+use secp256k1::Secp256k1;
 use tentacle::multiaddr::Multiaddr;
 use tokio::sync::{mpsc, RwLock};
 use tokio::{select, signal};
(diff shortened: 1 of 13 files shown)


tests/bruno/e2e/cross-chain-hub/13-node1-send-shutdown-channel.bru
```diff
@@ -41,5 +41,5 @@ assert {
 
 script:post-response {
   // Sleep for sometime to make sure current operation finishes before next request starts.
-  await new Promise(r => setTimeout(r, 100));
+  await new Promise(r => setTimeout(r, 1000));
```
chenyukang: I guess it is not a timing issue, this step failed with 500:
```
e2e/cross-chain-hub/17-node1-close-ln-channel (500 Internal Server Error) - 3 ms
```

src/main.rs
```diff
@@ -100,6 +102,12 @@ pub async fn main() {
                 node_public_key.clone().into(),
             )));
 
+            let secret_key = ckb_config.read_secret_key().unwrap();
+            let secp = Secp256k1::new();
```
contrun, line 106: Using `secp256k1_instance` in `src/fiber/types.rs` is better, because that function only initialize secp251k1 instance once and keep it in memory (there are many places we need to use the secp256k1 instance).

contrun (approved):

tests/bruno/e2e/cross-chain-hub/13-node1-send-shutdown-channel.bru
quake: yes, I found the root cause: the shutdown script is different with the one when building commitment tx, the latest commit should resolve it.

PR #226, opened 2024-10-14
Invoice rpc generate signed invoice if there is a config

chenyukang:
A follow-up for https://github.com/nervosnetwork/fiber/pull/225

### DIFF
--- a/src/rpc/invoice.rs
+++ b/src/rpc/invoice.rs
@@ -1,16 +1,16 @@
-use std::time::Duration;
-
 use crate::fiber::hash_algorithm::HashAlgorithm;
 use crate::fiber::serde_utils::{U128Hex, U64Hex};
-use crate::fiber::types::Hash256;
+use crate::fiber::types::{Hash256, Privkey};
 use crate::invoice::{CkbInvoice, Currency, InvoiceBuilder, InvoiceStore};
+use crate::FiberConfig;
 use ckb_jsonrpc_types::Script;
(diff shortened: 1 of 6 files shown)


src/rpc/invoice.rs
```diff
@@ -64,12 +64,26 @@ trait InvoiceRpc {
...
+    keypair: Option<(PublicKey, SecretKey)>,
 }
 
 impl<S> InvoiceRpcServerImpl<S> {
-    pub(crate) fn new(store: S, public_key: Option<PublicKey>) -> Self {
-        Self { store, public_key }
+    pub(crate) fn new(store: S, config: Option<FiberConfig>) -> Self {
+        let keypair = config.map(|config| {
+            let kp = config
+                .read_or_generate_secret_key()
```
doitian, line 74: Is it safe to be called in multiple threads concurrently?

doitian: I'm afraid not. Both threads may pass the test that the key does not exist. One thread will use the keypair which is later been overwritten.

https://github.com/chenyukang/fiber/blob/8015b035af68234d74fbcdb8aec0737e2b886693/src/fiber/key.rs#L26

chenyukang: I think it's not a real issue, `read_or_generate_secret_key` must be called firstly when starting the network actor in the main.

but yeah, we'd better make it thread-safe with lazy static or something.

PR #228, opened 2024-10-14
Add payment history for probability calculation in path finding

chenyukang:
1. Record payment history
2. Add probability calculation based on payment history


src/fiber/history.rs
```diff
@@ -0,0 +1,422 @@
...
+                {
+                    return;
+                }
+                current.fail_amount = amount;
+                current.fail_time = time;
+                if amount == 0 {
+                    current.success_amount = 0;
+                } else if amount <= current.success_amount {
+                    current.success_amount = amount - 1;
+                }
```
quake: the condition `amount == 0` implies `amount <= current.success_amount`. 

```suggestion
if amount <= current.success_amount {
                    current.success_amount = amount.saturating_sub(1);
}
```

src/fiber/history.rs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    // Get the probability of a payment success through a direct channel,
+    // suppose we know the accurate balance for direct channels, so we don't need to use `get_channel_probability`
+    // for the direct channel, this function is used disable the direct channel for a time preiod if it's failed
+    // currently we may mark the channel failed on graph level, so this function is not used now.
+    // FIXME: reconsider this after we already got the accurate balance of direct channels
+    //        related issue: https://github.com/nervosnetwork/fiber/issues/257
+    #[allow(dead_code)]
+    pub(crate) fn get_direct_probability(&self, from: Pubkey, target: Pubkey) -> f64 {
```
quake: suggest to use ref:
```suggestion
    pub(crate) fn get_direct_probability(&self, from: &Pubkey, target: &Pubkey) -> f64 {
```

quake (approved):

PR #230, opened 2024-10-15
Remove ldk

contrun:
In the initial iteration, I thought of integrating ldk into fiber node. Now that we are using lnd rpc for cross chain, we should just remove the obsolete code of ldk.


chenyukang:
nice, coverage +2.94%.

PR #243, opened 2024-10-17
Trivial fix on coverage

chenyukang:
1. `RUST_LOG=off` to remove noise messages from unit test.
2. Add `make coverage` for convinent
3. Ignore lines with all empty spaces, or only `)`, `}`, `);` etc.


chenyukang:
how to set to show `Codecov` results fro every PR?

Issue #252, opened 2024-10-19
[open channel] attempt to add with overflow panic


gpBlockchain:
funding_amount : 0xfffffffffffffffffffffffffffffff
```
curl --location 'http://127.0.0.1:8228' --header 'Content-Type: application/json' --data '{
    "id": 42,
    "jsonrpc": "2.0",
    "method": "open_channel",
    "params": [
        {
            "peer_id": "Qmai8nVGMooBv3WcspFKvQwjUKDXwu11PGQEfLbELMuJSo",
            "funding_amount": "0xfffffffffffffffffffffffffffffff",
            "public": true
        }
    ]
}'
response:
{"jsonrpc": "2.0", "result": {"temporary_channel_id": "0xf0f77ccdc459759411e2141e35691b6f7e098fd6403ec126752afcfc59965d58"}, "id": 42}

```
log
```
[...]

PR #256, opened 2024-10-20
check CKB amount range when open channel

chenyukang:
For issue #252, upgrade ckb-rust-sdk will resolve the issue, but user can not get a proper error message since it's an error handled by auto accept code.

This pr add an extra check to make sure CKB amount is less than `u64::max`, since we use `u128` for all amount considering UDT amount.

### DIFF
--- a/src/fiber/network.rs
+++ b/src/fiber/network.rs
@@ -2641,11 +2641,19 @@ where
         udt_type_script: &Option<Script>,
     ) -> Result<(u128, u64), ProcessingChannelError> {
         let reserved_ckb_amount = default_minimal_ckb_amount(udt_type_script.is_some());
-        if udt_type_script.is_none() && funding_amount < reserved_ckb_amount.into() {
-            return Err(ProcessingChannelError::InvalidParameter(format!(
-                "The value of the channel should be greater than the reserve amount: {}",
-                reserved_ckb_amount
-            )));
+        if udt_type_script.is_none() {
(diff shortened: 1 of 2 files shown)


src/fiber/network.rs
```diff
@@ -2641,11 +2641,19 @@ where
...
         let reserved_ckb_amount = default_minimal_ckb_amount(udt_type_script.is_some());
-        if udt_type_script.is_none() && funding_amount < reserved_ckb_amount.into() {
-            return Err(ProcessingChannelError::InvalidParameter(format!(
-                "The value of the channel should be greater than the reserve amount: {}",
-                reserved_ckb_amount
-            )));
+        if udt_type_script.is_none() {
+            if funding_amount < reserved_ckb_amount.into() {
+                return Err(ProcessingChannelError::InvalidParameter(format!(
+                    "The value of the channel should be greater than the reserve amount: {}",
```
quake:
```suggestion
                    "The funding amount should be greater than the reserved amount: {}",
```

src/fiber/network.rs
```diff
@@ -2641,11 +2641,19 @@ where
...
+                    reserved_ckb_amount
+                )));
+            }
+            if funding_amount >= u64::MAX as u128 {
+                return Err(ProcessingChannelError::InvalidParameter(format!(
+                    "The CKB amount of the channel should be less than {:?}",
```
quake:
```suggestion
                    "The funding amount should be less than {:?}",
```

PR #261, opened 2024-10-21
fix: force shutdown command should be handled specially

quake:
The force shutdown command should be handled specially because the ChannelActor may not exist when remote peer is disconnected.

Please notes that this PR contains incompatible change:

1. RPC will return error when call `shutdown` on a closed channel, ref: https://github.com/nervosnetwork/fiber/pull/261/commits/fbcc60c4e397fb78c65fbd7b728ff9c2740baffb

2. Remove the force field from p2p message, ref: https://github.com/nervosnetwork/fiber/pull/261/commits/af1eedd4ea2e46b2ee33089b283f1d6d73e90340


tests/bruno/e2e/3-nodes-transfer/22-node2-send-shutdown-channel-1.bru
chenyukang, line 1: need to change the previous e2e 23 step into 22?

quake: I don't think it's an issue as long as we keep the bruon script executing seq ordered

src/fiber/channel.rs
```diff
@@ -926,57 +916,34 @@ where
...
-                "Channel state updated to {:?} after processing shutdown command",
-                &state.state
-            );
+        let shutdown_info = ShutdownInfo {
+            close_script: command.close_script,
+            fee_rate: command.fee_rate.as_u64(),
+            signature: None,
+        };
+        state.local_shutdown_info = Some(shutdown_info);
+        state.update_state(ChannelState::ShuttingDown(
```
chenyukang, line 938: suppose the scenario node2 is offline, then node1 shotdown force, what will happen if `node2` restarted the node, `node2` will try to reestablish or maybe should also update the channel status too. maybe we need to add some code for node2 to handle this scneario.

tests/bruno/e2e/3-nodes-transfer/22-node2-send-shutdown-channel-1.bru
chenyukang: yeah, it just my OCD to keep file name with seq consistent 😅

src/fiber/network.rs
```diff
@@ -2758,12 +2759,59 @@ where
...
+                                ),
+                            ))
+                            .expect(ASSUME_NETWORK_ACTOR_ALIVE);
+
+                        state.update_state(ChannelState::ShuttingDown(
+                            ShuttingDownFlags::WAITING_COMMITMENT_CONFIRMATION,
+                        ));
+                        self.store.insert_channel_actor_state(state);
+
+                        let _ = rpc_reply.send(Ok(()));
```
contrun, line 2803: We didn't return error to the client if above code runs into an error condition.

src/fiber/tests/channel.rs
```diff
@@ -1155,3 +1155,34 @@ async fn test_reestablish_channel() {
...
+                    },
+                    rpc_reply,
+                ),
+            },
+        ))
+    };
+
+    call!(node_a.network_actor, message)
+        .expect("node_a alive")
+        .expect("successfully shutdown channel");
```
contrun, line 1187: It is better if we can check the final state of the channel after issuing the `Shutdown` command. We can check that by reading channels from the field `store` for `NetworkNode`. Here is an example https://github.com/nervosnetwork/fiber/pull/263/commits/2abf50c3eea90175ad0c3b1583d1376976ac34ec

src/fiber/channel.rs
quake: node2's watchtower will handle this

PR #266, opened 2024-10-21
chore: check before connect

quake:


contrun (approved):
LGTM. We can also remove the check in https://github.com/contrun/fiber/blob/d8405d14b46205d0e9bdaa756d06119f15f92fd3/src/fiber/network.rs#L1259 , which would become useless.

src/fiber/network.rs
```diff
@@ -1203,7 +1204,21 @@ where
             NetworkActorCommand::ConnectPeer(addr) => {
                 // TODO: It is more than just dialing a peer. We need to exchange capabilities of the peer,
                 // e.g. whether the peer support some specific feature.
-                // TODO: If we are already connected to the peer, skip connecting.
+
+                if let Some(peer_id) = extract_peer_id(&addr) {
```
contrun, line 1208: We also need to check whether the address belongs to ourselves. Otherwise, we may encounter errors like `HandshakeError(SecioError(ConnectSelf))`.

chenyukang (approved):

src/fiber/network.rs
quake: already checked in line 1213

PR #277, opened 2024-10-23
chore: refactor delay_epoch

quake:
1. add default delay epoch (24 hours), valid range is 1 hour ~ 14 days.
2. deleted unused fields (payment_base_key, delayed_payment_base_key, revocation_pubkey)
3. changed the channel id generation from `revocation_basepoint` to `tcl_basepoint`


src/rpc/channel.rs
```diff
@@ -290,6 +291,10 @@ where
                     funding_amount: params.funding_amount,
                     public: params.public.unwrap_or(false),
                     shutdown_script: params.shutdown_script.clone().map(|s| s.into()),
+                    commitment_delay_epoch: params
```
contrun, line 311: - The delay epoch is not validated here. It can be too long or too short.
- There are some bit operation involved in `EpochNumberWithFraction`. We are accepting a u64 as parameter here, but it can be hard for human to construct a valid and meaningful u64 for this `EpochNumberWithFraction` here.

contrun (approved):
Generally LGTM. Two small comments.

src/rpc/channel.rs
quake: resolved in commit 088e0db1ca1f598d73d89fd0d3975f773412cdee

tests/bruno/e2e/open-use-close-a-channel/02-open-channel.bru
```diff
@@ -23,7 +23,8 @@ body:json {
     "params": [
       {
         "peer_id": "{{NODE1_PEERID}}",
-        "funding_amount": "0xba43b7400"
+        "funding_amount": "0xba43b7400",
+        "commitment_delay_epoch": "0x30001000002"
```
contrun: This seems to be 2 and 1/3 epochs. DId you mean to use `0x20001000003` (if we are using big endian to de-serialize u64)?

quake: 3.5 epochs is presented as 3(E) and 1(I)/2(L): (L << 40) | (I << 24) | E = 0x20001000003,

ref: https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0017-tx-valid-since/e-i-l-encoding.png

contrun: Yeah. But the number in this bruno file is `0x30001000002`, instead of the number in your reply, `0x20001000003`.

quake: sorry, I thought that you are ref to https://github.com/nervosnetwork/fiber/blob/568eec50e27f8e4cfda3b1d2b9cb48fb810b93f3/tests/bruno/e2e/open-use-close-a-channel/README.md?plain=1#L30 , fixed in new commit.

PR #291, opened 2024-10-27
Add RPC documentation generator

chenyukang:
Some other ways I have tried:
1. use json format from `cargo doc`, but it's only available for nightly version of rustc, and we also need to convert from `json` format to markdown format.
2. use https://github.com/GREsau/schemars and add `JsonSchema` for all data structures which need to be listed in markdown, I think it's an invasive change, and there are some trivial issues need to be resolved, such as https://github.com/GREsau/schemars/pull/251

The current simple but dirty solution https://github.com/chenyukang/rpc-doc-gen is based on `syn` crate, which may have some pitfalls but I think we can improve it later.


contrun:
Thanks for tackling this. It is a life quality improvement. I skimmed through the generator implementation.

Current implementation didn't consider foreign types in enum/struct fields, right? For example, the `Multiaddr` parameter of the RPC `connect_peer` is not documented (if we are using cargo doc to generate json files, I would expect we have foreign type information).

Another thing users want to know is how are these types serialized and deserialized. For example, if I want to pass a `Multiaddr` to the RPC server, what is the appropriate format? I don't think it is quite possible for us to automatically document that, is it?

A small problem of the generated document is that it seems the type name of any `Option` fields are quoted, while other type names are not. For example, below the type name `Hash256` is not quoted but `Option<u64>` is.

> `channel_id` - Hash256, Channel ID for the CKB payment.
> `tlc_id` - `Option<u64>`, TLC ID for the CKB payment.

chenyukang:
yes, there are some improvements that need to be resolved:
1. we need to analyze all the code base, I already tried it but need some tweak for the generator to dump only the related types with RPC.
2. all types should be quoted, and `Option<u64>` will be better in the format `null` | `u64`, and `Option<SomeStruct>` will be in format ` null` | [SomeStruct](link_to_struct_type).

PR #295, opened 2024-10-29
chore: refactor ContractsContext and setup chain hash

quake:
Note: this PR is a breaking change, the config file's section `fiber` fields is changed, please refer to [config/testnet/config.yml](https://github.com/nervosnetwork/fiber/commit/a8b7ddc7bc5703e40461bd17a5fcd3ff660923bc#diff-95c9cfab0d508c6c678ffdc02d6688f2d7a2cd0eeeee1d2487504dfca15f880f)


src/main.rs
```diff
@@ -71,7 +74,21 @@ pub async fn main() {
...
+            let chain = fiber_config.chain.as_str();
+            let chain_spec = ChainSpec::load_from(&match chain {
+                "mainnet" => Resource::bundled("specs/mainnet.toml".to_string()),
+                "testnet" => Resource::bundled("specs/testnet.toml".to_string()),
+                path => Resource::file_system(Path::new(&config.base_dir).join(path)),
+            })
+            .expect("load chain spec");
+            let genesis_block = chain_spec.build_genesis().expect("build genesis block");
+
+            init_chain_hash(genesis_block.hash().into());
```
contrun, line 86: `init_chain_hash` and `init_contracts_context` should also be called in the test environment. It is strange that CI tests shall pass. I tried `cargo nextest run`, and I found the following tests were failing.

```
     Summary [   6.553s] 33/94 tests run: 25 passed, 8 failed, 0 skipped
        FAIL [   0.156s] fnn fiber::tests::graph::test_graph_build_route_below_min_htlc_value
        FAIL [   0.156s] fnn fiber::tests::graph::test_graph_build_route_exceed_max_htlc_value
        FAIL [   0.161s] fnn fiber::tests::graph::test_graph_build_route_three_nodes
        FAIL [   0.164s] fnn fiber::tests::graph::test_graph_channel_info
        FAIL [   0.218s] fnn fiber::tests::graph::test_graph_find_optimal_path
        FAIL [   0.200s] fnn fiber::tests::graph::test_graph_find_path_amount_failed
        FAIL [   0.176s] fnn fiber::tests::graph::test_graph_find_path_basic
        FAIL [   0.206s] fnn fiber::tests::graph::test_graph_find_path_cycle
```

But the return code of `cargo test` is 0. I don't know what's going on, but when I run `cargo test fiber::tests::graph::test_graph_find_path_cycle`, I can clearly see the unit test is failing.

chenyukang: when run in `cargo test` it will be called at here : 
https://github.com/nervosnetwork/fiber/blob/260912b5b507d16fa00722d83e12ecc1de12d7ed/src/ckb/tests/test_utils.rs#L107

so multiple test share the some global setting.
when run specific test, it's not be invoked, so it's failing.

src/ckb/contracts.rs
```diff
@@ -1,456 +1,207 @@
...
-                        .unwrap_or_else(|_| panic!("environment variable {tx1_env_name}")),
-                )
-                .expect("valid hash");
+            "0x10639e0895502b5688a6be8cf69460d76541bfa4821629d86d62ba0aae3f9606" => {
+                info!("Creating ContractsContext for testnet");
+            }
+            _ => {
+                info!("Creating ContractsContext for dev");
+                // index from 5 ~ 8 are the default contracts: CkbAuth, FundingLock, CommitmentLock, SimpleUDT
+                let ckb_auth_cell_dep = CellDep::new_builder()
```
chenyukang, line 93: it seems better with this trivial refactor: https://github.com/quake/cfn-node/pull/5

src/main.rs
quake: resolved in 817d861, thanks for finding.

chenyukang (approved):

PR #296, opened 2024-10-30
chore: refactoring htlc expiry field related code

quake:


src/fiber/channel.rs
```diff
@@ -138,7 +138,7 @@ pub struct AddTlcCommand {
     pub amount: u128,
     pub preimage: Option<Hash256>,
     pub payment_hash: Option<Hash256>,
-    pub expiry: LockTime,
+    pub expiry: u64,
```
contrun, line 141: Depending on the code location, there can be two interpretation of `expiry: u64`. One is the TLC will expire at some absolute timestamp, and the other is that the TLC will expire after some time duration.

Here we view the u64 number as the absolute timestamp for the TLC expiry. Renaming the field name to `expiry_timestamp_ms` will make this fact clearer.

tests/bruno/e2e/udt/08-node1-add-tlc.bru
```diff
@@ -25,7 +25,7 @@ body:json {
         "channel_id": "{{N1N2_CHANNEL_ID}}",
         "amount": "{{payment_amount}}",
         "payment_hash": "{{payment_hash}}",
-        "expiry": 40
+        "expiry": "0x20000000000"
```
contrun, line 28: The interpretation of this expiry number is a little bit muddy. Do we want the clients to pass time duration or absolute timestamp for the expiry time? Since we are using this number directly for the `AddTlc` message, current code deems this number as an absolute timestamp. This invalidates all of the tests here, as all of they are invalid parameters (0x20000000000 is already a time in the past).

No matter what meaning we decide to choose, current documentation should be changed accordingly. https://github.com/contrun/fiber/blob/4fdf0042a5af3878b069cbee84451f29b05e65a7/src/rpc/README.md?plain=1#L135

src/fiber/channel.rs
```diff
@@ -1105,7 +1105,7 @@ where
                 tlc_id: tlc.id.into(),
                 amount: tlc.amount,
                 payment_hash: tlc.payment_hash,
-                expiry: tlc.lock_time,
```
contrun, line 1108: We are sending an unverified timestamp to the peer. This may be a time in the past.

contrun (commented):
``

src/fiber/channel.rs
quake: all expiry time related verification should be done in a new PR

src/fiber/channel.rs
quake: > Depending on the code location, there can be two interpretation of `expiry: u64`. One is the TLC will expire at some absolute timestamp, and the other is that the TLC will expire after some time duration.

we only use absolute timestamp for tlc expiry time in our codebase

> Here we view the u64 number as the absolute timestamp for the TLC expiry. Renaming the field name to `expiry_timestamp_ms` will make this fact clearer.

add a field comment should be enough

PR #304, opened 2024-11-13
Add unit test for find router

chenyukang:
The test case `test_graph_build_route_with_double_edge_node` is used for the pay self corner case, indicating `last_hop_channels` must be a `HashMap`.

Since the iterate order is not determined for `get_node_inbounds` at https://github.com/nervosnetwork/fiber/blob/856b98bc7f9819bba4d46047b4314cf79e8a73a4/src/fiber/graph.rs#L660-L661, this test case may have some possibility to fail if we use `HashSet` for `last_hop_channels`.

To reproduce it run the script:

```bash
for i in {1..10}; do cargo test test_graph_build_route_with_double_edge_node  -- --nocapture; done
```

### DIFF
--- a/src/fiber/graph.rs
+++ b/src/fiber/graph.rs
@@ -721,7 +721,6 @@ where
                     );
 
                 if probability < DEFAULT_MIN_PROBABILITY {
-                    debug!("probability is too low: {:?}", probability);
                     continue;
                 }
                 let agg_weight =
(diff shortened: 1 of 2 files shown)


chenyukang:
if we sort the channels from `get_node_inbounds` according to fee rate from high to low, it will make it reproducible each time for this issue.

PR #327, opened 2024-11-20
Funding cell exclusion

### LINKED ISSUE #74: Cells exclusion for active funding tx negotiation
Exclude cells used in the pending funding tx negotiation so that cocurrent funding tx will not conflict with each other.

- [x] Add data structure to manage the exclusion list
- [x] Apply exclusion list before building tx
- [x] Update exclusion list after building tx
- [x] Remove tx from exclusion list when channel has been established
- [ ] Add a test case for in-memory cells exclusion (blocked by #348)
- [ ] Persist funding txs
- [ ] Restore exclusion list after restart
- [ ] Add RPC to cancel a pending funding tx

doitian:
Closes #74

### DIFF
--- a/src/ckb/actor.rs
+++ b/src/ckb/actor.rs
@@ -118,7 +118,7 @@ impl Actor for CkbChainActor {
                 let context = state.build_funding_context(&request);
                 if !reply_port.is_closed() {
                     tokio::task::block_in_place(move || {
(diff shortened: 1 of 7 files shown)


chenyukang:
Close this PR for https://github.com/nervosnetwork/fiber/pull/521

PR #339, opened 2024-11-25
Use rocksdb store in test

chenyukang:
`make test` time changed from 12.0 seconds to 14.0, I think it's affordable.


chenyukang:
TempDir will cleaned up when exiting: https://github.com/nervosnetwork/fiber/blob/7e310b6b3114448c57ee956c40badc39f7606073/src/fiber/tests/test_utils.rs#L63

quake (approved):

Issue #198, opened 2024-09-25
Integrate fiber sphinx to return errors of payment requests


doitian:
https://github.com/cryptape/fiber-sphinx

Error processing has been added in PR #191 without the real onion error packet.

doitian:
Related: https://github.com/cryptape/fiber-sphinx/pull/7

PR #354, opened 2024-11-29
Setup onion error packet

doitian:
Setup the framework to return payment error via onion error packet.

The spec can be found here: https://github.com/cryptape/fiber-sphinx/blob/develop/docs/spec.md#returning-errors

The encryption has not been enabled yet so I can split the changes into several PRs. The encryption can be enabled later without breaking the network protocol because the decryption phrase will check whether the error has been encrypted or not.

Refs #198


chenyukang (approved):

doitian:
Forced push after `cargo fmt`. Nothing new to review thus I'll merge it because it has been approved.

PR #358, opened 2024-11-30
feat: use type safe onion packet

doitian:
Use type safe onion pack instead of `Vec<u8>`, which is error-prone since at some places we store `PaymentOnionPacket`, and at other places we may store `PeeledPaymentOnionPacket`.

### DIFF
--- a/src/cch/actor.rs
+++ b/src/cch/actor.rs
@@ -595,7 +595,7 @@ impl CchActor {
                                 expiry: now_timestamp_as_millis_u64()
                                     + self.config.ckb_final_tlc_expiry_delta,
                                 hash_algorithm: HashAlgorithm::Sha256,
-                                onion_packet: vec![],
+                                peeled_onion_packet: None,
                                 previous_tlc: None,
                             },
                             rpc_reply,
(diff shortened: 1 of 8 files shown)


doitian:
The test case below will fail. @chenyukang can you help to debug? It seems that I have passed wrong parameters when forwarding the onion packet.

I also added `channel_outpoint` and `next_hop` to the struct `TLC`, because the current interface requires a `PeeledPaymentOnionPacket` to send the packet to the next hop.

```
cargo test fiber::tests::channel::test_send_payment_with_3_nodes -- --nocapture
```

doitian:
> The test case below will fail. @chenyukang can you help to debug? It seems that I have passed wrong parameters when forwarding the onion packet.
> 
> I also added `channel_outpoint` and `next_hop` to the struct `TLC`, because the current interface requires a `PeeledPaymentOnionPacket` to send the packet to the next hop.
> 
> ```
> cargo test fiber::tests::channel::test_send_payment_with_3_nodes -- --nocapture
> ```

Found the cause, the amount is not correct when creating outbounding tlc from the pending inbounding tlc.

quake (approved):

PR #391, opened 2024-12-15
test: add more watchtower e2e test and resolve some force close issues

quake:


src/rpc/dev.rs
```diff
@@ -0,0 +1,105 @@
...
+use tokio::sync::RwLock;
+
+use crate::{
+    ckb::CkbChainMessage,
+    fiber::{network::DEFAULT_CHAIN_ACTOR_TIMEOUT, serde_utils::U64Hex, types::Hash256},
+};
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub(crate) struct SubmitCommitmentTransactionParams {
```
chenyukang, line 22: we should move AddTlc and RemoveTlc from rpc/channel to here?

quake: yes, I plan to do it after this PR.

src/fiber/channel.rs
```diff
@@ -1264,9 +1206,42 @@ where
         state: &mut ChannelActorState,
         command: ShutdownCommand,
     ) -> ProcessingChannelResult {
-        // The force shutdown command has been handled speically in the `NetworkActorState#send_command_to_channel` function.
-        // We only need to handle the normal shutdown command here.
         debug!("Handling shutdown command: {:?}", &command);
+        if command.force {
+            match state.state {
+                ChannelState::ChannelReady() => {
```
chenyukang, line 1221: I'm not sure why we add force shutdown logic here, since it's moved into network actor in this PR: https://github.com/nervosnetwork/fiber/pull/261 . Now there are two copies of code since the related code in network actor is not deleted.

quake: When both nodes are online, if one of them force close the channel, current code only updates the actor state in the store, not the state in memory, which will cause the other node to continue to send it messages such as addtlc.

src/fiber/network.rs
```diff
@@ -3395,50 +3362,57 @@ where
...
-                            }
-                        };
+                if let Some(actor) = self.channels.get(&channel_id) {
+                    actor.send_message(ChannelActorMessage::Command(ChannelCommand::Shutdown(
+                        shutdown, rpc_reply,
+                    )))?;
+                    Ok(())
+                } else {
+                    match self.store.get_channel_actor_state(&channel_id) {
+                        Some(mut state) => {
```
chenyukang, line 3356: I guess our test will all go to above branch, in which scenario this branch will cover,maybe when restarting a killed node?

chenyukang (approved):

chenyukang:
This will resolve issue #202?

quake:
> This will resolve issue #202?

partially, some pending tlc cases are not covered yet

PR #397, opened 2024-12-17
Improve path finding and send payment error handling

chenyukang:
- considering direct channel balance in path finding
- add more unit testing for send payment and error handling
- code refactoring


src/fiber/graph.rs
```diff
@@ -779,6 +797,22 @@ where
...
                 }
 
+                // if this is a direct channel, try to load the channel actor state for balance
+                if from == self.source {
+                    if let Some(state) = self
+                        .store
+                        .get_channel_state_by_outpoint(&channel_info.out_point())
+                    {
+                        if amount_to_send > state.to_local_amount {
+                            debug!(
```
quake: suggest to remove this debug log, it's too noisy

quake (approved):

src/fiber/graph.rs
chenyukang: fixed.

PR #399, opened 2024-12-18
tune ci cache

doitian:
- ci: select rust toolchain before cache

    The build matrix should use separate caches


chenyukang (approved):

contrun:
Does it actually work or is this just a plausible guess?

doitian:
> Does it actually work or is this just a plausible guess?

See the first comment in the example usage https://github.com/Swatinem/rust-cache?tab=readme-ov-file#example-usage

PR #414, opened 2024-12-21
chore: simplify fn prune_messages_to_be_saved

quake:
The current implementation of prune_messages_to_be_saved is complex and unintuitive, it can be implemented in a much simpler way.

### DIFF
--- a/src/fiber/gossip.rs
+++ b/src/fiber/gossip.rs
@@ -1,6 +1,7 @@
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, HashSet},
     marker::PhantomData,
+    mem::take,
     sync::Arc,
     time::Duration,
 };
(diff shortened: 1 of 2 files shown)


src/fiber/gossip.rs
```diff
@@ -1017,83 +1018,30 @@ impl<S: GossipMessageStore> ExtendedGossipMessageStoreState<S> {
...
-                        sorted_messages.push(message);
-                    }
-                }
-            }
-            for message in messages {
-                sorted_messages.push(message);
-            }
-        }
+        let mut sorted_messages = complete_messages.into_iter().collect::<Vec<_>>();
+        sorted_messages.sort_unstable();
```
contrun, line 1028: This seems to use the Ord implementation of BroadcastMessageWithTimestamp, which will sort messages according to their timestamp. It is possible that the timestamp of ChannelUpdate is smaller than that of Channel announcement (we didn't have code to specifically make the timestamp of ChannelUpdate later than the timestamp of the funding transaction block, we just use the system time).

quake: will order by message_id first:

https://github.com/nervosnetwork/fiber/blob/1a2ff8f5a044b7041c5abe678a0386d74b77970b/src/fiber/types.rs#L2435-L2441

contrun: Sorry. I didn't look carefully. In that case, this PR LGTM. I created a few tests for the dependency of channel gossip messages in https://github.com/nervosnetwork/fiber/pull/418 . Can you take a look and check if the code change here does not break the tests (I believe it doesn't. Still, it is better to
have some tests)?

quake: rebased and pushed

contrun (approved):

src/fiber/gossip.rs
```diff
@@ -1017,83 +1018,30 @@ impl<S: GossipMessageStore> ExtendedGossipMessageStoreState<S> {
...
-                }
-                _ => {
-                    messages.insert(index, new_message);
-                }
-            }
-        }
+        let messages_to_be_saved = take(&mut self.messages_to_be_saved);
+        let (complete_messages, uncomplete_messages) = messages_to_be_saved
+            .into_iter()
+            .partition(|m| self.has_dependencies_available(m));
```
contrun, line 1024: There is a subtle bug here. Because we replaces `messages_to_be_saved` in line 1021, the line `1024` will always return false for `ChannelUpdate` messages that do not have corresponding `ChannelAnnouncement` in the store. Previously, the behavior is to also check corresponding `ChannelAnnouncement` in `messages_to_be_saved`. This bug is fixed in https://github.com/contrun/fiber/commit/8ca34e017989fb0c7f513751bff250faee2ae777 . I will create a PR later.

src/fiber/types.rs
```diff
@@ -2592,6 +2607,42 @@ pub enum BroadcastMessageID {
...
+            (
+                BroadcastMessageID::ChannelUpdate(outpoint1),
+                BroadcastMessageID::ChannelUpdate(outpoint2),
+            ) => outpoint1.cmp(outpoint2),
+            (
+                BroadcastMessageID::NodeAnnouncement(pubkey1),
+                BroadcastMessageID::NodeAnnouncement(pubkey2),
+            ) => pubkey1.cmp(pubkey2),
+            (BroadcastMessageID::ChannelUpdate(_), _) => Ordering::Less,
+            (BroadcastMessageID::NodeAnnouncement(_), _) => Ordering::Greater,
```
contrun, line 2628: Did you intend to order messages by making `NodeAnnouncement` in the first place, `ChannelAnnouncement` the second place and `ChannelUpdate` the last place (this way the sorted messages in the list `messages_to_be_saved` can be processed sequentially)? If that is the case, I think the correct Ordering should be https://github.com/contrun/fiber/commit/ea5c1fede9bffd8d3f3fa49d2570aadb971fcb6f. This problem is not discovered in the test because another bug above always excluding `ChannelUpdate` from being included in the complete messages when there is a corresponding `ChannelAnnouncement` in the `messages_to_be_saved`.

PR #416, opened 2024-12-22
Tlc operations wait dual ack

chenyukang:
- Rewrite tlc operations with statuses 
- Forwarding TLC need to ack from peer
- Removing TLC need ack from peer


src/fiber/channel.rs
```diff
@@ -654,73 +658,139 @@ where
...
+            let _ = self
+                .apply_remove_tlc_operation(myself, state, *tlc_id, reason.clone())
+                .await;
+        }
+
         Ok(())
     }
 
-    async fn flush_staging_tlc_operations(
+    async fn flush_tlc_operations(
```
contrun: It is my understanding that `flush_tlc_operations` is called whenever the peer ack'd our `CommitmentSigned` and we want to update the status of those TLCs added/removed in that commitment transaction. The function name `flush_tlc_operations` does not have any implication in this regard. Maybe we can rename this to something like `update_tlc_status_on_ack` to reflect more on why and when to call this function.

src/fiber/channel.rs
```diff
@@ -1126,7 +1212,16 @@ where
...
         state.save_remote_nonce_for_raa();
-
+        if state
+            .tlc_state
+            .all_tlcs()
+            .filter(|tlc| tlc.is_offered())
+            .count()
+            > 0
+        {
+            state.tlc_state.set_waiting_ack(true);
```
contrun: Why are we only expecting ack if there are offered TLCs? Say, A sent B a payment, and everything goes smoothly, B finally wants to remove this TLC with success, in this case, B is also expecting, a RevokeAndAck from A. It is hard for me to imagine any case where we send a `CommitmentSigned` and not expecting a `RevokeAndAck`.

src/fiber/channel.rs
```diff
@@ -2439,224 +2553,173 @@ impl TlcState {
...
+    pub fn set_offered_tlc_removed(
+        &mut self,
+        tlc_id: u64,
+        remove_at: CommitmentNumbers,
+        reason: RemoveTlcReason,
+    ) {
+        for tlc in self.offered_tlcs.iter_mut() {
+            if Into::<u64>::into(tlc.tlc_id) == tlc_id {
+                tlc.removed_at = Some((remove_at, reason.clone()));
+                tlc.status = TlcStatus::Outbound(OutboundTlcStatus::RemoteRemoved);
```
contrun: The status `RemoteRemoved` means remote has sent us a `RemoveTLC` and `CommitmentSigned` that confirms remote's intention to remove this TLC. The caller of the function `set_offered_tlc_removed` `handle_remove_tlc_peer_message` does not have any implication on whether there is a commitment transaction from the remote party to remove this TLC.

contrun: The comment here may be wrong. It seems that `RemoteRemoved` only means remote has sent the `RemoveTlc` message (there is no implication on whether it is covered by a commitment transaction whatsoever). Only the state `RemoveWaitAck` implies that the remote sent a `CommitmentSigned` message **and** we have replied a `RevokeAndAck`. I think we have to document the meaning of each state.

chenyukang: Ok, I will add more documents for each status

src/fiber/channel.rs
```diff
@@ -656,71 +669,107 @@ where
...
+            }
+        }
+
+        // flush remove tlc for received tlcs after replying ack for peer
+        self.apply_settled_remvoe_tlcs(myself, state, true).await;
         Ok(())
     }
 
-    async fn flush_staging_tlc_operations(
+    async fn apply_settled_remvoe_tlcs(
```
quake: typo
```suggestion
    async fn apply_settled_remove_tlcs(
```

PR #423, opened 2024-12-27
Don't panic when git is not available

### LINKED ISSUE #422: Compile error when I use the release source code
The compile config file seems forcing the git enabled, however, when developer download the releasd zip file, there is not git related files.

image

contrun:
Fixes #422


contrun:
Closing in favor of #424

PR #428, opened 2024-12-30
Query missing broadcast messages from peers

contrun:


quake:
please rebase to resolve the conflicting

src/fiber/gossip.rs
```diff
@@ -330,11 +333,88 @@ pub enum GossipActorMessage {
...
+        name: Option<String>,
+        gossip_network_maintenance_interval: Duration,
+        gossip_store_maintenance_interval: Duration,
+        announce_private_addr: bool,
+        num_targeted_active_syncing_peers: Option<usize>,
+        num_targeted_outbound_passive_syncing_peers: Option<usize>,
+        store: S,
+        chain_actor: ActorRef<CkbChainMessage>,
+        supervisor: ActorCell,
+    ) -> Self {
```
quake: suggest to return (Self, GossipProtocolHandle), then we can remove the network_control_sender field and create_protocol_handle fn

quake (changes requested):

quake (approved):

PR #429, opened 2024-12-30
Make number of active/passive syncers configurable

contrun:


src/fiber/gossip.rs
```diff
@@ -1368,11 +1368,11 @@ pub enum ExtendedGossipMessageStoreMessage {
 pub(crate) struct GossipActorState<S> {
     store: ExtendedGossipMessageStore<S>,
     control: ServiceAsyncControl,
-    num_targeted_active_syncing_peers: usize,
+    num_finished_active_syncing_peers: usize,
     // The number of active syncing peers that we have finished syncing with.
     // Together with the number of currect active syncing peers, this is
     // used to determine if we should start a new active syncing peer.
-    num_finished_active_syncing_peers: usize,
+    num_targeted_active_syncing_peers: usize,
```
quake, line 1377: the field comment should be swapped after change ordering.

PR #430, opened 2025-01-01
Retry all payment session with actor message

### LINKED ISSUE #407: Graph is not updated immediately when a `TlcErrData::ChannelFailed` is sent from a peer
The processing of `TlcErrData::ChannelFailed` https://github.com/contrun/fiber/blob/c583ac8d6f5d7b7c2d7278eb80e391b66e9d0557/src/fiber/network.rs#L1487-L1495 is asynchronous. But we want to update the graph synchronously because the same payment session may be retried a few times immediately. We need to use the `ChannelUpdate` message from `TlcErrData::ChannelFailed` to update the graph.

chenyukang:
Fixes https://github.com/nervosnetwork/fiber/issues/407


src/fiber/network.rs
```diff
@@ -1680,6 +1670,12 @@ where
         }
     }
 
+    fn register_payment_retry(&self, myself: ActorRef<NetworkActorMessage>, payment_hash: Hash256) {
+        myself.send_after(Duration::from_millis(500), move || {
```
contrun, line 1781: It is possible that even after 0.5 seconds the `ChannelUpdate` contained in the error message is still not processed. This is because this message would first be saved to the gossip message store and by default the interval to save messages to the store is 20s.  https://github.com/contrun/fiber/blob/ea5c1fede9bffd8d3f3fa49d2570aadb971fcb6f/src/fiber/gossip.rs#L1179-L1181 https://github.com/contrun/fiber/blob/ea5c1fede9bffd8d3f3fa49d2570aadb971fcb6f/src/fiber/config.rs#L61-L62 .

In this case, I think we need to notify the gossip actor to start process the messages in the queue immediately (either create a new message like `TickImmediately` for the gossip actor or rename the `ProcessBroadcastMessage` to `ProcessBroadcastMessageImmediately` and change the handler to immediately process messages in the queue). Note that it is still possible that when there are too many messages in the queue this message can still not be processed after 0.5s.

chenyukang: it's 60s, not 20s ?
https://github.com/nervosnetwork/fiber/blob/ba2d2bc20ca88371f990f6b956b64ccf409eb1d1/src/fiber/config.rs#L52-L54

chenyukang: got it.

chenyukang: yeah, there is a chance the message is not processed in 0.5 second, any other better way ? I think we should not make a huge change because of this requirement.

chenyukang: https://github.com/nervosnetwork/fiber/pull/430/commits/e909f9258d47656fd2e595e300074b91aab9f013
I make this change, how do you think about it?

contrun: I prefer changing the action `ProcessBroadcastMessage` to changing the underlying data structure `ChannelUpdate`. To be specific, I think add an additional boolean parameter of https://github.com/contrun/fiber/blob/b78812b7584c290563f54d21842eafce0d33d707/src/fiber/gossip.rs#L318-L321 `ProcessBroadcastMessage` to indicate if we need to process the message immediately will be better. This boolean parameter will intoxicate a few functions/data structures, but its meaning is clearer than `ChannelUpdateImmediately`.

chenyukang: done with https://github.com/nervosnetwork/fiber/pull/430/commits/37070e29c43856885174e34c66cb44b0ed8b22c2

chenyukang: use another solution: https://github.com/nervosnetwork/fiber/pull/430/commits/513ad0aa55d0b2a2743641bd2f417b45d3838375

contrun (approved):

PR #432, opened 2025-01-02
Add custom hop hints and fix self payment

### LINKED ISSUE #355: self-payment fails when both channels have zero remote_balance
When both channels A(100)->B(0) and B(100)->A(0) have a remote_balance of 0, self-payment fails with the error message: Send payment error: Failed to send onion packet with error IncorrectOrUnknownPaymentDetails.

It looks like both the `remote_balance` and `local_balance` of these two channels need to be sufficient for a successful self-payment, which seems illogical.

After many tests, this problem still occurs.

ref: https://github.com/nervosnetwork/fiber/issues/300#issuecomment-2484881011

### LINKED ISSUE #366: Cleanup channel RPC
The RPC like `add_tlc` and `remove_tlc` should be moved into integrating test module, it's not for endpoint usage.

chenyukang:
1. Move `add_tlc`, `remove_tlc`, `commitment_signed` to dev RPC module, fixes #366
2. Add hop hints for routing algorithm 
3. Use hop hint for self payment to handle more complex scenario, add test case for self pay, fixes #355


src/fiber/graph.rs
```diff
@@ -828,6 +825,33 @@ where
...
+            .collect::<HashMap<_, _>>();
+
+        let mut target = target;
+        let mut current_expiry = final_tlc_expiry_delta;
+        let mut last_edge = None;
+
+        if route_to_self {
+            // if there is no hint, we will randomly select a channel from the source node for route to self
+            // so that the following part of algorithm will always trying to find a path without cycle
+            let (new_taget, expiry, edge) =
```
quake: typo
```suggestion
            let (new_target, expiry, edge) =
```

src/rpc/README.md
```diff
@@ -524,6 +471,72 @@ Cancels an invoice, only when invoice is in status `Open` can be canceled.
...
+ suppose the payment router is with N hops, the total tlc expiry limit is at least (N-1) days
+ this is also the default value for the payment if this parameter is not provided
+* `invoice` - `Option<String>`, the encoded invoice to send to the recipient
+* `timeout` - `Option<u64>`, the payment timeout in seconds, if the payment is not completed within this time, it will be cancelled
+* `max_fee_amount` - `Option<u128>`, the maximum fee amounts in shannons that the sender is willing to pay
+* `max_parts` - `Option<u64>`, max parts for the payment, only used for multi-part payments
+* `keysend` - `Option<bool>`, keysend payment
+* `udt_type_script` - `Option<Script>`, udt type script for the payment
+* `allow_self_payment` - `Option<bool>`, allow self payment, default is false
+* `hop_hints` - `Option<Vec>`, Optional route hints to reach the destination through private channels.
```
quake, line 500: the generated doc should be `Option<Vec<HopHint>>`, may be generator bug?

src/fiber/graph.rs
```diff
@@ -828,6 +825,33 @@ where
             ));
         }
 
+        let hop_hint_map: HashMap<Pubkey, (OutPoint, bool)> = hop_hints
```
quake: I think the map type should be `HashMap<(Pubkey, bool), OutPoint>`, otherwise this case will go wrong:
```
we want to rebalance the channels 1 and 4, and provide two hints (c1/c4) for the payment
A -- Channel 1 --> B
A -- Channel 2 --> B
B -- Channel 3 --> A
B -- Channel 4 --> A
``` 

could you add an unit test for this case?

quake (changes requested):

src/rpc/README.md
chenyukang: yes, I will fix it in a separate pr.

src/fiber/graph.rs
chenyukang: fixed with commit https://github.com/nervosnetwork/fiber/pull/432/commits/d8c824bbb91615bb1b1ae397054b99734fb1c8fa

src/fiber/graph.rs
```diff
@@ -1005,10 +1030,70 @@ where
...
+        if direct_channels.is_empty() {
+            return Err(PathFindError::PathFind(
+                "no direct channel found for source node".to_string(),
+            ));
+        }
+        // a proper hop hint for route self will limit the direct_channels to only one
+        // if there are multiple channels, we will randomly select a channel from the source node for route to self
+        // so that the following part of algorithm will always trying to find a path without cycle
+        let rand_index = rand::thread_rng().gen_range(0..direct_channels.len());
+        let (from, _, channel_info, channel_update) = direct_channels[rand_index];
```
quake: suggest to use `SliceRandom::choose`
```suggestion
        if let Some((from, _, channel_info, channel_update)) = direct_channels.choose(&mut thread_rng()) {

        } else {
            return Err(PathFindError::PathFind(
                "no direct channel found for source node".to_string(),
            ));
        }
```

src/fiber/graph.rs
```diff
@@ -1005,10 +1031,74 @@ where
...
+            assert_ne!(target, from);
+            let target = from;
+            Ok((target, current_expiry, last_edge))
+        } else {
+            return Err(PathFindError::PathFind(
+                "no direct channel found for source node".to_string(),
+            ));
+        }
+        // let rand_index = rand::thread_rng().gen_range(0..direct_channels.len());
+        // let (from, _, channel_info, channel_update) = direct_channels[rand_index];
```
quake: please cleanup code

quake (approved):

src/fiber/graph.rs
chenyukang: fixed up, thanks.

PR #433, opened 2025-01-02
Avoid send broadcast message when shutdown private channel

### LINKED ISSUE #431: [open_channel] shut_down private channel will cause panic `public channel can generate channel update message`
1. open_channel(public:False)
2. shutdown_channel

err
```
thread 'tokio-runtime-worker' panicked at /Users/guopenglin/demo/tmp/fiber/src/fiber/channel.rs:3375:14:
public channel can generate channel update message
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
[node.log](https://github.com/user-attachments/files/18285678/node.log)

chenyukang:
Fixes https://github.com/nervosnetwork/fiber/issues/431
unresolved issue: https://github.com/nervosnetwork/fiber/issues/431#issuecomment-2567286109

### DIFF
--- a/src/fiber/channel.rs
+++ b/src/fiber/channel.rs
@@ -1760,16 +1760,18 @@ where
             }
             ChannelEvent::ClosingTransactionConfirmed => {
                 // Broadcast the channel update message which disables the channel.
-                let update = state.generate_disabled_channel_update(&self.network).await;
+                if state.is_public() {
+                    let update = state.generate_disabled_channel_update(&self.network).await;
 
-                self.network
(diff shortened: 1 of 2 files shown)


contrun (approved):

contrun:
You can cherry pick the commit https://github.com/nervosnetwork/fiber/pull/435/commits/b64a3a2fdce95291b8c151468a3b82f88978689e in https://github.com/nervosnetwork/fiber/pull/435 to solve the problem of the program not exiting even if the actor panicked. That PR can be closed after the cherry-picking.

PR #434, opened 2025-01-02
Save our own broadcast message directly and save raw gossip message from peers to memory first

contrun:
While verifying gossip messages, the first thing we need to do is to obtain the `ChannelAnnouncement` timestamp. This operation requires interaction with the CKB chain. As an implementation detail, we need to obtain the whole transaction for the timestamp in the block header. So it is possible that we will call the chain actor a few times in order to process the gossip message. This PR attempts to cache the `ChannelAnnouncement` timestamp and the channel transaction in memory to reduce the interaction with the CKB chain. It also fixes the two problems in the comment of https://github.com/nervosnetwork/fiber/pull/414

### DIFF
--- a/src/ckb/actor.rs
+++ b/src/ckb/actor.rs
@@ -41,19 +41,18 @@ impl TraceTxResponse {
     }
 }
 
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub enum GetBlockTimestampRequest {
-    BlockNumber(u64),
-    BlockHash(H256),
+#[derive(Debug, Clone)]
+pub struct GetBlockTimestampRequest {
+    block_hash: H256,
 }
 
 impl GetBlockTimestampRequest {
     pub fn from_block_hash(block_hash: H256) -> Self {
-        Self::BlockHash(block_hash)
+        Self { block_hash }
     }
 
-    pub fn from_block_number(block_number: u64) -> Self {
-        Self::BlockNumber(block_number)
+    pub fn block_hash(&self) -> H256 {
+        self.block_hash.clone()
     }
 }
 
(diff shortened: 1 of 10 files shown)


src/fiber/gossip.rs
```diff
@@ -1000,45 +1156,48 @@ impl<S: GossipMessageStore> ExtendedGossipMessageStoreState<S> {
...
+            .messages_to_be_saved
+            .iter()
+            .filter(|m| self.has_dependencies_available(m))
+            .cloned()
+            .collect::<HashSet<_>>();
+        self.messages_to_be_saved = self
+            .messages_to_be_saved
+            .difference(&complete_messages)
+            .cloned()
+            .collect();
```
quake:
```suggestion
        let (complete_messages, uncomplete_messages) = self.messages_to_be_saved.clone()
            .into_iter()
            .partition(|m| self.has_dependencies_available(m));
        self.messages_to_be_saved = uncomplete_messages;
```

src/fiber/network.rs
```diff
@@ -2673,30 +2682,72 @@ where
...
-                        block_number.into(),
-                        DUMMY_FUNDING_TX_INDEX,
-                    )
+                    tokio::spawn( async move  {
+                        const MAX_GET_BLOCK_TIMESTAMP_RETRY: u8 = 5;
+                        for _ in 0..MAX_GET_BLOCK_TIMESTAMP_RETRY {
+                            match call!(
+                                chain,
+                                |reply| CkbChainMessage::GetBlockTimestamp(
+                                    GetBlockTimestampRequest::from_block_number(block_number.into()), reply
```
quake: TxStatus has a `block_hash` field, it's better to use the hash to fetch timestamp.

contrun: Do we also need to remove all fields with type block_number?

quake: I think so

src/ckb/actor.rs
```diff
@@ -41,18 +41,19 @@ impl TraceTxResponse {
     }
 }
 
-#[derive(Debug, Clone)]
-pub struct GetBlockTimestampRequest {
-    block_hash: H256,
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub enum GetBlockTimestampRequest {
+    BlockNumber(u64),
```
quake: this change should be reverted also

PR #440, opened 2025-01-02
Add retryable forwarding tlc 

chenyukang:
1. Add retryable forwarding tlc
2. Add more bench test cases for send payment
3. Fix deadlock issues in channel actor, add ForwardTlcResult for retrying TLC forwarding.


src/fiber/channel.rs
```diff
@@ -1435,49 +1468,80 @@ where
...
+                    ) {
+                        Ok(_) | Err(ProcessingChannelError::RepeatedProcessing(_)) => false,
+                        Err(ProcessingChannelError::WaitingTlcAck) => true,
+                        Err(_err) => false,
                     }
                 }
-                RetryableRemoveTlc::RelayRemoveTlc(channel_id, tlc_id, ref reason) => {
-                    let (send, recv) = oneshot::channel::<Result<(), String>>();
+                RetryableTlcOperation::RelayRemoveTlc(channel_id, tlc_id, ref reason) => {
+                    // send replay remove tlc with network actor to previous hop
```
contrun: Typo: replay -> relay

src/fiber/channel.rs
```diff
@@ -1493,37 +1557,114 @@ where
...
+                    try_one_time,
+                ) => {
+                    // there is a potential deadlock for waiting the result from another channel actor
+                    // for the scenario these two things happen at the same time:
+                    //  1. channel A send forward tlc to channel B
+                    //  2. channel B send forward tlc to channel A
+                    // we may end up waiting for each other forever
+                    //
+                    // but we need the result for better error handling
+                    // so we introduce the ForwardTlcResult to get the resule based on actor message
```
contrun: Typo: resule -> result

src/fiber/channel.rs
```diff
@@ -1435,49 +1468,80 @@ where
...
+    }
+
+    pub async fn apply_retryable_tlc_operations(
+        &self,
+        myself: &ActorRef<ChannelActorMessage>,
+        state: &mut ChannelActorState,
+    ) {
+        let pending_tlc_ops = state.tlc_state.get_pending_operations();
+        for retryable_operation in pending_tlc_ops.into_iter() {
+            let keep = match retryable_operation {
```
contrun: It seems to me that the function of this code is to take out the tlc operations from the vector `retryable_tlc_operations`, apply these operations, then remove the applied operations (depending on the result).

I think using `self.retryable_tlc_operations.retain_mut(|tlc_operation| apply_tlc_operation(tlc_operation))` will greatly simplify the code here. You can also change the state of tlc operation while trying to `apply_tlc_operation` as `retain_mut` https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain_mut will pass a mutable reference. So the function `set_forward_tlc_status` can be redundant.

Also, TLC Operations can be viewed as a state machine. I think explicitly refactoring the code to state machine transition will help the readability of the code here.

chenyukang: seems ok
https://github.com/nervosnetwork/fiber/pull/440/commits/818a8ae5cbda25a8b050de5cde92e1d0d6713ed7

PR #444, opened 2025-01-03
chore: tweak gossip related log level

quake:
gossip protocol outputs a lot of unnecessary logs, this PR cleans up the logs.

### DIFF
--- a/src/fiber/gossip.rs
+++ b/src/fiber/gossip.rs
@@ -465,10 +465,6 @@ where
         match message {
             GossipSyncingActorMessage::RequestTimeout(request_id) => {
                 state.inflight_requests.remove(&request_id);
-                debug!(
-                    "Sending new GetBroadcastMessages request after timeout: id {}",
-                    request_id
-                );
                 // TODO: When the peer failed for too many times, we should consider disconnecting from the peer.
                 state.peer_state.failed_times += 1;
                 myself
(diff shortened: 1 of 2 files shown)


src/fiber/gossip.rs
```diff
@@ -2293,28 +2230,16 @@ where
...
                 state
                     .peer_states
                     .insert(peer_id.clone(), PeerState::new(session.id, session.ty));
             }
-            GossipActorMessage::PeerDisconnected(peer_id, session) => {
-                debug!(
-                    "Peer disconnected: peer {:?}, session {:?}",
-                    &peer_id, &session.id
-                );
-                drop(state.peer_states.remove(&peer_id));
```
contrun, line 2317: The drop here is useful, as that would stop the filter processor actor.

quake: why we need to explicitly call drop to release data removed from a HashMap?

contrun: This is used to drop the value saved in the HashMap. The value of type `PeerState` https://github.com/contrun/fiber/blob/b65b586f27ea79d6ea0f1cd677b64b693e12ce60/src/fiber/gossip.rs#L778-L805 maintains a few things.

quake: I don't get you, letting the removed value go out of scope is sufficient, and rust will handle memory cleanup automatically, could you please tell why this change will prevent the removed `PeerState` from dropping?
```diff
-                drop(state.peer_states.remove(&peer_id));
+                state.peer_states.remove(&peer_id); 
```

contrun: Sorry. I missed the following line. You're right.

PR #455, opened 2025-01-07
chore: upgrade ractor to new version

chenyukang:


src/actors.rs
```diff
@@ -46,10 +46,14 @@ impl Actor for RootActor {
 
     async fn post_stop(
         &self,
-        _myself: ActorRef<Self::Msg>,
+        myself: ActorRef<Self::Msg>,
         _state: &mut Self::State,
     ) -> Result<(), ActorProcessingErr> {
         debug!("Root actor stopped");
+        myself
```
contrun, line 53: The stopping of sub actors is not done automatically. We must explicitly call https://docs.rs/ractor/latest/ractor/actor/actor_cell/struct.ActorCell.html#method.stop_children_and_wait . Without https://github.com/nervosnetwork/fiber/pull/455/commits/5b1cc49215bebdc68b52481f3c2eca9011c64076, running `clear; cargo run -- -d ./tests/nodes/1` and press ctrl-c will not gracefully stop fnn.

contrun (approved):

quake (approved):

PR #459, opened 2025-01-09
Change service error log level

contrun:
Some important connection error message will not be shown to users if we are using `trace!` by default. For example, even the user specified an invalid boot node address, he/she may not know that there is an error while connecting to boot node.


quake:
user may monitor error/warn level log to raise an alert, when a node is unreachable or ip changed, it will raise many false alarms

PR #420, opened 2024-12-25
Unify various cryptography functions

contrun:

PR #468, opened 2025-01-12
Add migration from v0.2.0 to v0.2.1

contrun:
There is a breaking change [from v0.2.0 to v0.2.1](https://github.com/nervosnetwork/fiber/compare/v0.2.0...v0.2.1), which was introduced in https://github.com/nervosnetwork/fiber/pull/420. This PR tries to migrate data created from v0.2.0 to v0.2.1. The approach used in this PR is not perfect. There are 2 notable defects.

1. The `ChannelActorState` contains many references to the data structures defined in other places. I didn't duplicate all the data definitions (that requires a lot of manual copy and pastes). So if the dependent data structures change in the future, then the migration code presented here may break. For example, if we add additional field to [`ChannelConstraints`](https://github.com/contrun/fiber/blob/5a2fe087e2d0adde48db2e45515d9f2fd569c924/src/fiber/channel.rs#L2881-L2887) (say in the version 0.5.0 we added a new field), then the deserialization of `OldChannelActorState` and the serialization of `NewChannelActorState` in this code base will no longer work. This means that if the the user wants to upgrade data from v0.2.0 to v0.5.0, then he/she may have to first upgrade fnn from 0.2.0 to 0.2.2 (assume 0.2.2 contains this PR), and then to 0.5.0.
2. As noted in the comment, we can't determine which remote nonce we should set for the field `last_revoke_and_ack_remote_nonce` without the counterparty's help. This is because we don't know if the counterparty received our latest `RevokeAndAck` or not, and we didn't save the remote nonce we used while creating the `RevokeAndAck` message. This PR assumes the happy path, i.e. the counterparty received our latest `RevokeAndAck`. If this is not the case, the user has to force close the channel. The root cause of this problem is that which data to save in the `NewChannelActorState` depends on the peer's state, and we have no way to interact with the peer in the migration stage.

I have verified that the code here works by the following steps.

1. Run `git checkout v0.2.0` to checkout to the v0.2.0 version code.
2. Run `REMOVE_OLD_STATE=y ./tests/nodes/start.sh` to start a few fnn nodes with all new states.
3. Run `git checkout v0.2.1-migration-scripts` to change the code base to branch [v0.2.1-migration-scripts](https://github.com/contrun/fiber/tree/v0.2.1-migration-scripts) with some additional scripts to create channel.
4. Run `(cd ./tests/bruno; npm exec -- @usebruno/cli@1.20.0 run e2e/before-migration -r --env test)` to create a channel.
5. Stop the nodes started in step 2.
6. Run `cargo run -- -d ./tests/nodes/1 --migrate` and `cargo run -- -d ./tests/nodes/3 --migrate` to migrate (we need to type YES in the terminal prompt).
7. Run `./tests/nodes/start.sh` to start a few fnn nodes to start fnn nodes again.
8. Run `(cd ./tests/bruno; npm exec -- @usebruno/cli@1.20.0 run e2e/after-migration -r --env test)` to verify that we can make progress after the migration.

### DIFF
--- a/src/fiber/channel.rs
+++ b/src/fiber/channel.rs
@@ -3016,6 +3016,15 @@ pub struct ChannelActorState {
     #[serde_as(as = "Option<PubNonceAsBytes>")]
     pub last_committed_remote_nonce: Option<PubNonce>,
 
+    // While building a CommitmentSigned message, we use the latest remote nonce (the `last_committed_remote_nonce` above)
+    // to partially sign the commitment transaction. This nonce is also needed for the RevokeAndAck message
+    // returned from the peer. We need to save this nonce because the counterparty may send other nonces during
+    // the period when our CommitmentSigned is sent and the counterparty's RevokeAndAck is received.
+    // This field is used to keep the nonce used by the unconfirmed CommitmentSigned. When we receive a
+    // RevokeAndAck from the peer, we will use this nonce to validate the RevokeAndAck message.
+    #[serde_as(as = "Option<PubNonceAsBytes>")]
+    pub last_commitment_signed_remote_nonce: Option<PubNonce>,
+
     // While handling peer's CommitmentSigned message, we will build a RevokeAndAck message,
     // and reply this message to the peer. The nonce used to build the RevokeAndAck message is
     // an older one sent by the peer. We will read this nonce from the field `last_committed_remote_nonce`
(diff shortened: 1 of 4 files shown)


src/fiber/channel.rs
```diff
@@ -3016,6 +3016,15 @@ pub struct ChannelActorState {
...
 
+    // While building a CommitmentSigned message, we use the latest remote nonce (the `last_committed_remote_nonce` above)
+    // to partially sign the commitment transaction. This nonce is also needed for the RevokeAndAck message
+    // returned from the peer. We need to save this nonce because the counterparty may send other nonces during
+    // the period when our CommitmentSigned is sent and the counterparty's RevokeAndAck is received.
+    // This field is used to keep the nonce used by the unconfirmed CommitmentSigned. When we receive a
+    // RevokeAndAck from the peer, we will use this nonce to validate the RevokeAndAck message.
+    #[serde_as(as = "Option<PubNonceAsBytes>")]
+    pub last_commitment_signed_remote_nonce: Option<PubNonce>,
+
```
chenyukang, line 3027: why do you change the order here?

chenyukang:
>  This means that if the the user wants to upgrade data from v0.2.0 to v0.5.0, then he/she may have to first upgrade fnn from 0.2.0 to 0.2.2 (assume 0.2.2 contains this PR), and then to 0.5.0.

This is an assumption we need to hold, the binary of v0.5.0 must contain the code for upgrading from 0.2.0 to 0.2.2(if there is such an upgrading), and the migration will be executed by order.

@quake

chenyukang:
> As noted in the comment, we can't determine which remote nonce we should set for the field last_revoke_and_ack_remote_nonce without the counterparty's help. This is because we don't know if the counterparty received our latest RevokeAndAck or not, and we didn't save the remote nonce we used while creating the RevokeAndAck message. This PR assumes the happy path, i.e. the counterparty received our latest RevokeAndAck. If this is not the case, the user has to force close the channel. The root cause of this problem is that which data to save in the NewChannelActorState depends on the peer's state, and we have no way to interact with the peer in the migration stage.

This is the most challenge scenario may happened during migration, sometimes it's hard to determine the initial value of new fields. 

It's better to consider this when writing the new code, I'm thinking maybe we need a way to detect the necessary of migration, then we can add a check in CI, for this case, the code change in storage tests code maybe a good signal:

https://github.com/nervosnetwork/fiber/compare/v0.2.0...v0.2.1#diff-c8ce48166948578ad1a1ffbb233f3d0c04e43d3a528f59e73437e1e2aefbc869L357-L359

src/fiber/channel.rs
contrun: Because this is the right order. Previous comments were actually placed near to the wrong fields.

PR #471, opened 2025-01-13
Minimize work done by `build_route`

contrun:
Previously, we have some duplicated logic in `find_path` and `build_route`, which has two problems.

1. When building the complete data that can be used to generate relevant `AddTlc`s, we need to resort to `channels` in the graph to obtain the `fee_rate` and `expiry` info from `channels` field of graph. This makes it harder to use hints for path (users can pass this from invoices) that do not exist in the `channels` field of the graph.
2. Some subtle bugs went unnoticed. The unit test `test_graph_find_path_fee` actually succeeded only because we select the channel with newer timestamp.

This PR tries to minimize the work done in `build_route`, move most of its logic to `find_path` and returns all relevant information to `build_route`.


src/fiber/graph.rs
```diff
@@ -235,10 +225,20 @@ pub enum PathFindError {
     Other(String),
 }
 
-#[derive(Clone, Debug)]
+// An edge along the payment path from the source to the target.
+// This represents a TLC transfer from one node to another.
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct PathEdge {
+    pub source: Pubkey,
```
chenyukang: why do we need the source?
I mean the first hop must be the source, and if we know the target, we know the next hop, then next->next->source will be next->target, seems we don't need the `source` feild in this structure?

src/fiber/graph.rs
```diff
@@ -837,34 +798,43 @@ where
...
-            last_edge = edge;
-            current_expiry += expiry;
+            let (edge, expiry) = self.adjust_target_for_route_self(
+                &hop_hint_map,
+                amount,
+                final_tlc_expiry_delta,
+                source,
+                target,
+            )?;
+            target = edge.source;
```
chenyukang: can we add a assertion `assert_ne!(target, edge.source)` to make sure we pick a different target after `adjust_target_for_route_self`.

src/fiber/tests/graph.rs
```diff
@@ -364,23 +371,52 @@ fn test_graph_find_path_fee() {
     let mut network = MockNetworkGraph::new(5);
 
     network.add_edge(1, 2, Some(1000), Some(10000));
+    // means node 2 will charge fee_rate 10000 when forwarding tlc
```
chenyukang, line 382: please add some comments for `add_edge`, specific for the `fee_rate`, which maybe confused in future.

chenyukang (approved):

quake (approved):

PR #489, opened 2025-01-17
Use the fee rate in oubound channel to calculate the TLC fee

contrun:


src/fiber/graph.rs
```diff
@@ -755,8 +753,8 @@ where
             |(_, _, a_channel_info, a_channel_update_info),
              (_, _, b_channel_info, b_channel_update_info)| {
                 b_channel_update_info
-                    .inbound_liquidity
-                    .cmp(&a_channel_update_info.inbound_liquidity)
+                    .outbound_liquidity
```
quake, line 756: the code comment of `inbound_liquidity` should be renamed to `outbound_liquidity` also

src/fiber/network.rs
```diff
@@ -506,7 +506,9 @@ pub struct AcceptChannelCommand {
 #[derive(Debug, Clone)]
 pub struct SendOnionPacketCommand {
     pub peeled_onion_packet: PeeledPaymentOnionPacket,
-    pub previous_tlc: Option<(Hash256, u64)>,
+    // We are currently forwarding a previous tlc. The previous tlc's channel id, tlc id
+    // and the fee paid are included here.
+    pub previous_tlc: Option<(Hash256, u64, u128)>,
```
quake: there are multiple (u64, u128) tuples in the code, some are expiry_delta/fee, others are tlc_id/fee, to make the code easier to understand and maintain it is recommended to change it to a different struct.

src/fiber/channel.rs
```diff
@@ -178,7 +178,7 @@ pub struct AddTlcCommand {
     /// Save it for outbound (offered) TLC to backward errors.
     /// Use all zeros when no shared secrets are available.
     pub shared_secret: [u8; 32],
-    pub previous_tlc: Option<(Hash256, u64)>,
+    pub previous_tlc: Option<(Hash256, u64, u128)>,
 }
```
chenyukang, line 182: how to do migration for this change?

quake: we don't save the `AddTlcCommand` in storage, I don't think we need migration for it.

src/fiber/network.rs
contrun: Except for the `previous_tlc` field in `TlcInfo` https://github.com/contrun/fiber/blob/872598317c75190848d6056ed7529a94e1216722/src/fiber/channel.rs#L2507, other previous tlcs are now using a dedicated struct. I didn't touch `TlcInfo` because this is saved in `ChannelActorState`. Changing this will require a migration.

gpBlockchain:
LGTM

quake (approved):

PR #501, opened 2025-01-24
Add hold invoice support

contrun:


contrun:
The code here is included in https://github.com/nervosnetwork/fiber/pull/511

PR #511, opened 2025-02-17
Support fiber multiple hops payment in cross chain transaction

contrun:


src/ckb/config.rs
```diff
@@ -170,6 +170,16 @@ pub struct UdtScript {
     pub args: String,
 }
 
+impl UdtScript {
+    pub fn allow_all_for_script(script: &Script) -> Self {
```
quake, line 174: I think it's only used in unit test code, suggest to move it to the test utils, and it should be a helper fn.

src/fiber/channel.rs
```diff
@@ -840,6 +829,28 @@ where
...
+        let tlc_id = match state.get_received_tlc_with_hash(hash) {
+            Some(tlc) => tlc.tlc_id,
+            None => {
+                return;
+            }
+        };
+        // Only settle down this TLC if it is not already settled down.
+        if state.tlc_state.applied_add_tlcs.insert(tlc_id) {
+            self.try_to_settle_down_tlc(myself, state, tlc_id).await;
+        }
```
quake:
```suggestion
        if let Some(tlc) = match state.get_received_tlc_with_hash(hash) {
        let tlc_id = tlc.tlc_id;
        if state.tlc_state.applied_add_tlcs.insert(tlc_id) {
            self.try_to_settle_down_tlc(myself, state, tlc_id).await;
        }
        }
```

src/invoice/store.rs
```diff
@@ -20,4 +31,105 @@ pub trait InvoiceStore {
...
+    InvoiceNotFound,
+    #[error("Hash mismatch")]
+    HashMismatch,
+    #[error("Internal error: {0}")]
+    InternalError(String),
+}
+
+// This function is used to settle an invoice. It is extracted from the `InvoiceRpcServerImpl` struct
+// to share it with test functions.
+pub(crate) fn settle_invoice<S: InvoiceStore>(
```
quake: I don't think extract function is necessary, it introduces unnecessary ChannelCommand/NetworkActorCommand/NetworkActorMessage dependencies to the store module.

src/store/subscription_impl.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    pub fn send_update(&self, update: PaymentUpdate) -> bool {
+        self.receiver.send_message(update).is_ok()
+    }
+}
+
+#[derive(Default)]
+pub struct SubscriptionActorState {
+    next_subscriber_id: SubscriptionId,
+    invoice_subscriptions: HashMap<Hash256, Vec<InvoiceSubscriber>>,
```
quake, line 111: I think a HashMap is simpler:
```suggestion
    invoice_subscriptions: HashMap<Hash256, HashMap<SubscriptionId, DerivedActorRef<InvoiceUpdate>>>,
```

tests/nodes/deployer/config.yml
```diff
@@ -23,7 +23,8 @@ rpc:
 
 cch:
   ignore_startup_failure: true
-  wrapped_btc_type_script_args: "0x32e555f3ff8e135cece1351a6a2971518392c1e30375c1e006ad0ce8eac07947"
+  # will be generated by udt-init
+  wrapped_btc_type_script: 0x55000000100000003000000031000000e1e354d6d643ad42724d40967e334984534e0367405c5ae42a9d7d63d77df419022000000032e555f3ff8e135cece1351a6a2971518392c1e30375c1e006ad0ce8eac07947
```
quake, line 27: we may use ckb json type `Script` here, it's easy to read than a hex string:
```suggestion
  wrapped_btc_type_script:
      code_hash: 0xe1e354d6d643ad42724d40967e334984534e0367405c5ae42a9d7d63d77df419
      hash_type: Data1
      args: 0x....
```

src/rpc/README.md
```diff
@@ -566,6 +579,9 @@ Sends a payment to a peer.
 * `max_fee_amount` - <em>`Option<u128>`</em>, the maximum fee amounts in shannons that the sender is willing to pay
 * `max_parts` - <em>`Option<u64>`</em>, max parts for the payment, only used for multi-part payments
 * `keysend` - <em>`Option<bool>`</em>, keysend payment
+* `hold_payment` - <em>`Option<bool>`</em>, a hold_payment is a payment with the preimage set to an all zero hash
```
quake: Besides hold invoice payments, what other scenarios require that the hold_payment parameter be set to true? Is it possible to remove this parameter from the rpc interface?

src/store/store.rs
```diff
@@ -24,11 +27,20 @@ use serde::Serialize;
 use std::{path::Path, sync::Arc};
 use tentacle::secio::PeerId;
 
+/// A generic store that allows for different types of hooks to be called when some data are updated.
+/// An example of a hook is the [SubscriptionImpl] which is used to notify subscribers of updates of
+/// invoices and payments. Another example is the [NoopStoreUpdateHook] which does nothing.
 #[derive(Clone, Debug)]
-pub struct Store {
+pub struct GenericStore<IH: InvoiceUpdateHook, PH: PaymentUpdateHook> {
```
quake, line 34: There's no need to make a `GenericStore` abstraction of the store here, it violates the principle of composition over inheritance, you can just use a store + hook struct wherever you need a store callback.

quake (changes requested):

src/rpc/README.md
contrun: Done in https://github.com/nervosnetwork/fiber/pull/511/commits/548259b0d1c51febfcbb39443355482c2aed4acd and https://github.com/nervosnetwork/fiber/pull/511/commits/a0448747056ebf9795c238ea255d305aad2a679c

tests/nodes/deployer/config.yml
contrun: wrapped_btc_type_script can also be specified with environment variable `CCH_WRAPPED_BTC_TYPE_SCRIPT`. It is not very easy to write nested structures with environment variable. Can we both do a hex decoding and a `ckb_json_types::Script` decoding? That makes both `0x55000000100000003000000031000000e1e354d6d643ad42724d40967e334984534e0367405c5ae42a9d7d63d77df419022000000032e555f3ff8e135cece1351a6a2971518392c1e30375c1e006ad0ce8eac07947` and `code_hash: 0xe1e354d6d643ad42724d40967e334984534e0367405c5ae42a9d7d63d77df419
      hash_type: Data1
      args: 0x....` legitimate.

src/ckb/config.rs
contrun: Done in https://github.com/nervosnetwork/fiber/pull/511/commits/dac7a4185a082d2c5ef59cd6ad4b00ae47f9bc72

src/fiber/channel.rs
contrun: Done in https://github.com/nervosnetwork/fiber/pull/511/commits/dac7a4185a082d2c5ef59cd6ad4b00ae47f9bc72

src/invoice/store.rs
contrun: Done in https://github.com/nervosnetwork/fiber/pull/511/commits/c53f6dfebba73ae7f9b3b75622abd7e8132eb3b6

src/store/store.rs
contrun: Do you mean that the hooks can be passed to every one that needs them (instead of bundling them with the store)? There are many places in current code base which alters the status of an invoice, for example, https://github.com/contrun/fiber/blob/c53f6dfebba73ae7f9b3b75622abd7e8132eb3b6/src/rpc/invoice.rs#L340-L348  https://github.com/contrun/fiber/blob/c53f6dfebba73ae7f9b3b75622abd7e8132eb3b6/src/fiber/channel.rs#L1058-L1062 https://github.com/contrun/fiber/blob/c53f6dfebba73ae7f9b3b75622abd7e8132eb3b6/src/store/store.rs#L444-L447 , all the code listed above must have a reference to the `InvoiceUpdateHook`. So it seems to me define a
```rust
Struct StoreWithHook<IH: InvoiceUpdateHook> {
   store: Store,
   hook: IH,
}
```
and pass this `StoreWithHook` to above code would make life easier. That is essentially `GenericStore<IH: InvoiceUpdateHook>`. 

The thing is that Hook and Store can't really work work independently. Hook (calling it StoreUpdateHook is better) should be part of Store because the hooks run when store is updated. We can't compose these two inter-dependent component.

quake: keep the original store implementation unchanged:
```rust
impl InvoiceStore for Store {
    fn insert_invoice(...) {
        ...
    }
}
```

and add a compsited store implementation:
```rust
struct InvoiceStoreWithHook {
    store: Store,
    invoice_hook: ConcreteInvoiceHook,
}

impl InvoiceStore for InvoiceStoreWithHook {
    fn insert_invoice(...) {
        self.store.insert_invoice(...);
        self.invoice_hook.on_invoice_updated(...);
    }
}
```

no need to introduce GenericStore struct and InvoiceUpdateHook trait

src/fiber/tests/test_utils.rs
```diff
@@ -272,6 +345,120 @@ impl NetworkNodeConfigBuilder {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn establish_udt_channel_between_nodes(
+    node_a: &mut NetworkNode,
```
chenyukang: can we merge this with `establish_channel_between_nodes`? seems only have a extra `udt_script`.

src/fiber/channel.rs
```diff
@@ -1006,33 +1007,65 @@ where
...
+                            let status = self
+                                .store
+                                .get_invoice_status(&payment_hash)
+                                // The sender sent a TLC with no invoice associated with it.
+                                .ok_or(ProcessingChannelError::FinalIncorrectPaymentHash)?;
+                            let is_active = status == CkbInvoiceStatus::Open
+                                || status == CkbInvoiceStatus::Received;
+                            let is_settled =
+                                self.store.get_invoice_preimage(&payment_hash).is_some();
+                            if is_active && !is_settled {
```
chenyukang: the `is_settled` could be removed, since if `is_some` return true the code will goto `Some(preimage)` branch and will never reach here.

src/fiber/channel.rs
```diff
@@ -1006,33 +1007,65 @@ where
...
+                                // This TLC is added to applied_add_tlcs in above, but
+                                // TLCs in the list applied_add_tlcs wouldn't be processed again.
+                                // For the unsettled active hold invoice TLCs, we should process them indefinitely
+                                // until they expire or are settled.
+                                state.tlc_state.applied_add_tlcs.remove(&add_tlc.tlc_id);
```
chenyukang, line 1027: `applied_add_tlcs` is introduced to resolve some concurrency issue, I'm not sure is it ok to remove it from here.

src/fiber/channel.rs
```diff
@@ -1006,33 +1007,65 @@ where
...
-            } else {
-                return Err(ProcessingChannelError::FinalIncorrectPaymentHash);
+                    .update_invoice_status(&payment_hash, CkbInvoiceStatus::Received)
+                    .expect("update invoice status failed");
             }
+            self.store
+                .insert_payment_preimage(payment_hash, preimage)
+                .map_err(|_| {
+                    ProcessingChannelError::InternalError("insert preimage failed".to_string())
+                })?;
```
chenyukang: seems `ProcessingChannelError::FinalIncorrectPaymentHash` does not return to caller. we should return the error at line 1019?

src/rpc/invoice.rs
```diff
@@ -324,4 +359,43 @@ where
...
+
+        let message = move |rpc_reply| -> NetworkActorMessage {
+            NetworkActorMessage::Command(NetworkActorCommand::SettleInvoice(
+                *payment_hash,
+                *payment_preimage,
+                rpc_reply,
+            ))
+        };
+
+        match call_t!(network_actor, message, RPC_TIMEOUT_MS).map_err(|ractor_error| {
```
chenyukang: maybe can reuse macro handle_actor_call like rpc/channel.rs.

src/store/store.rs
contrun: In https://github.com/contrun/fiber/tree/support-fiber-multiple-hops-payment-in-cross-chain-transaction-pass-hooks, I tried to use this approach (as I understood it). Can you confirm this align well with the approach you described? My code in that branch is very complicated, I don't know if that is the simplest way to accomplish what's on your mind. The complexity of my implementation is due to the following points:

1. Hooks has to be initialized in the main function (we need to have a singleton to receive all the updates and push them to the subscribers).
2. There are many modules which use `InvoiceStore`, so all of they have to call `InvoiceStoreWithHooks::new`, which is a function to initialize a `InvoiceStore` with a hook. This means we have to pass the invoice hooks created in the main function down to many modules.
3. Channel actors use `ChannelActorStateStore` along with `InvoiceStore`. So we have to implement `ChannelActorStateStore` for the `InvoiceStoreWithHooks`.
4. The same thing needs to be done for `NetworkGraphStateStore` (yeah, confusingly, this trait updates payment session), because we need to inject hooks for payments.

I didn't have a complete implementation yet. If my current code is what's on your mind. I can finish injecting payment hooks and removing the `GenericStore`.

src/fiber/tests/test_utils.rs
contrun: I didn't do this because there are some many places in the code base calling `establish_channel_between_nodes`. This PR is so large that we may encounter a few conflicts when this PR is ready to merge (this happened to the function `create_n_nodes_and_channels_with_index_amounts`).  When this PR is ready to merge, I will do what you said.

src/fiber/channel.rs
contrun: I think so. Because we are actually adding and removing the tlc from exactly the same function `apply_add_tlc_operation_with_peeled_onion_packet`.

src/fiber/channel.rs
contrun: The store interface is a little bit confusing to me. Most of the store traits will never return errors. I don't know under what situation `insert_payment_preimage` may return an error (the current code does not return any error https://github.com/contrun/fiber/blob/fed2e63c7ec248d7319449d9bd7f388f3a943224/src/store/store.rs#L479-L488). It was my understanding that failure at this stage should be internal errors related to the store saving (not data validation error like the hashing result of this preimage does not match the hash). Can you recall why the `InvoiceStore` has `Result` as its return value? I thought all the business logic (data validation etc) should be done by the caller of the store trait (instead of in the store trait).

src/rpc/invoice.rs
contrun: Forgot about this sweet. Thanks.

src/fiber/channel.rs
chenyukang: never mind, 1019 will return error, I got it wrong.
or you can remove the returning result for `insert_payment_preimage` by the way.

contrun: I changed this to `FinalIncorrectPaymentHash` anyway.

doitian:
I'll take over this PR. Unresolved review issues are listed in #615

PR #513, opened 2025-02-18
chore: public RPC structs

jjyr:
Change RPC structures & fields to public

### DIFF
--- a/src/fiber/serde_utils.rs
+++ b/src/fiber/serde_utils.rs
@@ -99,7 +99,7 @@ macro_rules! uint_as_hex {
             pub $name,
             $ty,
             |u: &$ty| format!("0x{:x}", u),
-            |hex: &str| -> Result<$ty, String> {
+            |hex: String| -> Result<$ty, String> {
                 let bytes = hex.as_bytes();
                 if bytes.len() < 3 || &bytes[..2] != b"0x" {
                     return Err(format!("uint hex string does not start with 0x: {}", hex));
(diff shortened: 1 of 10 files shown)


chenyukang:
what's the context to change to `pub`?

src/fiber/serde_utils.rs
```diff
@@ -99,7 +99,7 @@ macro_rules! uint_as_hex {
             pub $name,
             $ty,
             |u: &$ty| format!("0x{:x}", u),
-            |hex: &str| -> Result<$ty, String> {
+            |hex: String| -> Result<$ty, String> {
```
chenyukang, line 102: why change this type?

jjyr: An error occured when deserialize U128Hex with `toml`.

``` txt
[2025-02-18T07:42:59Z INFO  fiber_autopilot] Using config file: fiber-autopilot.toml
Error: TOML parse error at line 5, column 19
  |
5 | available_funds = "0xE8D4A51000"
  |                   ^^^^^^^^^^^^^^
invalid type: string "0xE8D4A51000", expected a borrowed string
```

This line [fiber-autopilot](https://github.com/jjyr/fiber-autopilot/blob/master/src/config.rs#L45) caused the error.


I have not fully understand this issue, but follow the suggestion can fix this error https://github.com/jonasbb/serde_with/issues/620

jjyr:
> what's the context to change to `pub`?

To implement the fiber RPC client.

https://github.com/jjyr/fiber-autopilot/blob/master/src/rpc/client.rs

chenyukang (approved):

PR #517, opened 2025-02-22
feat: ensure funding tx is built incrementally

doitian:
This simplify funding tx watching that a success on-chain funding tx
will invalidate all the other versions for the same channel.


doitian:
Will create a new PR to add more verifications on funding tx

PR #518, opened 2025-02-25
fix: resolve pending tlc settlement issue

quake:


src/fiber/channel.rs
```diff
@@ -1304,9 +1304,18 @@ where
...
         state.check_for_tlc_update(None, true, false)?;
         state.check_remove_tlc_with_reason(TLCId::Received(command.id), &command.reason)?;
-        state
+        let payment_hash = state
             .tlc_state
             .set_received_tlc_removed(command.id, command.reason.clone());
+        if let RemoveTlcReason::RemoveTlcFulfill(RemoveTlcFulfill { payment_preimage }) =
+            command.reason
+        {
+            self.store
```
chenyukang, line 1313: There is the same store logic in `handle_remove_tlc_peer`, maybe it's should be moved to `apply_remove_tlc_operation`, which will invoked after two peers singed, it's both for inbound and outbound tlcs.

src/fiber/channel.rs
```diff
@@ -4983,6 +5009,40 @@ impl ChannelActorState {
...
+            (received_tlcs, offered_tlcs)
+        } else {
+            for tlc in received_tlcs.iter_mut().chain(offered_tlcs.iter_mut()) {
+                // Need to flip these fields for the counterparty.
+                tlc.flip_mut();
+            }
+            (offered_tlcs, received_tlcs)
+        };
+        a.sort_by(|x, y| u64::from(x.tlc_id).cmp(&u64::from(y.tlc_id)));
+        b.sort_by(|x, y| u64::from(x.tlc_id).cmp(&u64::from(y.tlc_id)));
```
chenyukang, line 5027: This part of get tlcs logic is same with `get_active_tlcs`, maybe we should split current `get_active_tlcs` into two function, and `get_active_tlcs` is just filtering out tlcs we are interested in.

src/store/store.rs
```diff
@@ -436,6 +436,13 @@ impl InvoiceStore for Store {
         batch.commit();
         Ok(())
     }
+
+    fn search_payment_preimage(&self, payment_hash_prefix: &[u8]) -> Option<Hash256> {
+        let prefix = [&[CKB_INVOICE_PREIMAGE_PREFIX], payment_hash_prefix].concat();
+        let mut iter = self.prefix_iterator(prefix.as_slice());
+        iter.next()
+            .map(|(_key, value)| deserialize_from(value.as_ref(), "Hash256"))
+    }
```
chenyukang, line 445: will it be an issue if the first 20 bytes of the payment hash mapped into multiple preimages?

tests/bruno/e2e/watchtower/force-close-with-pending-tlcs/21-check-balance-node1.bru
```diff
@@ -0,0 +1,40 @@
...
+
+script:pre-request {
+  let script = bru.getVar("NODE1_FUNDING_SCRIPT");
+  let body = req.getBody();
+  body.params[0].script = script;
+  req.setBody(body);
+}
+
+assert {
+  Number(BigInt(bru.getVar("NODE1_BALANCE")) - BigInt(res.body.result.capacity)): lt 6005000
```
chenyukang, line 39: why we can not use `eq` for assertion here?

chenyukang (approved):

chenyukang:
maybe we need to add more unit test for `src/watchtower/actor.rs` in future, current e2e testing does not cover all the corner cases.

tests/bruno/e2e/watchtower/force-close-with-pending-tlcs/21-check-balance-node1.bru
quake: Here the fee will be different depending on the settlement tx size, for example, after running several e2e tests in CI server, it may take 1 ~ 2 input cells to fill up the fee, so the whole transaction size will be different, usually the fee is 2000~3000 shannons, so here is a little bit of buffer, to assert the fee is less than 5000.

src/store/store.rs
quake: It's true that there is a negligible chance of a hash conflict, but it's not a problem, and unlocking works fine with this preimage.

src/fiber/channel.rs
quake: let me create a new PR for refactoring, thanks.

src/fiber/channel.rs
quake: let me create a new PR for refactoring, thanks.

PR #522, opened 2025-02-26
Persist cch state and make cch a standalone service

contrun:
This PR is based on https://github.com/nervosnetwork/fiber/pull/511.  We can see the code changes better with https://github.com/nervosnetwork/fiber/compare/c53f6dfebba73ae7f9b3b75622abd7e8132eb3b6...contrun:fiber:cch-persistence (but this link is not good to leave comment).


doitian:
Continue in #615

PR #487, opened 2025-01-17
Handle private channel hop hints

contrun:
This PR depends on [Use the fee rate in oubound channel to calculate the TLC fee by contrun · Pull Request #489 · nervosnetwork/fiber](https://github.com/nervosnetwork/fiber/pull/489).

The migration code is not well-tested. I only tried the following.
1. Run `git checkout v0.3.1` to checkout to the v0.3.1 version code.
2. Run `REMOVE_OLD_STATE=y ./tests/nodes/start.sh` to start a few fnn nodes with all new states.
3. Run `(cd ./tests/bruno; npm exec -- @usebruno/cli@1.20.0 run e2e/router-pay -r --env test)` to create a few payment sessions.
4. Run `ldb --db=tests/nodes/1/fiber/store --secondary_path=/tmp/test/ dump --hex | awk '/0xC0/ {print $1}' | sed 's/0xC0/0x/g'` to see all the payment sessions created.
5. Run `ldb --db=tests/nodes/1/fiber/store --secondary_path=/tmp/test/ dump --hex | awk '/0xC0/ {print $1}' | sed 's/0xC0/0x/g' | xargs -I __ curl -sS --json '{"id": "42", "jsonrpc": "2.0", "method": "get_payment", "params": [{"payment_hash": "__"}]}' http://127.0.0.1:21714/ | jq -s` to see all the payment session detailed information.
6. Stop the nodes created in step 2.
7. Run `git checkout handle-private-channel-hop-hints` to checkout the code of this PR.
8. Run `(cd migrate; cargo run -- --path ../tests/nodes/1/fiber/store)` to start migration.
9. Run `cargo build && (clear; cd ./tests/nodes/; ../../target/debug/fnn -d 1)` to start node 1.
10. Run `ldb --db=tests/nodes/1/fiber/store --secondary_path=/tmp/test/ dump --hex | awk '/0xC0/ {print $1}' | sed 's/0xC0/0x/g' | xargs -I __ curl -sS --json '{"id": "42", "jsonrpc": "2.0", "method": "get_payment", "params": [{"payment_hash": "__"}]}' ` again to check the payment session data have been migrated (by eyes).

PR #528, opened 2025-02-27
Confirmed failed tlc should not in check limit and build settlement outputs

chenyukang:


src/fiber/channel.rs
```diff
@@ -4653,31 +4653,28 @@ impl ChannelActorState {
...
+        {
+            return Err(ProcessingChannelError::RepeatedProcessing(format!(
+                "Trying to insert tlc with duplicate payment hash {:?}",
+                payment_hash
+            )));
         }
+
         if tlc.is_offered() {
-            let sent_tlc_value = self.get_offered_tlc_balance(false);
+            let sent_tlc_value = self.get_offered_tlc_balance(true);
```
quake: If you change it in this way, the get_offered_tlc_balance and  get_received_tlc_balance fn doesn't need a bool argument anymore, right?

chenyukang: yes, we can remove the flag, filter out failed_removed_confirmed tlcs by default.

contrun (approved):
LGTM. There are two tests in https://github.com/nervosnetwork/fiber/pull/487 that require the code in this PR.

PR #529, opened 2025-02-28
Fix tempdir issue in testing utils

chenyukang:
The previous `generate_store` have a bug that the temporary directory will be deleted when `generate_store` returns, it depends on OS, if the delete operation is finished quickly, the test will fail when writing data to db.


src/fiber/tests/history.rs
```diff
@@ -18,9 +20,15 @@ impl Round for f64 {
...
+fn generate_history() -> (PaymentHistory<Store>, TempDir) {
+    let (store, temp_dir) = generate_store();
+    let history = PaymentHistory::new(gen_rand_fiber_public_key(), None, store);
+    (history, temp_dir)
+}
+
 #[test]
 fn test_history() {
-    let mut history = PaymentHistory::new(gen_rand_fiber_public_key(), None, generate_store());
+    let (mut history, _temp_dir) = generate_history();
```
contrun: It is still super easy for someone unfamiliar with this code to write code like `let mut history = generate_history().1;`. I don't know if that still works. May be we should wrap `(PaymentHistory<Store>, TempDir)` all together to a struct.

chenyukang: Is it better to add a new option field for Store, such as `Option<Arc<TempDir>>`, so that Store will hold the TempDir if it's generated from a temp directory.

Your branch also changed this part, maybe we can do this after merging your branch.

chenyukang: Fixed with https://github.com/nervosnetwork/fiber/pull/529/commits/8639eb6aea0ea5b0e53c50649b6984e8beebf7be

PR #538, opened 2025-03-03
fix: resolve watchtower submit settlement tx error

quake:
Add delay_epoch checking to ensure commitment transactions are ready to settle before proceeding and modify the params of `collect_live_cells` to use false to avoid lock cells.


chenyukang (approved):

chenyukang:
There are a lot of `debug!` left.

quake:
> There are a lot of `debug!` left.

There's still a small chance that the CI will fail, and I'd like to keep this log for future investigation.

PR #549, opened 2025-03-07
BREAKING CHANGE: Empty cell_deps before compute signing message

jjyr:
This PR change the signing message of funding lock and commitment lock (while have pending htlcs).

The details https://github.com/nervosnetwork/fiber-scripts/pull/17

Binary scripts are compiled on my local machine, and will be replaced after https://github.com/nervosnetwork/fiber-scripts/pull/17 get merged.


chenyukang:
@quake seems there will be duplicated helper functions in `fiber` and `fiber-script`, such as `compute_tx_message` and some other functions.

can we reorganize `fiber-script` to split a crate which `fiber` could depend on?

jjyr:
`fiber` and `fiber-scripts` share several functions, but the actual requirement here is a stable fiber-signing interface as a standalone crate that fiber-scripts tests and third-party tools can depend on. Other than a crate includes serveral functions that we can only use internally, it only brings more maintainance work.

IMO, we should postpone this until we are ready to provide serious crates to external developers, such as an fiber RPC client, fiber type crates (like ckb-types), and transaction composition and signing crates.

quake:
LGTM, but please hold for merging, plan to release with 0.5.0

quake (approved):

jjyr:
Update contracts binaries https://github.com/nervosnetwork/fiber-scripts/pull/18

PR #563, opened 2025-03-08
chore: fix not stable tests 

chenyukang:
send_payment will return error when amount is 0, which makes CI not stable.


chenyukang:
- https://github.com/nervosnetwork/fiber/actions/runs/13735858375/job/38419699087?pr=556
- https://github.com/nervosnetwork/fiber/actions/runs/13745550718/job/38439995272?pr=564

PR #569, opened 2025-03-11
Find path will try other channels considering busyness level 

chenyukang:
Find path will track how many payments are pending on a channel, then try to pick other channel which is not busy.


chenyukang:
I refactor testing code to add benchmark testing, from the result, we can get:

1. the performance for multiple channel scenarios improved
2. the performance does not change when channel number is 1
3. performance will be improved much more when the number of channels increasing

image

doitian:
Could you please explain the algorithm to find the route based on busyness level?

chenyukang:
https://github.com/nervosnetwork/fiber/pull/569/files#diff-ac11142a8dee80631cd1352a43e26ba373f19dbd6009bcc71c66812b508ca47cR795-R811

Supposing there are multiple channels between two nodes, use these two functions to track the inflight payment number of a channel(an inflight payment means payment is sent out but haven't received a result).

if we have the inflight number for a channel, we can add a penalty when calculating the weight for a channel in path finding:
https://github.com/nervosnetwork/fiber/pull/569/files#diff-ac11142a8dee80631cd1352a43e26ba373f19dbd6009bcc71c66812b508ca47cR1354-R1370

quake (approved):

PR #570, opened 2025-03-12
fix: resolve pending tlc expiry time bug

quake:
The bug is that the code incorrectly compares `tlc.expiry` with current_time, which is divided by 1000 in witness, we should use the settlement data's expiry time.


chenyukang (approved):

gpBlockchain:
LGTM

PR #575, opened 2025-03-13
Fix errors in the Testnet Public Nodes User Manual

sunchengzhu:
1. open_channel requires only 20 RUSD.
2. The open_channel_auto_accept_min_ckb_funding_amount is 438 CKB, not 500 CKB.
3. Change nodeA (500 CKB) ⟺ node1 (0) to nodeA (500 CKB) ⟺ node1 (250 CKB) because node1 also contributes funding.
4. Add “Why is the local_balance 0xa32aef600 (43,800,000,000) and the remote_balance 0x460913c00 (18,800,000,000)?”


chenyukang:
To fix the typos ci fail,

set this to `typos.toml`:

```toml
[files]
extend-exclude = [
  "*.patch",   # Automatically generated files that should not be manually modified.
  "*.bin",     # Binary files
  ".git/",     # Version control files
]
ignore-hidden = false

[default.extend-words]
pn = "pn"
ba = "ba"

```

chenyukang:
Could you please format `docs/testnet-nodes.md`?
It will makes inline json looks much better.

sunchengzhu:
> Could you please format `docs/testnet-nodes.md`? It will makes inline json looks much better.

“pn” is part of the data in the `invoice_address` and `invoice`. This is a false positive; the typos tool incorrectly identified it as a spelling error. 

To resolve this CI issue, I reverted the `invoice_address` and `invoice` data back to a previous version that did not cause any errors.

PR #577, opened 2025-03-13
Add BuildRouter and SendPaymentWithRouter

chenyukang:
- Add rpc `build_router`
- Add rpc `send_payment_with_router`, allows users to specify a full route manually, used for   the scenario like rebalancing.
- Fix a amount to send checking issue in `find_path`


chenyukang:
migration is need for https://github.com/nervosnetwork/fiber/pull/577/files#diff-176f6d4f89e68f8dda8ca44f71496fbfc28d7fbf135f1b5f2795c8a183a0bc18R435

added new field for `SendPaymentData`.

src/rpc/README.md
```diff
@@ -675,6 +679,77 @@ Retrieves a payment.
...
+ This is a strong restriction given on payment router, which means these specified hops and channels
+ must be adapted in the router. This is different from hop hints, which maybe ignored by find path.
+ If channel is not specified, find path algorithm will pick a channel within these two peers.
+
+ An error will be returned if there is no router could be build from given hops and channels
+* `final_tlc_expiry_delta` - <em>`Option<u64>`</em>, the TLC expiry delta should be used to set the timelock for the final hop, in milliseconds
+
+##### Returns
+
+* `hops_info` - <em>Vec<[PathEdge](#type-pathedge)></em>, The hops information for router
```
Xcodes-chain: The RPC method build_router has a naming issue in its request and response parameters. Currently, both are named hops_info , which could be confusing for developers even though they represent different structures.

Suggested improvements:

1. For request parameter: rename to hop_hints to match the payment API
2. For response parameter: rename to route_hops or routing_path to better describe the routing result
This would make the API more intuitive and consistent with other parts of the codebase.

chenyukang: `hops_info` is different with `hop_hints`, we should keep different names.
I will change the name of field in return response.

src/fiber/graph.rs
```diff
@@ -1311,6 +1327,119 @@ where
...
+    // we still try to select the best channel based on the channel's capacity and fee rate,
+    // there difference is that we don't need to use a aggregated distance which is used
+    // in the path finding algorithm, because we assume all the nodes in the path are
+    // already selected, and we only need to find the best channel for each hop.
+    pub(crate) fn build_path(
+        &self,
+        source: Pubkey,
+        command: BuildRouterCommand,
+    ) -> Result<Vec<RouterHop>, PathFindError> {
+        let mut router_hops = command.hops_info.clone();
```
quake: suggest to deconstruct `BuildRouterCommand` which will remove this unnecessary clone.

src/fiber/graph.rs
```diff
@@ -1311,6 +1327,119 @@ where
...
+        let mut path = vec![];
+        // If not set, the minimum routable amount `1` is used
+        let mut agg_amount = command.amount.unwrap_or(1);
+        let mut agg_tlc_expiry = command
+            .final_tlc_expiry_delta
+            .unwrap_or(DEFAULT_TLC_EXPIRY_DELTA);
+        for current in 0..router_hops.len() {
+            let cur_hop = &router_hops[current];
+            let prev_hop = &router_hops.get(current + 1);
+            let prev_hop_pubkey = prev_hop.map(|x| x.pubkey).unwrap_or(source);
```
quake:
```suggestion
        for (idx, cur_hop) in router_hops.iter().enumerate() {
            let prev_hop_pubkey = router_hops.get(idx + 1).map(|h| h.pubkey).unwrap_or(source);
```

src/fiber/graph.rs
```diff
@@ -1311,6 +1327,119 @@ where
...
+                    .channel_outpoint
+                    .clone()
+                    .unwrap_or(channel_outpoint.clone())
+                    != channel_outpoint
+                {
+                    continue;
+                }
+
+                let mut current_amount = agg_amount;
+                if current_amount > channel_info.capacity() {
```
quake: I think this check should be moved after forward fee calculation, could you add an unit test for this?

chenyukang: Fixed with https://github.com/nervosnetwork/fiber/pull/577/commits/23f5925558d69b05a04492eb4e398be09e72d40c

quake (approved):

PR #580, opened 2025-03-14
Fix migrate CI fail caused by CKB deps

chenyukang:
Backport https://github.com/nervosnetwork/fiber/commit/9da423e584d8bf2e213dd353365a737c810ffd88#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542 to old fiber crate, and make new tags for them.


.github/workflows/ci.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       - run: |
           cargo check
           make check-dirty-rpc-doc
+          cd migrate && cargo check
```
doitian, line 16: We can use matrix to check all projects: the root, migrate and udt-init

doitian:
Could you also paste diffs between the patches and the original tags?

chenyukang:
> Could you also paste diffs between the patches and the original tags?

https://github.com/nervosnetwork/fiber/commit/0671b08065a8d6c7264a66ebb07bc9003cf964a8

PR #594, opened 2025-03-17
BREAKING CHANGE: fix graph channel rpc for channel update info

chenyukang:
This PR return `ChannelUpdateInfo` directly in the graph_channel rpc, this is a broken change since the format changed.

Or we just add new fields in the conversion so we can keep old fields the same as before.


src/fiber/graph.rs
```diff
@@ -127,14 +128,11 @@ impl ChannelInfo {
         &self.udt_type_script
     }
 
-    // Whether this channel is explicitly disabled in either direction.
-    // TODO: we currently deem a channel as disabled if one direction is disabled.
-    // Is it possible that one direction is disabled while the other is not?
     pub fn is_explicitly_disabled(&self) -> bool {
```
quake: suggest  to change fn to private

src/fiber/graph.rs
```diff
@@ -127,14 +128,11 @@ impl ChannelInfo {
...
         match (&self.update_of_node2, &self.update_of_node1) {
-            (Some(update1), _) if !update1.enabled => true,
-            (_, Some(update2)) if !update2.enabled => true,
-            _ => false,
+            (Some(update1), _) if update1.enabled => false,
+            (_, Some(update2)) if update2.enabled => false,
+            _ => true,
```
quake: this fn will return true when `update_of_node1` and `update_of_node2` are None, this doesn't seem to fit the intent of the fn name, could you check it again?

quake:
Please rebase with develop branch.

chenyukang:
This PR don't need migration, the InvoiceData fingerprint changed because of new added comments, and upgrade `check-migration` tool to a new version to ignore comments changed.

quake (approved):

PR #595, opened 2025-03-18
allow chained funding tx

doitian:
This is useful for nodes that have a lot of funds in a single cell and want to open multiple channels with the cell.

### DIFF
--- a/src/ckb/funding/funding_tx.rs
+++ b/src/ckb/funding/funding_tx.rs
@@ -41,7 +41,7 @@ pub struct FundingTx {
 
 #[derive(Debug, Clone)]
 struct LiveCellsExclusion {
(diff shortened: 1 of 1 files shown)


doitian:
Chained funding tx makes the channels creation unstable, and when a node is creating channels with multiple peers, the change outputs can be unknown to some peers and they must wait for receiving the transactions in their CKB nodes. Thus, the change requires a step further to make the design.

doitian:
> Chained funding tx makes the channels creation unstable, and when a node is creating channels with multiple peers, the change outputs can be unknown to some peers and they must wait for receiving the transactions in their CKB nodes. Thus, the change requires a step further to make the design.

Close because this feature requires more efforts to design.

PR #607, opened 2025-03-21
Add list peers rpc

### LINKED ISSUE #574: New RPC Method: listpeers
## Description
I propose adding a new RPC method to the Fiber project that lists all active, currently connected peers, similar to [the listpeers method available in the LND API](https://lightning.engineering/api-docs/api/lnd/lightning/list-peers/).

This method will provide essential information about the peers connected to the Fiber node, which is crucial for network monitoring and debugging.

e.g.
Since [open_channel](https://github.com/nervosnetwork/fiber/blob/main/src/rpc/README.md#channel-open_channel) RPC requires `the peer must be connected first`, it would be nice to make sure the target peer is connected before opening a channel.

### Related Code
https://github.com/nervosnetwork/fiber/blob/79d3e35b8e3788c09c722325ce4d78046c295b45/src/rpc/peer.rs#L50-L73

chenyukang:
Fixes #574

Pending on #605


quake:
#605 was merged, please rebase this PR.

quake (approved):

PR #615, opened 2025-03-25
feat: support fiber multiple hops payment in cross chain transaction

doitian:
Continuation of #511

- [x] Rebase PR and fix conflicts
    - [x] #511
    - [x] #522: will be in a new PR
- [x] Resolve review issues in #511 
    - [x] https://github.com/nervosnetwork/fiber/pull/511#discussion_r1958908708
        - #736 
    - [x] https://github.com/nervosnetwork/fiber/pull/511#discussion_r1960869375
- [x] Pass rust compiler
- [x] Pass clippy
- [x] Pass cargo test: Summary [ 359.136s] 383 tests run: 377 passed, 6 failed, 2 skipped
    - [x] FAIL [   0.436s] fnn fiber::tests::channel::test_create_channel_with_too_large_amounts
    - [x] FAIL [   1.637s] fnn fiber::tests::payment::test_send_payment_hophint_for_mixed_channels_with_udt
    - [x] FAIL [   0.871s] fnn fiber::tests::payment::test_send_payment_self_with_mixed_channel
    - [x] FAIL [   1.684s] fnn fiber::tests::payment::test_send_payment_with_mixed_channel_hops
    - [x] Revert the change to stop cleaning preimage for payer. (See https://github.com/nervosnetwork/fiber/pull/615#discussion_r2146379461)
        - [x] FAIL [   5.276s] fnn fiber::tests::channel::test_network_send_payment_normal_keysend_workflow
        - [x] FAIL [   3.875s] fnn fiber::tests::channel::test_send_payment_with_3_node
    - [x] Refactor #736 to settle ReceiveBTC order and fix cch test cases (See https://github.com/nervosnetwork/fiber/pull/615#discussion_r2146379461)
- [x] Pass e2e test for cross-chain-hub --env test
- [x] Refactor `new_invoice` RPC to auto generate preimage when neither preimage nor hash is set. https://github.com/nervosnetwork/fiber/pull/615/commits/57f6e0982f5460f5fc62dd037d7fa3bcb6196662
- [x] Refactor to make cross-chain-hub non-wasm only

Co-Authored-By: @contrun


doitian:
I did a quick rebase and there are a log of conflicts. I just ensure `cargo check` pass and have not read block by block to apply the changes in #511 correctly.

Next action: scan block by block to ensure the original changes in #511 are correctly applied to the current develop.

doitian:
Due to the efforts required in https://github.com/nervosnetwork/fiber/pull/615#issuecomment-2750012011 and the benifits is limited, this PR will be cancelled for now.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1196,8 +1228,14 @@ where
 
         let (tlc_info, remove_reason) = state.remove_tlc_with_reason(tlc_id)?;
 
-        if matches!(remove_reason, RemoveTlcReason::RemoveTlcFulfill(_)) {
+        // We should only update invoice status to paid if the peer is paying us (i.e., the TLC is received).
+        // For the TLC sent by ourselves (we're paying the peer), we should leave any invoice as it is.
+        // The reason for this is that it is possible that we're paying another peer by sending a TLC and
+        // we have a local invoice with the same payment hash. See the test
+        // test_store_update_subscription_mock_cross_chain_payment for a concrete example.
+        if tlc_id.is_received() && matches!(remove_reason, RemoveTlcReason::RemoveTlcFulfill(_)) {
```
chenyukang: This check makes the `test_network_send_payment_normal_keysend_workflow` failed.

doitian: @contrun is it safe to revert the change here?

contrun: I don't think we should remove this check, as explained in the comments, this condition is used to avoid situations like the test test_store_update_subscription_mock_cross_chain_payment. It is better to check the concrete reason why test_network_send_payment_normal_keysend_workflow failed and update the conditions if necessary.

doitian: Keeping this, these 2 tests will fail:

- fiber::tests::channel::test_network_send_payment_normal_keysend_workflow
- fiber::tests::channel::test_send_payment_with_3_nodes

Removing this, these 4 tests will fail

- cch::tests::payment::test_cross_chain_payment_hub_receive_btc_always_success_multiple_hops
- cch::tests::payment::test_cross_chain_payment_hub_receive_btc_always_success_single_hop
- cch::tests::payment::test_cross_chain_payment_hub_receive_btc_simple_udt_multiple_hops
- cch::tests::payment::test_cross_chain_payment_hub_receive_btc_simple_udt_single_hop

doitian: For keysend, the payer should ensure generating a random preimage.

For non-keysend, the payee is responsible to ensure a unique preimage for each payment. If payee uses a known preimage, the intermidiate hops can claim the payment without sending the money to the final payee. Hower, the payer has the signed invoice and the preimage thus can proove the payment. It seems that what the comment describes is not a problem.

The reason why `fiber::tests::channel::test_network_send_payment_normal_keysend_workflow` and `fiber::tests::channel::test_send_payment_with_3_nodes` fails is that they assert that the payer will remove the preimage from local store once the payment succeeds.

doitian: I'll keep the original logic and see how to fix the 4 new cch test cases. This change is unrelated to the cch feature. If it is really a concern, we should fix it in a new PR.

doitian: The cause that the 4 new cch test cases fail is that preimage has been cleaned when fetching the payment updated event in the StorePublisher actor asynchrouniously. I have refactored #736 to emit the store update event synchrouniously.

doitian: Future works should be done in #741

crates/fiber-lib/src/invoice/invoice_impl.rs
```diff
@@ -627,11 +626,18 @@ impl InvoiceBuilder {
...
+    /// The hash of the preimage. If hash is set, preimage should be absent.
+    /// This condition indicates a 'hold invoice' for which the
+    /// tlc must be accepted and held until the preimage becomes known.
     pub fn payment_hash(mut self, payment_hash: Hash256) -> Self {
         self.payment_hash = Some(payment_hash);
         self
     }
 
+    /// The preimage to settle an incoming TLC payable to
+    /// this invoice. If preimage is set, hash should not. If both
```
chenyukang: `payment_hash` should **not be set**.

crates/fiber-lib/src/invoice/tests/invoice_impl.rs
```diff
@@ -401,6 +401,19 @@ fn test_invoice_builder_both_payment_hash_preimage() {
     );
 }
 
+#[test]
+fn test_invoice_builder_neither_payment_hash_not_preimage() {
```
chenyukang, line 420: Please also add a unit test for `BothPaymenthashAndPreimage` error.

crates/fiber-lib/src/rpc/invoice.rs
```diff
@@ -31,9 +31,9 @@ pub struct NewInvoiceParams {
     pub description: Option<String>,
     /// The currency of the invoice.
     pub currency: Currency,
-    /// The payment preimage of the invoice, may be empty for a hold invoice.
+    /// The preimage to settle an incoming TLC payable to this invoice. If preimage is set, hash should not. If both preimage and hash are absent, a random preimage is generated.
```
chenyukang: `hash should not be set`
or
`hash should be absent`

crates/fiber-bin/src/main.rs
```diff
@@ -303,19 +310,8 @@ pub async fn main() -> Result<(), ExitMessage> {
...
-                    );
-                    subscribers.settled_tlcs_subscribers.subscribe(
-                        actor.clone(),
-                        |tlc_notification| {
-                            Some(CchMessage::SettledTlcNotification(tlc_notification))
-                        },
-                    );
-
+                    // Subscribe the actor to the store so it can receive updates
+                    store.subscribe(Box::new(actor.clone()));
```
chenyukang, line 334: Add a info log here so that we know cch started succesfully.
`info!("cch started successfully ...");`

When I run e2e test, I find `Starting cch` in the log and don't know when to run bruno test.

crates/fiber-lib/src/cch/order.rs
```diff
@@ -71,12 +72,10 @@ pub struct SendBTCOrder {
     pub wrapped_btc_type_script: ckb_jsonrpc_types::Script,
 
     pub btc_pay_req: String,
-    pub ckb_pay_req: String,
+    pub fiber_payee_pubkey: Pubkey,
+    pub fiber_pay_req: String,
```
chenyukang: from the code, I found it's a fiber invoice, maybe `fiber_pay_invoice` is a better name?

chenyukang: it's better to use `Option<CkbInvoice>` here so we don't need to convert between string format.
or seems it's better if we generate `CkbInvoice` in SendBTCOrder constructor(Perhaps there will be some redundant code for processing parameters), so we can use `CkbInvoice` here.

crates/fiber-lib/src/cch/actor.rs
```diff
@@ -440,39 +439,25 @@ impl CchActor {
...
+        let hash = Hash256::from_str(&event.payment_hash)?;
+
         order.status = event.status;
-        if let (Some(preimage), Some(network_actor), Some(channel_id), Some(tlc_id)) = (
-            event.preimage,
-            &self.network_actor,
-            order.channel_id,
-            order.tlc_id,
-        ) {
+        if let Some(preimage_str) = event.preimage {
```
chenyukang, line 445: do we need to verify whether it's match for `payment_hash` and `preimage` here?

crates/fiber-lib/src/cch/actor.rs
```diff
@@ -578,43 +552,33 @@ impl CchActor {
...
-                                onion_packet: None,
-                                shared_secret: NO_SHARED_SECRET,
-                                previous_tlc: None,
-                            },
-                            rpc_reply,
-                        ),
+                NetworkActorMessage::Command(NetworkActorCommand::SendPayment(
+                    SendPaymentCommand {
+                        invoice: Some(order.fiber_pay_req.clone()),
+                        ..Default::default()
```
chenyukang, line 567: why we don't need to specify other parameters here and using the default values?

crates/fiber-lib/src/fiber/tests/payment.rs
```diff
@@ -29,6 +31,12 @@ use std::time::SystemTime;
 use tracing::debug;
 use tracing::error;
 
+fn get_simple_udt_script() -> Script {
```
chenyukang: there are other 2 places in testing code using `32e555f3ff8e135cece1351a6a2971518392c1e30375c1e006ad0ce8eac07947` coded args, maybe we should add this in common test utils and call it from all the locations.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -977,35 +981,63 @@ where
...
+                                // The sender sent a TLC with no invoice associated with it.
+                                .ok_or(ProcessingChannelError::FinalIncorrectPaymentHash)?;
+                            let is_active = status == CkbInvoiceStatus::Open
+                                || status == CkbInvoiceStatus::Received;
+                            if is_active {
+                                // This TLC is added to applied_add_tlcs in above, but
+                                // TLCs in the list applied_add_tlcs wouldn't be processed again.
+                                // For the unsettled active hold invoice TLCs, we should process them indefinitely
+                                // until they expire or are settled.
+                                state.tlc_state.applied_add_tlcs.remove(&add_tlc.tlc_id);
```
chenyukang: I think we should not modify `applied_add_tlcs` here, seems it's a fix for test `test_send_payment_succeed_settle_hold_invoice_multiple_times`, please refer to https://github.com/nervosnetwork/fiber/commit/1c1633ccbecb7edcbf52f349ceb261e3a0a65283

crates/fiber-lib/src/invoice/tests/invoice_impl.rs
doitian: There's already a test case `test_invoice_builder_both_payment_hash_preimage`

crates/fiber-lib/src/cch/actor.rs
doitian: Network actor will return the error

crates/fiber-lib/src/cch/actor.rs
doitian: Do we need to repeat the fields in invoice?

crates/fiber-lib/src/fiber/channel.rs
doitian: I don't know. I would say let's remove it and see what will happen.

doitian: Cherry picked [cc69a53](https://github.com/nervosnetwork/fiber/pull/615/commits/cc69a532e6768babf079607dadc187eab58ccb11)

doitian:
Actions required: refactor the code to make cch a non-wasm feature. Current code base has to put `cfg` for non-wasm environments in many places.

crates/fiber-bin/src/main.rs
```diff
@@ -307,7 +311,10 @@ pub async fn main() -> Result<(), ExitMessage> {
                 new_tokio_task_tracker(),
                 new_tokio_cancellation_token(),
                 root_actor.get_cell(),
-                network_actor.clone(),
```
jjyr, line 310: Return an ExitMessage instead of panic.

For example
``` rust
            if fiber_config.standalone_watchtower_rpc_url.is_none()
                && fiber_config.disable_built_in_watchtower.unwrap_or_default()
            {
                return ExitMessage::err(
                    "fiber config requires standalone watchtower rpc url or built-in watchtower to be enabled"
                        .to_string(),
                );
            }
```

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1155,17 +1162,37 @@ where
...
+                    .ok_or_else(|| {
+                        error!("preimage is not found for payment hash: {:?}", payment_hash);
+                        ProcessingChannelError::FinalIncorrectPaymentHash
+                    })?;
+                if status == CkbInvoiceStatus::Open {
+                    self.store
+                        .update_invoice_status(&payment_hash, CkbInvoiceStatus::Received)
+                        .expect("update invoice status failed");
+                }
+                if let Err(e) = self.store.add_invoice_channel_info(
```
jjyr, line 1179: With MPP support, an invoice may paid by multiple channels. We already have the payment information stored in Attempt, so maybe we can fetch these information from MPP Holdtlc instead of adding another payment-to-channel mapping.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1155,17 +1162,37 @@ where
 
                 self.store_preimage(payment_hash, preimage);
             } else {
-                error!("preimage is not found for payment hash: {:?}", payment_hash);
-                return Err(ProcessingChannelError::FinalIncorrectPaymentHash);
+                // Check HODL invoice. The updating of hold invoice is always done in settle_invoice rpc call.
+                let status = self
+                    .store
+                    .get_invoice_status(&payment_hash)
+                    // The sender sent a TLC with no invoice associated with it.
```
jjyr, line 1169: https://github.com/nervosnetwork/fiber/pull/615/files#diff-0089cc1b51a74d30eb20e939ada0141f62d61c26d2754442900a2c7229aa027bR1137

Can we just insert preimage to store in the settle_invoice rpc call?

If it is possible the code above has a fallback to read preimage from DB, then we can reuse the same logic to settledown. The else branch can be removed.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2276,6 +2298,50 @@ where
...
+        if let Some(CkbInvoiceStatus::Received) = self.store.get_invoice_status(payment_hash) {
+            let channels = self.store.get_invoice_channel_info(payment_hash);
+            let total_amount: u128 = channels.iter().map(|c| c.amount).sum();
+            match invoice.amount() {
+                Some(amount) if total_amount < amount => {
+                    return Ok(());
+                }
+                _ => {
+                    // Only settle the invoice if the client has paid the full amount.
+                    for channel in channels {
```
jjyr, line 2330: https://github.com/nervosnetwork/fiber/blob/develop/crates/fiber-lib/src/fiber/network.rs#L1662

Can we reuse MPP logic

crates/fiber-lib/src/fiber/channel.rs
jjyr: https://github.com/nervosnetwork/fiber/pull/615/files#diff-14e04ef264a3a4c449eeb77f85a2a5db75b714848b0483b9122d03d516a3e769R2318

I have noticed we already do it in rpc

crates/fiber-lib/src/fiber/channel.rs
jjyr: https://github.com/nervosnetwork/fiber/blob/af3aaa8df2b81a6649dd6901a52a6b29ac3b30af/crates/fiber-lib/src/fiber/channel.rs#L2719

MPP hold tlc

jjyr:
We should let MPP to handle the multi path payment and remove the `InvoiceChannelInfo` related logic.
The CCH module only needs to check payment status and trigger payment settlement.

crates/fiber-lib/src/cch/actor.rs
chenyukang: emm, I forget why I asked this question.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -170,6 +171,7 @@ pub enum ChannelCommand {
     BroadcastChannelUpdate(),
     Update(UpdateCommand, RpcReplyPort<Result<(), String>>),
     ForwardTlcResult(ForwardTlcResult),
+    SettleHeldTlc(Hash256),
```
chenyukang, line 174: do we have expiry check for `HeldTlc`?

crates/fiber-lib/src/fiber/channel.rs
doitian: In CCH, the preimage is not available when the TLC is received. So the hold invoice for CCH has no preimage yet in the store here.

doitian:
> We should let MPP to handle the multi path payment and remove the `InvoiceChannelInfo` related logic. The CCH module only needs to check payment status and trigger payment settlement.

It seems that we can replace `InvoiceChannelInfo` with `HoldTlc`.

doitian:
All features in this PR have been implemented after #950 is merged.

PR #619, opened 2025-03-27
chore: Fix outdated hop hint documentation

chenyukang:
The meaning of hop hint is different from the initial version, so update the documentation.

### DIFF
--- a/src/rpc/README.md
+++ b/src/rpc/README.md
@@ -624,10 +624,11 @@ Sends a payment to a peer.
   }
  ```
 * `hop_hints` - <em>Option<Vec<[HopHint](#type-hophint)>></em>, Optional route hints to reach the destination through private channels.
+ Note: this is only used for the private channels with the last hop.
  A hop hint is a hint for a node to use a specific channel, for example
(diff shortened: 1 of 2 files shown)


Xcodes-chain (commented):
The fee_rate and tlc_expiry_delta parameters in the hop_hints are of type u64 . Previously, we mistakenly treated them as u128 and converted them to hex strings using hex() . This was incorrect. These values should be passed as regular integers without hex conversion.

chenyukang:
> The fee_rate and tlc_expiry_delta parameters in the hop_hints are of type u64 . Previously, we mistakenly treated them as u128 and converted them to hex strings using hex() . This was incorrect. These values should be passed as regular integers without hex conversion.

no, every parameter in fiber with integer type should be in hex encoded, if not, it's a bug.

chenyukang:
it should be in `U64Hex`:
https://github.com/nervosnetwork/fiber/blob/fa9028470da8b421c2dab9c5fe4c1178dbf394a6/src/rpc/payment.rs#L204-L209

quake (approved):

PR #623, opened 2025-03-28
fix: should force close channel when htlc is expired without response 

quake:
Consider a HTLC chain Peer A -> Peer B -> Offline Peer C

Peer B should force close the channel between B and C after htlc is expired without response, and A should force close the channel cascade too.

## Break change

* Modified `ShutdownChannelParams` to make `close_script` and `fee_rate` optional and added validation to ensure they are set correctly based on the `force` parameter.

### DIFF
--- a/src/rpc/channel.rs
+++ b/src/rpc/channel.rs
@@ -488,14 +488,12 @@ where
                     Some(params),
                 ));
             }
-        } else {
-            if params.close_script.is_none() || params.fee_rate.is_none() {
-                return Err(ErrorObjectOwned::owned(
-                    CALL_EXECUTION_FAILED_CODE,
-                    "close_script and fee_rate should be set when force is false",
-                    Some(params),
-                ));
-            }
+        } else if params.close_script.is_none() || params.fee_rate.is_none() {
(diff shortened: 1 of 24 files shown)


chenyukang (approved):

src/rpc/channel.rs
```diff
@@ -279,12 +279,13 @@ pub struct ShutdownChannelParams {
...
-    /// Whether to force the channel to close
-    pub force: Option<bool>,
+    pub close_script: Option<Script>,
     /// The fee rate for the closing transaction, the fee will be deducted from the closing initiator's channel balance
-    #[serde_as(as = "U64Hex")]
-    pub fee_rate: u64,
+    #[serde_as(as = "Option<U64Hex>")]
+    pub fee_rate: Option<u64>,
+    /// Whether to force the channel to close, when set to false, `close_script` and `fee_rate` should be set, default is false.
+    /// When set to true, `close_script` and `fee_rate` will be ignored and will use the default value when opening the channel.
```
chenyukang, line 287:
```suggestion
    /// When set to true, `close_script` and `fee_rate` will be ignored and will use the default value specified during the opening of channel.
```

PR #640, opened 2025-04-08
fix: watchtower should claim upstream tlc with onchain preimage

quake:
Consider a HTLC chain Peer A -> Peer B -> Peer C

Peer C  force close the channel between B and C, then claim the tlc amount by preimage on-chain. B should be able to found this preimage and claim the tlc amount between B and A.

### DIFF
--- a/src/fiber/channel.rs
+++ b/src/fiber/channel.rs
@@ -2690,10 +2690,7 @@ impl TlcInfo {
     }
 
     pub fn id(&self) -> u64 {
-        match self.tlc_id {
-            TLCId::Offered(id) => id,
-            TLCId::Received(id) => id,
-        }
+        self.tlc_id.into()
     }
 
     pub fn is_offered(&self) -> bool {
--- a/src/fiber/channel.rs
+++ b/src/fiber/channel.rs
@@ -2689,6 +2689,13 @@ impl TlcInfo {
         )
     }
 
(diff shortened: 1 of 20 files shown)


src/fiber/channel.rs
```diff
@@ -2689,6 +2689,13 @@ impl TlcInfo {
         )
     }
 
+    pub fn id(&self) -> u64 {
+        match self.tlc_id {
+            TLCId::Offered(id) => id,
+            TLCId::Received(id) => id,
+        }
```
chenyukang:
```suggestion
       self.tlc_id.into()
```

src/fiber/network.rs
```diff
@@ -1162,6 +1162,33 @@ where
...
+                                    debug!(
+                                        "Found payment preimage for channel {:?} tlc {:?}",
+                                        channel_id,
+                                        tlc.id()
+                                    );
+                                    let (send, _recv) = oneshot::channel();
+                                    let rpc_reply = RpcReplyPort::from(send);
+                                    state
+                                        .send_command_to_channel(
+                                            channel_id,
```
chenyukang, line 1151: is there a possibility that a channel both have a tlc with preimage and also expired tlcs?
maybe it is safer that we iterate channels twice, the first iteration is check tlc with preimage, the second iteration is check expired.

chenyukang (approved):

PR #655, opened 2025-04-11
fix: use saturating_sub to avoid potential underflow

quake:


chenyukang (approved):

chenyukang:
seems a dep of cargo shear upgrade minimal Rust version, I will update rust version.

PR #657, opened 2025-04-11
Fix invoice cancel bug caused by check channels

### LINKED ISSUE #654: [cancel_invoice] send payment success after cancel invoice
channels
a-b-c-d-a
1. a generate 50 invoice
2. for i in range(invoices)
        a.cancel(invoice) 
        a.send_payment(cancel invoice)

3. check invoice status
some of invoice return success

[node.log](https://github.com/user-attachments/files/19701132/node.log)

chenyukang:
This is a corner case for a node pay self, when there are many TLC opes, check channels cron job may try to RemoveTlc for invoice with preimage, but we didn't check invoice status.

Fixes #654


quake (approved):

gpBlockchain:
LGTM

PR #667, opened 2025-04-16
chore: print ckb key path if it is missing

chenyukang:


docs/testnet-nodes.md
doitian, line 1: Only format changes in this file?

chenyukang: yes, formatted by vscode.

doitian (approved):

Issue #606, opened 2025-03-21
Cannot close channel stuck in AWAITING_CHANNEL_READY state


Flouse:
version: "0.4.1",
  commit_hash: "4b05a7ec",
### List Channel

```json
{
  "channel_id": "0x20f4057f1e86f2e3df02b7a2fe168a45c3e24b149e9f6888fe41ccb5a99b5d6c",
  "is_public": true,
  "channel_outpoint": "0xae95ec3cfd75e687977c88df441f764092d7d6e2ecf1dc62dc4a89c88d37557e00000000",
  "peer_id": "QmQV3TSUHForNR3MDxeirxUr8aPuc3twfLkvVByCD2XD9C",
  "funding_udt_type_script": null,
  "state": {
    "state_name": "AWAITING_CHANNEL_READY",
    "state_flags": ""
  },
  "local_balance": "0x0",
  "offered_tlc_balance": "0x0",
  "remote_balance": "0x15d6ea6a08",
  "received_tlc_balance": "0x0",
  "latest_commitment_transaction_hash": "0xd496e9788714f34a685d9a731c9dde2de66454200e2ab377f8a5c04c1858dab5",
  "created_at": "0x195ae068e45",
  "enabled": true,
  "tlc_expiry_delta": "0x5265c00",
  "tlc_fee_proportional_millionths": "0x3e8"
}
```
Channel appears properly funded but never progressed beyond initial state.

### Channel Outpoint
2025/03/19 18:52:00
https://testnet.explorer.nervos.org/transaction/0xae95ec3cfd75e687977c88df441f764092d7d6e2ecf1dc62dc4a89c88d37557e#0

It's been 2 days. The channel is still stuck.

## RPC call "shutdown_channel" failed due to error
1. When attempting to close a channel that's stuck in the AWAITING_CHANNEL_READY state, I received an error indicating invalid state handling:
`error: RPC call "shutdown_channel" failed due to error: Messaging failed because channel is closed`

**Note: Its state is not CLOSED.**

2. Force close didn't help
`error: RPC call "shutdown_channel" failed due to error: Failed to processing channel: Invalid state: Handling force shutdown command invalid state AwaitingChannelReady(AwaitingChannelReadyFlags(0x0))`

## Need HELP

1. Why does a channel remain stuck in AWAITING_CHANNEL_READY despite confirmed funding transactions?
   #### Peer Coordination Failures?
[...]

PR #669, opened 2025-04-17
feat: close funding channels on failures

doitian:
- Close funding channels on failures
- Send TxAbort to notify the peer
- Close funding channels on receiving TxAbort


doitian:
Refs #606 

Actions:

- [x] Add tests

jjyr (approved):

PR #670, opened 2025-04-17
Fix signature error when channel reestablishing

chenyukang:
There are several issues with channel reestablishment:

1. the retryable tasks in actor may have effect on the processing of reestablishing, we need to avoid run tasks.
2. the relay remove tlc may need more retry if previous channel disconnect, otherwise the RemoveTlc may not be relayed into previous hop after reconnecting finished, since we stored previous ChannelActorState in the same db, we can query whether RemoveTlc relay is done.
3. we may update Tlc status and nonces after RevokeAndAck finished, so it is hard to resend RevokeAndAck msg after reconnect, this PR add a new field `last_revoke_ack_msg` for resending it after channel reestablished.
4. take care of some other corner cases.


src/fiber/tests/payment.rs
```diff
@@ -3081,18 +3081,46 @@ async fn test_send_payment_remove_tlc_with_preimage_will_retry() {
...
             if status == PaymentSessionStatus::Success {
                 payments.remove(payment_hash);
             }
-            tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+            tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
         }
         if payments.is_empty() {
             break;
         }
+        let escaped = SystemTime::now()
```
doitian:
```suggestion
        let elapsed = SystemTime::now()
```

src/fiber/tests/payment.rs
```diff
@@ -3174,3 +3202,72 @@ async fn test_send_payment_invoice_cancel_multiple_ops() {
         }
     }
 }
+
+#[tokio::test]
+async fn test_send_payment_no_preimage_invoice_will_make_payment_failed() {
```
doitian, line 4293: From the code, the invoice does have a preimage. I'm confused what this test case is testing.

src/fiber/channel.rs
```diff
@@ -1613,25 +1635,41 @@ where
...
+                                                    id: (*tlc_id),
+                                                    reason: reason.clone(),
+                                                },
+                                                port,
+                                            ),
+                                        },
+                                    ),
+                                ))
+                                .expect(ASSUME_NETWORK_ACTOR_ALIVE);
+                            // the previous hop will automatically retry if there is Waiting_Ack error
```
jjyr, line 1704: should we return false after this line?

src/fiber/tests/payment.rs
chenyukang: The preimage is only create not stored in the db of node1, so node1 can not find them when settle payment.

src/fiber/channel.rs
chenyukang: no, return `true` to keep it until next retry.

quake (approved):

PR #675, opened 2025-04-21
Fix Musig2Error caused by ShuttingDown state

chenyukang:
Test `test_send_payment_shutdown_under_send_each_other` is not stable, under the scenario like A and B send tlc to each other, and B begin to execute `ShutDown` command, A may run `AddTlc` at the same time(adding tlc will run following CommitmentSigned), if `AddTlc` peer message can not be processed by B, B will verify the following signed message with Musig2Error, since tlcs are not synced. 

This PR add a check to avoid normal shutdown workflow when there are still pending HTLC need commitment signed, according to bolt https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#closing-initiation-shutdown

Add some debug utility for debugging signature errors, we always need these logs for fixing this kind of error.


tests/bruno/e2e/cross-chain-hub/14-node3-send-shutdown-channel.bru
chenyukang, line 1: I remove this request because after step 13, channel already gone into the shutdown flow with auto accept @doitian

quake (approved):

PR #679, opened 2025-04-23
Early stage preparation for wasm

Officeyutong:
This PR includes some changes to provide early stage support for wasm:
- Split `fnn` into `fiber-lib` and `fiber-bin`. `fiber-lib` is a library containing main implementation of fiber, while `fiber-bin` is a CLI entry for fiber which won't be supported in wasm
- Isolate interactions with database implementation into `fiber-lib/src/store/store_impl`, clean up rocks db operations in migrate and replace them with operations on `struct Store`.
- Find out crates and code pieces that won't be supported on wasm and isolate them to native only. Note we are stuck by `ckb-sdk` now since it's a complex crates with a lot parts can't be migrates to wasm.


fiber-lib/src/cch/error.rs
```diff
@@ -1,6 +1,7 @@
 use std::time::SystemTimeError;
 
 use jsonrpsee::types::{error::CALL_EXECUTION_FAILED_CODE, ErrorObjectOwned};
+// use jsonrpsee::types::{error::CALL_EXECUTION_FAILED_CODE, ErrorObjectOwned};
```
chenyukang: clean up.

fiber-lib/tests/bruno/e2e/3-nodes-transfer/01-node1-connect-node2.bru
chenyukang, line 1: `tests` directory is used for e2e testing, seems we can keep the old position?

chenyukang:
The top level directory `migrate` is empty now?

fiber-lib/src/cch/error.rs
Officeyutong: fixed

Officeyutong:
> The top level directory `migrate` is empty now?

`migrate` directory has been moved to `fiber-lib`, should we keep it at the original position?

chenyukang:
> > The top level directory `migrate` is empty now?
> 
> `migrate` directory has been moved to `fiber-lib`, should we keep it at the original position?

I prefer to keep it the original position, it's not a part of fiber library but a standalone binary.

Officeyutong:
> > > The top level directory `migrate` is empty now?
> > 
> > 
> > `migrate` directory has been moved to `fiber-lib`, should we keep it at the original position?
> 
> I prefer to keep it the original position, it's not a part of fiber library but a standalone binary.

It's moved to the original path now

chenyukang (approved):

jjyr (approved):

quake (approved):

PR #680, opened 2025-04-24
Use CKB SDK async API

doitian:
- chore(deps): upgrade ckb-sdk to 3.7.0
- use ckb sdk async api in ckb chain module
- use ckb sdk async api in watchtower module

### DIFF
--- a/src/watchtower/actor.rs
+++ b/src/watchtower/actor.rs
@@ -6,7 +6,7 @@ use ckb_sdk::{
     traits::{CellCollector, CellQueryOptions, DefaultCellCollector, ValueRangeOption},
     transaction::builder::FeeCalculator,
     util::blake160,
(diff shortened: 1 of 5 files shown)


src/watchtower/actor.rs
```diff
@@ -131,7 +131,7 @@ where
                     }
                 }
             }
-            WatchtowerMessage::PeriodicCheck => self.periodic_check(state),
+            WatchtowerMessage::PeriodicCheck => self.periodic_check(state).await,
```
doitian, line 134: Should `periodic_check` run in the background without blocking the actor to handle the next message?

chenyukang: `periodic_check` will block actor to handle the next message as before.

PR #681, opened 2025-04-24
Fix payself bug with mixed channels of UDT 

### LINKED ISSUE #678: [send_payment]: The allow_self_payment parameter affects unrelated XUDT channels
version: 0.5.0
network: testnet

After repeated testing, sending CKB with the allow_self_payment parameter causes the Seal channel balance to change.No anomalies occur when the parameter is not included.Additionally, sending USDI affects the RUSD channel.

Image

### create invoce
```
{
    "id": 42,
    "jsonrpc": "2.0",
    "method": "new_invoice",
    "params": [
        {
            "amount": "0xba43b7400",
            "currency": "Fibt",
            "description": "node",
            "expiry": "0x2581",
            "final_cltv": "0x28",
            "payment_preimage": "0xc11111111111111111111111bb164b1b1141f011aefec11e1111111111111111",
            "hash_algorithm": "sha256"
        }
    ]
}
```

### send_payment
```
echo '{
    "id": 42,
    "jsonrpc": "2.0",
    "method": "send_payment",
    "params": [
        {
          "invoice": "fibt500000000001p49t2j4c5ujgfw0pnm6vkhwhejnk05jkgxmqq09qwnuhtfnk2e5yk8v26t4j7n37ysp4ztrgfhvar0ufe8up58t052s677d0z69nec4tzs7jpegpj0mzerxzk0smfxcfmauaxc56ram3p5c0y93f80wy37svrwzrrlj3fm888u8txur4f5hqmc4c8xwe3m3uuf8xaq037fjhrp260mgunqpawyjev3fjwqr6deg98zy66zzmgra2rytl0mqverpxva8g6chyy4uuyjv32e8hd7jwn2vgs6jwgw4377sks3q8zazjqm0n50jnxxr9qql7yfrd",
          "allow_self_payment": true
        }
    ]
}' \
| curl -H 'content-type: application/json' -d @- \
http://127.0.0.1:8227
```

chenyukang:
Fixes #678

Refactor test utilities to support creating UDT channels in unit tests.

### DIFF
--- a/src/fiber/graph.rs
+++ b/src/fiber/graph.rs
@@ -680,6 +680,25 @@ where
         self.channels.get(outpoint)
     }
 
+    pub fn get_outbound_channel_info_and_update(
+        &self,
+        outpoint: &OutPoint,
+        from: Pubkey,
+    ) -> (Option<&ChannelInfo>, Option<&ChannelUpdateInfo>) {
+        let channel_info = self.get_channel(outpoint);
+        if let Some(channel_info) = channel_info {
+            if channel_info.node1() == from {
+                (Some(channel_info), channel_info.update_of_node1.as_ref())
+            } else if channel_info.node2() == from {
+                (Some(channel_info), channel_info.update_of_node2.as_ref())
+            } else {
+                (None, None)
+            }
+        } else {
+            (None, None)
+        }
+    }
+
     pub fn get_channels_with_params(
         &self,
         limit: usize,
@@ -937,7 +956,7 @@ where
         // The source node of the edge.
         from: Pubkey,
         // The target node of the edge.
-        to: Pubkey,
+        target: Pubkey,
         // The amount that the source node will send to the target node.
         next_hop_received_amount: u128,
         // The amount that the source node will receive from forwarding `amount_to_send` to the target.
@@ -959,7 +978,7 @@ where
         let probability = cur_probability
             * self.history.eval_probability(
                 from,
-                to,
+                target,
                 channel_outpoint,
                 next_hop_received_amount,
                 channel_capacity,
@@ -967,7 +986,7 @@ where
 
         debug!(
             "probability: {} for channel_outpoint: {:?} from: {:?} => to: {:?}",
-            probability, channel_outpoint, from, to
+            probability, channel_outpoint, from, target
         );
         if probability < DEFAULT_MIN_PROBABILITY {
             debug!("probability is too low: {:?}", probability);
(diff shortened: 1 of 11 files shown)


gpBlockchain:
lgtm

src/fiber/graph.rs
```diff
@@ -680,6 +679,25 @@ where
         self.channels.get(outpoint)
     }
 
+    pub fn get_outbound_channel_info_and_update(
```
quake: suggest to remove `pub`

src/fiber/graph.rs
```diff
@@ -1054,58 +1071,59 @@ where
...
-                } = hint;
+                // hop hint only referring to private channels for sender node,
+                // if we get public channel information for this hophint, we just ignore this hophint
+                if let (Some(channel_info), Some(channel_update)) =
+                    self.get_outbound_channel_info_and_update(&hint.channel_outpoint, hint.pubkey)
+                {
+                    if channel_info.udt_type_script != udt_type_script || channel_update.enabled {
+                        continue;
+                    }
+                }
```
quake: suggest to use `hop_hints.retain(...)` to filter out public channel first then loop the hints.

src/fiber/graph.rs
```diff
@@ -1250,54 +1249,60 @@ where
...
+                )
             })
-            .collect();
-        for (from, _, channel_info, channel_update) in self.get_node_inbounds(node) {
-            if let Some(balance) = channel_update.outbound_liquidity {
-                if balance < amount {
-                    continue;
+            .filter(|(_, _, expiry, _, channel_info, channel_update)| {
+                // we also check the validity of hint, if the hint is not valid, we will skip it
+                // for instance, if user specify a hint channel with UDT, but the payment is not UDT,
```
quake: Is this code comment outdated, because we have removed the hop_hints param from `adjust_target_for_route_self` now, and does not need to mention hint in the comment?

src/fiber/graph.rs
chenyukang: `hop_hints` is not mutable, if we want to use `retain` we need to change the function signature or use a `clone` for `hop_hints`.

I think we should not require mutable here, I changed the parameter from `hop_hints: Vec<HopHint>`  to `hop_hints: &[HopHint]`, so we don't need to clone it when call `find_path`.

quake: I see, then we may use filter first:
```rust
            let filtered_hints = hop_hints.iter().filter(|hint|
                match self.get_outbound_channel_info_and_update(&hint.channel_outpoint, hint.pubkey) {
                    (Some(channel_info), Some(channel_update)) => {
                        channel_info.udt_type_script == udt_type_script && !channel_update.enabled
                    },
                    _ => true,
                }
            );
            for hint in filtered_hints {
             }
```

quake (approved):

PR #691, opened 2025-05-06
Payment session will retry to send if channel is reestablished

chenyukang:
When peer is disconnected, some payments may be left in `Created` status (which means we haven't sent out TLC for the first hop), after channel is reestablished, we need retry these payments if needed.

### DIFF
--- a/src/fiber/channel.rs
+++ b/src/fiber/channel.rs
@@ -99,6 +99,7 @@ pub const COMMITMENT_CELL_WITNESS_LEN: usize = 16 + 1 + 32 + 64;
 pub const INITIAL_COMMITMENT_NUMBER: u64 = 0;
 
 const RETRYABLE_TLC_OPS_INTERVAL: Duration = Duration::from_millis(1000);
+const WAITING_REESTABLISH_FINISH_TIMEOUT: Duration = Duration::from_millis(4000);
 
 // if a important TLC operation is not acked in 30 seconds, we will try to disconnect the peer.
 #[cfg(not(test))]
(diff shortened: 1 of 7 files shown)


jjyr (approved):

chenyukang:
I think commit https://github.com/nervosnetwork/fiber/pull/691/commits/a6dfa9eb49d72d5085e0b6d41378dfc0ba83304e is not a completed fix, the ci fail https://github.com/nervosnetwork/fiber/pull/691/commits/a6dfa9eb49d72d5085e0b6d41378dfc0ba83304e

because there is a commitment number gap between remote peer and local node:
image

quake (changes requested):
hold for a better solution

src/fiber/channel.rs
```diff
@@ -6256,10 +6244,23 @@ impl ChannelActorState {
...
+        // If the channel is already ready, we should notify the network actor.
+        // so that we update the network.outpoint_channel_map
+        let channel_id = self.get_id();
+        let outpoint = self.must_get_funding_transaction_outpoint();
+        let peer_id = self.get_remote_peer_id();
+        self.network()
+            .send_after(WAITING_REESTABLISH_FINISH_TIMEOUT, move || {
+                NetworkActorMessage::new_event(NetworkActorEvent::ChannelReady(
+                    channel_id,
+                    peer_id.clone(),
```
quake: unnecessary clone?
```suggestion
                    peer_id,
```

src/fiber/graph.rs
```diff
@@ -1612,6 +1614,10 @@ impl PaymentSession {
         !self.request.router.is_empty()
     }
 
+    pub fn first_hop_channel_outpoint(&self) -> Option<OutPoint> {
```
quake: suggest to add an out point parameter since the only use case is to comparison:
```rust
    pub fn first_hop_channel_outpoint_eq(&self, out_point: &OutPoint) -> bool {
        self.route.nodes.first().map(|x| x.channel_outpoint.eq(out_point)).unwrap_or_default()
    }
```

chenyukang: fixed.

quake (approved):

PR #703, opened 2025-05-13
chore: add watchtower feature flag

quake:


chenyukang:
will also need to add feature flag here?
https://github.com/nervosnetwork/fiber/blob/04c1519714f5586bf088409345ff26d2fc3dc57d/crates/fiber-bin/src/main.rs#L16-L18

Otherwise, compiling without watchtower will get an error.

quake:
> will also need to add feature flag here?
> 
no, fiber-bin crate is using fnn-lib with default features, which include the feature flag by default.

chenyukang:
> > will also need to add feature flag here?
> 
> no, fiber-bin crate is using fnn-lib with default features, which include the feature flag by default.

got it.

Officeyutong (approved):

PR #705, opened 2025-05-15
Build support for `wasm32-unknown-unknown`

Officeyutong:
This PR makes fiber able to build for target `wasm32-unknown-unknown`, major changes are:
- async traits are marked by `#[async_trait::async_trait(?Send)]` on wasm, since futures are not `Send`
- RPC parts are changed, API implementation were stripped from jsonrpsee-generated traits. Since traits generated by jsonrpsee are not Send, but APIs could be reused on wasm
- Some calls (except for watchtower part) to CKB API client are changed to use `CKBAPIAsyncClient`, to avoid use sync APIs on wasm
- Currently this PR uses a ractor fork at https://github.com/Officeyutong/ractor/tree/use-non-send-future-in-wasm, since the PR described at todo hasn't been merged yet
- ~~Some sleeps in e2e tests (watchtower and router-pay) are adapted due to changes introduced by async API~~

TODO:
This PR requires some changes to `ractor`, related PRs are waiting for review
- https://github.com/slawlor/ractor/pull/355


Officeyutong:
For now, this PR is ready to be reviewed

- ractor and ckb-sdk in `Cargo.toml` are now referenced by github repo. These will be replaced by referencing through versions after new versions been published.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1686,18 +1687,20 @@ where
             }
         }
         state.retryable_task_last_run_at = Some(now_timestamp_as_millis_u64());
-        let mut pending_tlc_ops = state.tlc_state.get_pending_operations();
-        pending_tlc_ops.retain_mut(|retryable_operation| {
+        let mut check = async |retryable_operation: &mut RetryableTlcOperation| {
```
chenyukang: This closure applies the tlc operations, so maybe `apply_tlc_op` is a better name.

chenyukang:
@gpBlockchain does this branch introduce performance difference?

crates/fiber-lib/src/fiber/channel.rs
Officeyutong: Fixed

chenyukang (approved):

sunchengzhu:
> @gpBlockchain does this branch introduce performance difference?

@chenyukang @gpBlockchain The performance difference between the wasm-try branch and the latest develop branch is not significant, but both are much worse than fiber 0.5.1.
When you have a moment, it might be worth investigating which PR led to this issue.

TPS vs  Threads

crates/fiber-lib/src/build.rs
```diff
@@ -107,7 +109,11 @@ fn main() {
             }
         }
     }
-
+    if latest_db_version.is_empty() {
+        // If there is no migrations, `latest_db_version` is set to today.
```
quake: We can't do that here, let's say the user is running on date A, but we add a migration script before date A (which hasn't been released yet) to our development branch, then new version is released, the migration will be ignored.

quake (changes requested):

crates/fiber-lib/src/build.rs
Officeyutong: Fixed

quake (approved):

PR #711, opened 2025-05-20
Reject invalid open channel message

chenyukang:


quake (approved):

quake:
can we add an unit test?

chenyukang:
> can we add an unit test?

we need to modify the code to reproduce this scenario.
https://github.com/nervosnetwork/fiber/blob/f67e096d45c8a0aa85517866372bf61d438b4fba/crates/fiber-lib/src/fiber/network.rs#L2567

jjyr (approved):

PR #713, opened 2025-05-21
Check shutdown fee rate for shutdown peer message

chenyukang:
If a malicious node sends `Shutdown` command with a too large `feerate`, previously peer node will only update ChannelState to `ShuttingDown` and will not auto accept shutdown, so the state will remain in the state of `ShuttingDown`.

This PR will check the fee rate from peer node, and if the calculated fee exceeded, it will report with proper error message and do not update channel state.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -1124,6 +1124,12 @@ where
             }
         };
 
+        if !state.check_shutdown_fee_valid(shutdown.fee_rate.as_u64()) {
+            return Err(ProcessingChannelError::InvalidParameter(
+                "Shutdown fee is invalid".to_string(),
+            ));
+        }
+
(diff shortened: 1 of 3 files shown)


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -5388,7 +5387,17 @@ impl ChannelActorState {
...
+    }
+
+    fn check_valid_to_auto_accept_shutdown(&self) -> bool {
+        let Some(remote_fee_rate) = self.remote_shutdown_info.as_ref().map(|i| i.fee_rate) else {
+            return false;
+        };
+        if remote_fee_rate < self.commitment_fee_rate {
+            return false;
+        }
+        true
```
quake:
```suggestion
    self.remote_shutdown_info
        .as_ref()
        .map_or(false, |i| i.fee_rate >= self.commitment_fee_rate)
```

quake (approved):

PR #715, opened 2025-05-23
update doc: Testnet Public Nodes User Manual

sunchengzhu:
1. Do not use a specific version of fnn
2. Update the startup command

### DIFF
--- a/docs/testnet-nodes.md
+++ b/docs/testnet-nodes.md
@@ -19,12 +19,14 @@
 
 1. Download fnn
 
-   Example for macOS (Apple silicon):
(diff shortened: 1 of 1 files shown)


docs/testnet-nodes.md
```diff
@@ -67,8 +66,10 @@
 
 5. Start the node A
 
+   You need to set a `FIBER_SECRET_KEY_PASSWORD` environment variable in the startup command to encrypt your wallet private key file. In this example, I used '123', but you can choose any password.
```
chenyukang:
```suggestion
   You need to set a `FIBER_SECRET_KEY_PASSWORD` environment variable in the startup command to encrypt your wallet private key file. I used `123` here for demo purposes, but I recommend using a strong password.
```

chenyukang (approved):

PR #717, opened 2025-05-27
Add check for invalid amounts

chenyukang:
- Add amounts checks
- Remove `get_total_udt_amount`, since it's same with `get_liquid_capacity`


quake (approved):

quake:
there is a clippy error

PR #719, opened 2025-05-28
Add feature vector

### LINKED ISSUE #664: How to get the version of other fiber nodes
Different versions of Fiber may not be compatible with each other, as highlighted in the v0.5.0 release notes, this version includes breaking changes to the contract and configuration file. Without knowing the node versions, it's impossible to ensure that all nodes in the testnet are running compatible software, leading to potential issues like:

- Inability to establish channels: Nodes running incompatible versions might fail to establish new channels or process transactions correctly.

- Unexpected behavior: Mismatched versions can result in unpredictable behavior and make it difficult to do various tests
  

## Questions

- [ ] What is the versions of [the public testnet nodes](https://github.com/nervosnetwork/fiber/blob/73705222d2d7288d1a709919b32e1c4f3335c0cf/docs/testnet-nodes.md)?

- [ ] Knowing the exact versions allows others to reproduce the test environment accurately. Is there a simple way to get the version of other nodes in the network? e.g. https://testnet.explorer.nervos.org/fiber/graph/nodes

chenyukang:
1. Add FeatureVector for each node
2. Connected peers need to send Init Message to each before open channel.
3. Add proper checks on features for compatibility issues
4. Add version in node announcement and close #664


jjyr (approved):

chenyukang:
Need a migration:

```console
migration-check -s ./crates/fiber-lib/src -o ./crates/fiber-lib/src/store/.schema.json
Type fingerprint changed: NodeAnnouncement ac4c8a05c83ea0ee9081879c08840eeec0a7e45b7e948e4519f81061dd14539f -> e31283839981f3222888f8b5eaaf476f194b5cd99295d0e23ea848db74d28d7c
Type dependency chain:
  BroadcastMessage -> NodeAnnouncement
```

chenyukang:
Add migration for invoice:

```console
CkbInvoice -> InvoiceData -> Attribute
```

crates/fiber-lib/src/fiber/config.rs
```diff
@@ -507,6 +509,13 @@ impl FiberConfig {
         self.sync_network_graph
             .unwrap_or(DEFAULT_SYNC_NETWORK_GRAPH)
     }
+
+    pub fn gen_node_features(&self) -> FeatureVector {
```
doitian, line 513: What's the usage of this function?

chenyukang: https://github.com/nervosnetwork/fiber/blob/9b9c02c6de31555e7934327415a96406573a9e3a/crates/fiber-lib/src/fiber/network.rs#L3615

here used to initialize features.

doitian: Why two sets of "default" feature vectors?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -227,6 +232,8 @@ pub enum NetworkActorCommand {
     MaintainConnections,
     // Check all channels and see if we need to force close any of them or settle down tlc with preimage.
     CheckChannels,
+    // Check peer send us Init message in a expected time, otherwise disconnect with the peer.
```
quake:
```suggestion
    // Check peer send us Init message in an expected time, otherwise disconnect with the peer.
```

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -851,11 +858,22 @@ where
 
     pub async fn handle_peer_message(
         &self,
+        myself: ActorRef<NetworkActorMessage>,
         state: &mut NetworkActorState<S>,
         peer_id: PeerId,
         message: FiberMessage,
     ) -> crate::Result<()> {
         match message {
+            FiberMessage::Init(init_message) => {
```
quake, line 883: If the remote node hasn't sent us an init message, it looks like the current PR will still handle other messages from the remote node ? (e.g., OpenChannel, etc.).

crates/fiber-lib/src/fiber/config.rs
chenyukang: here I left a place for configuration file to override default feature vector.

crates/fiber-lib/src/fiber/network.rs
chenyukang: fixed https://github.com/nervosnetwork/fiber/pull/719/commits/4c8db436c3d152184e7dd546cf742a31884d0a8d

quake (approved):

PR #720, opened 2025-05-29
Fix amounts overflow issue in funding tx and code refactor

chenyukang:
1. `found_udt_amount += cell_udt_amount` will overflow when two UDT cells values are too large, it may hard to happen in real world seceario, but it's better to take care.
2. When construct a funding tx, only add changed outputs when `change_amount > 0`
3. Some trivial code cleanup.


chenyukang:
@gpBlockchain

gpBlockchain:
LGTM

quake (approved):

PR #728, opened 2025-06-04
fix: resolve handle_reestablish_channel_message overlfow

quake:


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -6543,6 +6543,11 @@ impl ChannelActorState {
                             ))
                             .expect(ASSUME_NETWORK_ACTOR_ALIVE);
                     }
+                } else if peer_local_commitment_number == u64::MAX {
```
chenyukang: should we add a check at the beginning of this function?

```rust
                if peer_local_commitment_number.abs_diff(my_remote_commitment_number) > 1
                    || peer_remote_commitment_number.abs_diff(my_local_commitment_number) > 1
                {
                    return Err(ProcessingChannelError::InvalidParameter(
                        "reestablish channel message with invalid commitment numbers".to_string(),
                    ));
                }
```

quake: it's good idea, updated, please help to review again, thanks.

chenyukang (approved):

PR #730, opened 2025-06-04
Support MPP

jjyr:


crates/fiber-lib/src/fiber/graph.rs
```diff
@@ -21,6 +21,7 @@ use crate::fiber::types::PaymentHopData;
 use crate::invoice::CkbInvoice;
 use crate::now_timestamp_as_millis_u64;
 use ckb_types::packed::{OutPoint, Script};
+use lightning_invoice::payment;
```
chenyukang: unused import

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -851,7 +868,21 @@ where
...
+            .list_hold_tlcs_by_payment_hash(&tlc_info.payment_hash)
+            .collect();
         if let Some(invoice) = self.store.get_invoice(&tlc.payment_hash) {
+            // TODO check if tlc is MPP
+            // TODO check if invoice support MPP
+            let total_amount = tlcs.iter().map(|tlc| tlc.amount).sum::<u128>();
+
+            let is_fulfilled = total_amount >= invoice.amount.unwrap_or_default();
+            if !is_fulfilled {
+                // TODO add to hold tlc list
```
chenyukang: why do we need to add to hold tlc list?
I assume the last tlc is settled, the whole MPP payment will be settled?

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2724,6 +2763,8 @@ pub struct TlcInfo {
     pub tlc_id: TLCId,
     pub amount: u128,
     pub payment_hash: Hash256,
+    /// The attempt id associate with the tlc, only on outbound tlc
```
chenyukang, line 2796: only used for the first hop's outbound tlc.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -4969,6 +5025,13 @@ impl ChannelActorState {
         self.tlc_state.get(&tlc_id)
     }
 
+    pub fn list_hold_tlcs_by_payment_hash<'a>(
```
chenyukang: seems redudant wrapper, call `state.tlc_state.list_hold_tlcs_by_payment_hash` directly is ok.

crates/fiber-lib/src/fiber/graph.rs
```diff
@@ -1957,3 +1968,9 @@ impl Attempt {
         !self.is_settled() && !self.is_failed()
     }
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HoldTlc {
+    pub channel_actor_state_id: Hash256,
+    pub tlc_id: u64,
```
chenyukang, line 1989: HoldTlc is not part of graph, maybe it's better to move it to types.

PR #732, opened 2025-06-06
Support Basic MPP

chenyukang:
1. Support basic MPP
2. Add payment attempts for the payment session
3. Refactor onion packet peeling in channel actor


chenyukang:
Need migration:

```console
Type dependency chain:
  ChannelActorState -> TlcState -> PendingTlcs -> TlcInfo
```

```console
PaymentSession
```

crates/fiber-lib/src/fiber/types.rs
```diff
@@ -1570,6 +1568,7 @@ pub enum TlcErrorCode {
     ChannelDisabled = UPDATE | 20,
     ExpiryTooFar = PERM | 21,
     InvalidOnionPayload = PERM | 22,
+    HoldTlcTimeout = 23,
```
chenyukang: use `PERM | 23` here, indicates it's a permanent error.

migrate/src/migrations/mig_20250617.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+fn convert_send_payment_data(request: OldPaymentData) -> NewPaymentData {
+    let mut value = serde_json::to_value(&request).unwrap();
+    let obj = value.as_object_mut().unwrap();
+    obj.insert("allow_mpp".to_string(), serde_json::json!(true));
+    serde_json::from_value(value).unwrap()
+}
+
+fn convert_payment_session_status(status: OldPaymentSessionStatus) -> NewPaymentSessionStatus {
```
chenyukang: use `use crate::util::convert` to avoid impl specific for this type.

migrate/src/migrations/mig_20250617.rs
```diff
@@ -0,0 +1,253 @@
...
+        _pb: Arc<dyn Fn(u64) -> ProgressBar + Send + Sync>,
+    ) -> Result<&'a Store, Error> {
+        info!(
+            "MigrationObj::migrate to {} ...........",
+            MIGRATION_DB_VERSION
+        );
+
+        // NodeAnnouncement is changed, we need to delete the old broadcast messages
+        info!("clear broadcast messages ...");
+        delete_broadcast_messages(db);
```
chenyukang: we now use `load_from_store` to build network graph when starting a node, so if we delete all boradcast messages, network graph will be empty.

https://github.com/nervosnetwork/fiber/blob/a3e022e67d245992089f62775b2e1c88aa9fc857/crates/fiber-lib/src/fiber/graph.rs#L628-L638

I'm not sure what the consequence of this.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -832,54 +846,167 @@ where
...
+                    // fail tlc if invoice is not fulfilled
+                    if !is_fulfilled {
+                        return TlcSettleDownResult::Failed(
+                            tlcs,
+                            RemoveTlcReason::RemoveTlcFail(TlcErrPacket::new(
+                                TlcErr::new(TlcErrorCode::AmountBelowMinimum),
+                                &tlc.shared_secret,
+                            )),
+                        );
+                    }
```
chenyukang, line 934: this will never reached for non MPP payment? since we have check before this on line: 899.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -832,54 +846,167 @@ where
...
+            }
+            _ => {
+                let total_amount = tlc.total_amount.unwrap_or(tlc.amount);
+
+                if total_amount < invoice.amount.unwrap_or_default() {
+                    return TlcSettleDownResult::Failed(
+                        tlcs,
+                        RemoveTlcReason::RemoveTlcFail(TlcErrPacket::new(
+                            TlcErr::new(TlcErrorCode::AmountBelowMinimum),
+                            &tlc.shared_secret,
```
chenyukang: these part of error check should removed `apply_add_tlc_operation`, so if error happended, the later `process_add_tlc_error` will take care of remove tlc operation(it also retryable), you don't need the types like `TlcSettleDownResult`.

quake (approved):

PR #733, opened 2025-06-10
Database implementation for fiber wasm

Officeyutong:
This PR adds database (IndexedDB) implementation for fiber wasm (in browser). The database implementation is similar to ckb-light-client:
- A worker runs independently to operate on IndexedDB, to avoid transactions being commited unexpectedly when returning from sync calling stack
- The main worker communicate with the database worker through Atomics API and SharedArrayBuffer

Currently, ractor is using my fork, since ractor doesn't has full support for wasm futures


crates/fiber-lib/src/ckb/config.rs
```diff
@@ -76,7 +87,11 @@ impl CkbConfig {
             Ok(())
         }
     }
-
+    #[cfg(target_arch = "wasm32")]
+    pub fn read_secret_key(&self) -> Result<SecretKey> {
+        Ok(self.wasm_secret_key.expect("SecretKey not found on wasm"))
```
chenyukang, line 91: is it in plain format? do we have the security issue which we introduced ENV_FIBER_SECRET_KEY_PASSWORD to resolve.

crates/fiber-lib/src/ckb/config.rs
```diff
@@ -2,11 +2,17 @@ use ckb_hash::blake2b_256;
...
-use tracing::info;
+#[cfg(not(target_arch = "wasm32"))]
+use std::fs;
 
+#[cfg(not(target_arch = "wasm32"))]
 use crate::utils::encrypt_decrypt_file::{decrypt_from_file, encrypt_to_file};
-use crate::{Error, Result};
+#[cfg(not(target_arch = "wasm32"))]
+use crate::Error;
+use crate::Result;
```
chenyukang:
```suggestion
use crate::{Error, Result};
```

crates/fiber-lib/src/ckb/config.rs
Officeyutong: @quake Do you have any idea about providing secret key in a more secure way on wasm? Currently it's passed through argument in plain text, since there is no filesystem nor environment variable in browser

quake: we may consider saving the encrypted key to local storage.

quake: let's do it in another  PR.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1996,7 +1995,10 @@ where
                 } else {
                     None
                 };
-
+                trace!(
```
quake: please remove this debug code

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -3602,7 +3604,7 @@ pub struct ChannelActorState {
     pub reestablishing: bool,
     pub last_revoke_ack_msg: Option<RevokeAndAck>,
 
-    pub created_at: SystemTime,
+    pub created_at: crate::time::SystemTime,
```
quake: please use `import` instead of full crate path

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1161,7 +1164,7 @@ where
             NetworkActorCommand::ConnectPeer(addr) => {
                 // TODO: It is more than just dialing a peer. We need to exchange capabilities of the peer,
                 // e.g. whether the peer support some specific feature.
-
+                debug!("Received ConnectPeer for connecting to {:?}", addr);
```
quake: please remove these debug/trace code.

crates/fiber-lib/src/fiber/key.rs
```diff
@@ -1,11 +1,12 @@
 use ckb_hash::new_blake2b;
+use serde::{Deserialize, Serialize};
 use std::{fs, path::Path};
 use tracing::warn;
 
 // TODO: we need to securely erase the key.
 // We wrap the key in a struct to obtain create a function to obtain secret entropy from this key.
 // Unfortunately, SecioKeyPair does not allow us to obtain the secret key from the key pair.
-#[derive(Clone)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
quake, line 9: why we need to de/serialize key pair here?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -700,7 +700,10 @@ macro_rules! debug_event {
 
 #[derive(Clone, Debug)]
 pub enum NetworkServiceEvent {
-    NetworkStarted(PeerId, MultiAddr, Vec<Multiaddr>),
+    #[cfg(not(target_arch = "wasm32"))]
+    NetworkStarted(PeerId, Vec<MultiAddr>, Vec<Multiaddr>),
```
quake, line 710: no need to define different var here, we may use empty vec for announced addr on wasm32

crates/fiber-lib/src/utils/encrypt_decrypt_file.rs
```diff
@@ -40,7 +42,11 @@ pub fn encrypt_to_file<P: AsRef<Path>>(
     fs::write(file, file_bytes).map_err(|err| format!("failed to write to file: {}", err))
 }
 
-pub fn decrypt_from_file<P: AsRef<Path>>(file: P, password: &[u8]) -> Result<Vec<u8>, String> {
+pub fn decrypt_from_file<P: AsRef<Path> + Debug>(
+    file: P,
+    password: &[u8],
+) -> Result<Vec<u8>, String> {
+    debug!("Decrypting key from file {:?}", file);
```
quake: please revert this debug change

Officeyutong: Fixed

crates/fiber-lib/src/fiber/network.rs
Officeyutong: Fixed

crates/fiber-lib/src/fiber/channel.rs
Officeyutong: Fixed

crates/fiber-lib/src/fiber/channel.rs
Officeyutong: Fixed

crates/fiber-lib/src/fiber/network.rs
Officeyutong: Fixed

crates/fiber-lib/src/fiber/key.rs
Officeyutong: `pub wasm_key_pair: Option<KeyPair>` is in `struct FiberConfig` to provide fiber key on wasm, though it has attribute `#[arg(skip)]` to avoid appear in command line, `ClapSerde` still requires it to implement `Serialize` and `Deserialize`

quake (approved):

PR #734, opened 2025-06-11
feat: auto close funding channel on timeout

doitian:
When there's no signature sent yet, it's safe to abort the funding channel and forget it.

The default timeout in this PR is 1 day.

It is configurable via config file, command line argument or environment variable.

Channel does not read config before, so I changed the channel startup args a bit to include a ephemeral config struct which is not persisted along with the channel state. Suggestions are recommended for a better way to pass config options to channel actors.


doitian:
~~TODO: need to make the timeout configurable for testing~~

Done

doitian:
@gpBlockchain ready for review and test

quake (approved):

PR #736, opened 2025-06-12
feat: store pub-sub

doitian:
Add store pub-sub to emit update events.

- Add an actor StorePublisher to accepts events from database, fetch more rich information and broadcast these events to subscribers.
- Add StoreWithPubSub to emit store update events to StorePublisher.
- Replace Store with StoreWithPubSub on app boot.

This PR is a refactoring of the most controversial part of #511 to avoid introducing too many changes to store traits and the Store struct. Instead, a new store struct StoreWithPubSub is added as a wrapper of Store to emit update events.

A family of higher-layer traits `*StoreDeref` are added to simplify forwarding store traits to the wrapper store struct. This can be used for future store decorators.


crates/fiber-lib/src/store/tests/store_with_pub_sub_unit_tests.rs
```diff
@@ -0,0 +1,147 @@
...
+    store.subscribe(Box::new(subscriber_ref));
+    store.insert_payment_session(session);
+
+    ractor::concurrency::timeout(Duration::from_millis(100), subscriber_handle)
+        .await
+        .expect("Test actor failed in exit")
+        .unwrap();
+}
+
+#[tokio::test]
```
Officeyutong: If a unit test is expected to work on wasm, you can use the following attribute macro to replace `#[tokio::test]` to make wasm CI runner to execute the unit test:
```
#[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
```

doitian: Currently PubSub is non-WASM only.

chenyukang (commented):
the preimage issues

crates/fiber-lib/src/store/pub_sub/store_with_pub_sub.rs
```diff
@@ -0,0 +1,286 @@
...
+    fn publish_payment_updated_event_when_inserting_payment_session(
+        &self,
+        payment_hash: Hash256,
+        status: PaymentStatus,
+    ) {
+        let payload_opt = status.try_into().ok().or_else(|| match status {
+            // If preimage is not available, defer the notification on `insert_preimage`.
+            // See `publish_payment_updated_event_when_removing_preimage`.
+            PaymentStatus::Success => self
+                .get_preimage(&payment_hash)
```
chenyukang, line 147: please note we only call insert_preimage for a payment for middle hops of a payment path,
so it only make sense to call get_preimage for these hops,
and we only have payment session in the payment sender.

so i'm not sure how can we get_preimage when inserting payment session?

crates/fiber-lib/src/store/pub_sub/subscription.rs
```diff
@@ -0,0 +1,144 @@
...
+
+// but with additional information for downstream services.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum PaymentUpdatedPayload {
+    /// initial status, payment session is created, no HTLC is sent
+    Created,
+    /// the first hop AddTlc is sent successfully and waiting for the response
+    Inflight,
+    /// related HTLC is successfully settled
+    Success { preimage: Hash256 },
```
chenyukang, line 71: for MPP/AMP, how to specify this preimage field?
now after merging MPP, one payment may have multiple payment attempts, for basic MPP they share the same preimage,
but for AMP, different payment attempts using different preimages.

crates/fiber-lib/src/store/pub_sub/store_with_pub_sub.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    fn update_invoice_status(
+        &self,
+        id: &Hash256,
+        status: CkbInvoiceStatus,
+    ) -> Result<(), InvoiceError> {
+        let _span = tracing::info_span!("update_invoice_status", invoice_hash = ?id).entered();
+
+        self.inner.update_invoice_status(id, status)?;
+        let payload_opt = status.try_into().ok().or_else(|| match status {
```
chenyukang, line 94: how about we add a field(something like `event_publisher`) in `Store`, and keep to use store here, so that when we need to notify some events, we just call `event_publisher.send_event` here.

seems the current implementation mixed the logic of sending events with current storing logic here.

and we also need to add delegating for all kinds of store, seems not good.

doitian: Seems like a good idea. We can do transform in another layer of pub/sub.

crates/fiber-lib/src/store/pub_sub/subscription.rs
doitian: CCC supports only single path now. MPP/AMP needs extra works since lnd should also use MPP/AMP and the same number of paths.

doitian:
Close and reword based on #930

PR #746, opened 2025-06-17
Check peer and channel map for channel messages

chenyukang:


crates/fiber-lib/src/fiber/network.rs
```diff
@@ -897,13 +897,25 @@ where
...
             }
-            FiberMessage::ChannelNormalOperation(m) => {
-                let channel_id = m.get_channel_id();
+            FiberMessage::ChannelNormalOperation(msg) => {
+                let channel_id = msg.get_channel_id();
+                state
+                    .peer_session_map
+                    .get(&peer_id)
+                    .and_then(|(session, ..)| state.session_channels_map.get(session))
+                    .and_then(|channels| channels.get(&channel_id))
```
quake: It might be more intuitive to replace it with `channels.contains`

quake (approved):

crates/fiber-lib/src/fiber/network.rs
chenyukang: fixed it.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -897,13 +897,27 @@ where
...
-                let channel_id = m.get_channel_id();
+            FiberMessage::ChannelNormalOperation(msg) => {
+                let channel_id = msg.get_channel_id();
+
+                let found = state
+                    .peer_session_map
+                    .get(&peer_id)
+                    .and_then(|(session, ..)| state.session_channels_map.get(session))
+                    .map(|channels| channels.contains(&channel_id))
+                    .unwrap_or(false);
```
quake:
```suggestion
.map_or(false, |channels| channels.contains(&channel_id));
```

chenyukang: I find `is_some_and` is better.

jjyr (approved):

PR #747, opened 2025-06-18
feat: throttle to be accepted channels

doitian:
Limit the total number and bytes of open channel requests that a peer can send us.

- The default number limit is 20
- The default storage size limit is 50KB

When a peer reaches the limit, following open channel requests are rejected.

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -678,7 +678,7 @@ pub enum NetworkServiceEvent {
     PeerDisConnected(PeerId, Multiaddr),
     // An incoming/outgoing channel is created.
     ChannelCreated(PeerId, Hash256),
-    // A outgoing channel is pending to be accepted.
+    // An incoming channel is pending to be accepted.
     ChannelPendingToBeAccepted(PeerId, Hash256),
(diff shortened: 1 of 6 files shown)


crates/fiber-lib/src/fiber/network.rs
```diff
@@ -3232,6 +3234,29 @@ where
...
+            Err(ProcessingChannelError::RepeatedProcessing(_)) => {
+                // Ignore repeated processing error and do not send any notifications
+                return Ok(());
+            }
+            Err(_) => NetworkServiceEvent::ChannelPendingToBeRejected(peer_id, id),
+        };
+
+        self.network
+            .send_message(NetworkActorMessage::new_notification(event))
+            .expect(ASSUME_NETWORK_MYSELF_ALIVE);
```
chenyukang: is `ChannelPendingToBeRejected` only used for testing or debugging?
maybe using `debug_event!` is enough.

doitian: 👍

chenyukang (approved):

chenyukang (approved):

quake (approved):

PR #751, opened 2025-06-25
fix: resolve public token reused issue

quake:


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -450,6 +459,7 @@ where
                 state.handle_tx_collaboration_msg(TxCollaborationMsg::TxComplete(tx))?;
                 if let ChannelState::CollaboratingFundingTx(flags) = state.state {
                     if flags.contains(CollaboratingFundingTxFlags::COLLABORATION_COMPLETED) {
+                        state.increment_local_commitment_number();
```
doitian, line 493: Why this is absent before but added in this PR?

quake: In the previous code, because the commitment number was not updated, the first round the commitment tx was updated after the channel was created, the previous nonce was reused.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -6471,22 +6554,7 @@ impl ChannelActorState {
                 {
                     // commitments are the same, sync up the tlcs
                     self.set_waiting_ack(myself, false);
-                    self.resend_tlcs_on_reestablish(false)?;
-
-                    // there is a scenario that two peers are both in WaitingAck state
-                    // and if two parties send CommitmentSigned message to each other there maybe be a Musig2VerifyError
-                    // eventually, we need to keep the linearity of the signing and verification operation
-                    // so here we pick a order that the acceptor will send the CommitmentSigned message first
-                    if self.is_acceptor {
```
chenyukang, line 6789: why we remove this logic here?

quake: In this PR, there will no longer be a situation where both sides are in `WaitingAck`, so I don't think this is necessary, and I'll see if I can add a unit test to simulate it.

chenyukang (approved):

PR #759, opened 2025-06-27
Avoid wait too long for payment status in unit tests

chenyukang:
- Some unit tests may hang forever when the are unexpected behaviors, let's add a timeout for `wait_until_*` so that we know which test case goes wrong.
- Fix a possible panic with new test case `test_send_payment_shutdown_channel_actor_may_already_stopped`.

### DIFF
--- a/crates/fiber-lib/src/fiber/tests/channel.rs
+++ b/crates/fiber-lib/src/fiber/tests/channel.rs
@@ -2768,6 +2768,8 @@ async fn do_test_add_tlc_waiting_ack() {
         }
     }
 
+    tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
+
     // send from b to a
     for i in 1..=2 {
         let add_tlc_command = AddTlcCommand {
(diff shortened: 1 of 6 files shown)


crates/fiber-lib/src/fiber/tests/channel.rs
```diff
@@ -2768,6 +2768,8 @@ async fn do_test_add_tlc_waiting_ack() {
         }
     }
 
+    tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
jjyr, line 2771: Why sleep here?

chenyukang: don't want previous part effect following loop.

chenyukang: I found it have a very low possibility this may happen in CI machine.

crates/fiber-lib/src/fiber/tests/payment.rs
```diff
@@ -3499,7 +3501,6 @@ async fn test_send_payment_with_one_node_stop() {
     let mut check_count = 0;
     while check_count < 100 {
         for payment_hash in all_sent.clone().iter() {
-            nodes[0].wait_until_final_status(*payment_hash).await;
```
chenyukang, line 3502: @jjyr this can not use `wait_until_final_status`, because a payment may left in `InFlight` state, it depends on the event order, for example when RemoveTlc is passing from receiver to sender, and the middle hop node stopped, the sender will not get the final status of this payment. 

For this test case, we only assert if a middle hop stopped, the payment sent later than the stopped time point will be failed.

chenyukang: otherwise, `wait_until_final_status` will be hang.

quake (approved):

PR #762, opened 2025-07-01
Add unit tests for reestablsihing

chenyukang:


crates/fiber-lib/src/fiber/tests/payment.rs
```diff
@@ -4405,6 +4412,98 @@ async fn test_send_payment_remove_tlc_with_preimage_will_retry() {
     }
 }
 
+#[tokio::test]
+async fn test_send_payment_send_each_other_reestablishing() {
```
chenyukang, line 4418: This test will run into here, and then get a signature error.

https://github.com/nervosnetwork/fiber/blob/348dfb90a38eef4cef9e163104c2f45752b23cd9/crates/fiber-lib/src/fiber/channel.rs#L6531-L6534

chenyukang: the posibility is 50% to fail on my machine.

quake (approved):

PR #766, opened 2025-07-02
Add limit for protocal service data length

chenyukang:


crates/fiber-lib/src/fiber/network.rs
```diff
@@ -590,6 +593,18 @@ impl SendPaymentData {
             ));
         }
 
+        if let Some(custom_records) = &command.custom_records {
+            if custom_records.data.keys().len() > 32 {
```
jjyr: IMO limit on binary size is enough to stop malicious attack. 32 is too small to exceed.

chenyukang: removed.

jjyr (approved):

PR #770, opened 2025-07-04
chore: sync Cargo.lock

doitian:
bech32 has been downgraded in 04c1519714f5586bf088409345ff26d2fc3dc57d


chenyukang:
it's better we upgrade bech32 in fiber-lib/Cargo.toml to 0.9?

PR #776, opened 2025-07-08
Fix overflow issues in amounts for adding tlc

chenyukang:


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -5020,13 +5015,27 @@ impl ChannelActorState {
...
                 (self.to_local_amount, self.to_remote_amount);
             if current.is_offered() {
+                if to_local_amount < current.amount {
+                    return Err(ProcessingChannelError::InternalError(format!(
+                        "Cannot remove tlc {:?} with amount {} from local balance {}",
+                        tlc_id, current.amount, to_local_amount
+                    )));
+                }
                 to_local_amount -= current.amount;
                 to_remote_amount += current.amount;
```
jjyr: we can use checked version ops here, its better than check amount before each ops.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -5551,6 +5567,13 @@ impl ChannelActorState {
                 return Err(ProcessingChannelError::TlcValueInflightExceedLimit);
             }
         } else {
+            // remote peer can not sent more tlc amount than they have
+            let received_tlc_value = self.get_received_tlc_balance();
+            debug_assert!(self.to_remote_amount >= received_tlc_value);
+            if add_amount > self.to_remote_amount - received_tlc_value {
```
jjyr: Use checked version arithmetic ops, or saturating_xxx

jjyr (approved):

PR #785, opened 2025-07-15
feat: funding tx testing

doitian:
Allow building funding tx via an external shell script so it's easier to inject invalid funding tx in tests.

PR #794, opened 2025-07-23
feature/funding tx verification

doitian:
Based on #785 and #792


doitian:
Future works:

- Add funding tx size limit to avoid spending too much on tx fee.
- Add test cases using #785

quake (approved):

PR #806, opened 2025-07-30
Support unit tests on wasm

Officeyutong:
This PR adapts some unit tests for wasm, and adds a workflow file to execute them.
- Unittests without database access and network access can be directly used in wasm
- Unittests with database access will use a in-memory database implementation on wasm, since using original wasm databae implementation is hard for wasm-bindgen-test
- Unittests with network access (such as RPC server) or filesystem access (such as reading configuration) can't be adapted for wasm


gpBlockchain:
lgtm

quake (approved):

PR #827, opened 2025-08-11
Atomic mpp

chenyukang:
Implementation for https://github.com/nervosnetwork/fiber/issues/911


crates/fiber-lib/src/fiber/types.rs
```diff
@@ -3784,6 +3785,64 @@ impl BasicMppPaymentData {
...
 
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub struct AmpPaymentData {
+    pub total_amp_count: u16,
+    pub payment_hash: Hash256,
+    pub child_desc: AmpChildDesc,
+}
+
+impl AmpPaymentData {
+    pub const CUSTOM_RECORD_KEY: u32 = USER_CUSTOM_RECORDS_MAX_INDEX + 2;
```
jjyr: Maybe we should use a single module for example `builtin_records.rs` to store all builtin recrod key.

crates/fiber-lib/src/invoice/invoice_impl.rs
```diff
@@ -681,36 +705,47 @@ impl InvoiceBuilder {
...
-            payment_hash
-        } else {
-            // generate a random payment hash if not provided
-            gen_rand_sha256_hash()
+    pub fn allow_atomic_mpp(self, allow_atomic_mpp: bool) -> Self {
+        self.update_feature_vector(|feature_vector| {
+            if allow_atomic_mpp {
+                feature_vector.set_atomic_mpp_required();
+            } else {
+                feature_vector.unset_basic_mpp_optional();
```
jjyr: typo?

crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -272,6 +273,12 @@ impl StoreKeyValue for KeyValue {
                 &attempt_id.to_le_bytes(),
             ]
             .concat(),
+            KeyValue::AttemptToPaymentHash((attempt_hash, attempt_id), _) => [
```
jjyr, line 306: We can get payment_hash from attempt struct if not consider the performance of loading the attempt.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -895,62 +897,69 @@ where
...
                     // this maybe happened when process is killed and restart
                     error!("invoice already paid, ignore");
                     return;
                 }
-                _ if invoice.allow_mpp() => {
+                _ if mpp_mode.is_some() => {
                     // add to pending settlement tlc set
                     // the tlc set will be settled by network actor
                     state
                         .pending_notify_mpp_tcls
```
jjyr:
```suggestion
                        .pending_notify_mpp_tlcs
```

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2923,6 +2960,8 @@ pub struct TlcInfo {
     pub total_amount: Option<u128>,
     /// bolt04 payment secret, only exists for last hop in multi-path payment
     pub payment_secret: Option<Hash256>,
+    /// atomic parent payment_hash
+    pub parent_payment_hash: Option<Hash256>,
```
jjyr, line 3045: How about make the `parent_payment_hash` always exist, but make the amp_payment_hash optional?

The `parent_payment_hash` is related to Invoice or Payment, which technically should be always exist, but amp payment_hash is depend on amp which is optional, if it is empty we just fallback to legacy TLC.

 `parent_payment_hash` -> `payment_hash: H256`, `payment_hash` -> `hash: Option<H256>`.

chenyukang: this is also an choice, but I considering a lot of current code now using `payment_hash` in tlc info, if we choose to use `parent_payment_hash` and `amp_payment_hash`, I guess a lot of code need to adjust.

chenyukang: `parent_payment_hash` now only exists for last hop.

crates/fiber-lib/src/fiber/channel.rs
chenyukang: fix in develop: https://github.com/nervosnetwork/fiber/pull/876

crates/fiber-lib/src/store/store_impl/mod.rs
chenyukang: yeah, seems we can remove this.

chenyukang: emm, no, because to load attempt from db, we need to know the parent payment hash...

chenyukang: it's used here: https://github.com/nervosnetwork/fiber/blob/7828c824fcc204d7a6e472e9f3d56750ffb8cdd0/crates/fiber-lib/src/fiber/network.rs#L1247-L1248

crates/fiber-lib/src/fiber/types.rs
chenyukang: added.

crates/fiber-lib/src/fiber/channel.rs
jjyr: May be we should provide clear interface to access the two fields:
-> payment_hash() -> for query invoice, payment
-> tlc_hash() -> for check preimage

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1654,48 +1707,143 @@ where
...
+                    match mpp_mode {
+                        MppMode::BasicMpp => {
+                            let Some(preimage) = self.store.get_preimage(&payment_hash) else {
+                                validation_fail!(
+                                    "basic MPP can not get preimage for payment: {:?}"
+                                );
+                            };
+
+                            for tlc in tlcs.iter() {
+                                tlc_preimage_map.insert((tlc.channel_id, tlc.id()), preimage);
```
jjyr, line 1943: IMO use enum here may improve the readability.

```
enum TlcSetPreimage {
  BasicMPP(H256),
  AMP(HashMap<_>),
}

impl TlcSetPreimage {
  fn get(tlc: &Tlc) -> H256 {
    //....
  }
}
```

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `a3428c8fdcc58e04a484d355f6d9bee3c697d9c3..2ae33b47ed6d9db2a1824086164a7fb642fd6605`

Validation:
- Focused static review found a persistence migration concern: PR 827 changes serialized payment/channel state (`SendPaymentData`, `Attempt`, and `TlcInfo`) and adds store prefixes without a matching migration/defaulting path, while existing store reads use strict `bincode` deserialization. Existing nodes with pre-PR payment/session/channel rows may fail validation, query, retry/resume, or channel restore after upgrade.

Notes:
- Suggested fix: add a migration or backward-compatible serde/default handling for old rows, including tests that open pre-PR serialized `PaymentSession`, `Attempt`, and channel/TLC state with the PR build.

PR #831, opened 2025-08-12
doc: add README for wasm-test-runner

doitian:
Add guides on how to run the WASM Test in Ubuntu.

### DIFF
--- a/.github/wasm-test-runner/.gitignore
+++ b/.github/wasm-test-runner/.gitignore
@@ -1,1 +1,2 @@
 /node_modules
+/pnpm-lock.yaml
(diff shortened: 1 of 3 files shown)


Officeyutong:
A more easier way is to use `wasm-pack`
```
cargo install wasm-pack
cd crates/fiber-lib
wasm-pack test --firefox --no-default-features
```
When it shows something like `Interactive browsers tests are now available at http://127.0.0.1:8000`, open `http://127.0.0.1:8000` in firefox to execute the tests.

This way requires manually operate with browsers, so it's not suitable in CI

PR #833, opened 2025-08-12
test: e2e for funding tx verification

doitian:
1. Modify tests/nodes/start.sh to accept the e2e test case as command
   line arg, so it can customize the started processes for different
   test cases.
2. Add `rpc_url` and `funding_source_lock_script` to funding tx shell
   builder input. These two fields can simplify external shell builder
   implementation.
3. Move funding tx shell builder option to CKB config and the building logic to
   the CKB Chain Actor. The CKB Chain Actor holds the information of the
   new added fields `rpc_url` and `funding_source_lock_script`.
4. Add a new cargo project in `tests/funding-tx-builder` as external
   funding tx shell builder for e2e testing. It contains 3 test cases
   now, which name is specified via the command line argument
   `FUNDING_TX_VERIFICATION_CASE=name`
    - `remove_change`: remove change output from the peer
    - `modify_change`: modify the capacity of the change output from the
      peer
    - `fund_from_peer`: use live cells from the peer to fund our part of
      the channel
5. Run the new e2e test cases via GitHub Action


crates/fiber-lib/src/ckb/actor.rs
```diff
@@ -301,3 +303,70 @@ impl CkbChainState {
...
+}
+
+#[cfg(target_arch = "wasm32")]
+async fn fund_via_shell(
+    _shell_script: &str,
+    _tx: FundingTx,
+    _request: FundingRequest,
+    _context: FundingContext,
+) -> Result<FundingTx, FundingError> {
+    todo!();
```
quake: why need to implement this fn in wasm32 target?

doitian: No.

quake (approved):

PR #837, opened 2025-08-14
Check channel shutdown periodically

jjyr:
This PR fixes channel force shutdown issue by periodically check for channel shutdown transactions on-chain. 

1. In ChannelActor, the CheckShutdown task will run every 300s to check if the channel is force shutdown by counter-party.
2. A `CloseFlags::CLOSE_BY_REMOTE` is introduce, this flag only meaningful if channel is force closed.
3. A dev API of `check_channel_shutdown` is introduced for testing purpose.
4. E2E tests shutdown-force is modified to test this new change.

--------------------------

The watchtower also watches `commitment_tx` of channels, but watchtower is a relatively separated module, and can not send notify back to the fiber node, especially in RPC watchtower mode. Thus, we decide to use the current solution.


crates/fiber-lib/src/ckb/actor.rs
```diff
@@ -82,6 +82,44 @@ impl From<Option<ckb_jsonrpc_types::TransactionWithStatusResponse>> for GetTxRes
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct GetShutdownTxRequest {
```
chenyukang, line 88: seems there is no need to introduce this new struct type.

chenyukang: never mind, seems it's following message style...

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2196,6 +2204,59 @@ where
                 }
                 Ok(())
             }
+            ChannelCommand::CheckShutdown => {
+                // check channel ready state
+                if state.state == ChannelState::ChannelReady {
```
chenyukang: prefer to add a handle function for this, such as `handle_check_peer_shutdown`

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -164,6 +170,7 @@ pub enum ChannelCommand {
     ForwardTlcResult(ForwardTlcResult),
     #[cfg(any(test, feature = "bench"))]
     ReloadState(ReloadParams),
+    CheckShutdown,
```
chenyukang: maybe `CheckPeerShutdown`

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -860,7 +868,7 @@ pub enum NetworkActorEvent {
     FundingTransactionFailed(OutPoint),
 
     /// A closing transaction has been confirmed.
-    ClosingTransactionConfirmed(PeerId, Hash256, Byte32, bool),
+    ClosingTransactionConfirmed(PeerId, Hash256, Byte32, bool, bool),
```
chenyukang, line 868: please add commands for fields, otherwise it's hard to understand the meaning of them.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -118,6 +121,9 @@ pub const PEER_CHANNEL_RESPONSE_TIMEOUT: u64 = 30 * 1000;
 #[cfg(any(test, feature = "bench"))]
 pub const PEER_CHANNEL_RESPONSE_TIMEOUT: u64 = 10 * 1000;
 
+// The interval time to check shutdown transaction
+pub const CHANNEL_CHECK_SHUTDOWN_INTERVAL: Duration = Duration::from_secs(300);
```
chenyukang: 5 minutes seems too long? 
what's the consideration to choose 5 minute.

crates/fiber-lib/src/ckb/actor.rs
```diff
@@ -301,3 +349,34 @@ impl CkbChainState {
         }
     }
 }
+
+async fn get_shutdown_tx(
+    client: &CkbRpcAsyncClient,
+    GetShutdownTxRequest {
```
chenyukang, line 425: it's better only use `Script` type as the parameter?

crates/fiber-lib/src/rpc/dev.rs
```diff
@@ -102,6 +102,13 @@ pub struct SubmitCommitmentTransactionResult {
     pub tx_hash: Hash256,
 }
```
chenyukang, line 103: seems this rpc is only used for e2e test, maybe you should use a smaller `CHANNEL_CHECK_SHUTDOWN_INTERVAL`, such as 6 seconds, and wait enough time in e2e test to make sure cron job is invoked automatically. and we can remove this rpc.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2196,6 +2204,59 @@ where
...
                 Ok(())
             }
+            ChannelCommand::CheckShutdown => {
+                // check channel ready state
+                if state.state == ChannelState::ChannelReady {
+                    // check shutdown transactions
+                    let request = GetShutdownTxRequest {
+                        funding_lock_script: state.get_funding_lock_script(),
+                    };
+                    match call!(self.network, |tx| NetworkActorMessage::Command(
```
chenyukang: I think this cron job is better moved into Network actor, `call!` is a bad smell for me, since we met deadlock bug because of channel actor send message to network actor and waiting for reply, and at the same time network actor send message to channel actor and waiting for reply. seems moving to network actor could avoid use `call!`.

crates/fiber-lib/src/fiber/channel.rs
jjyr: This check only works when remote force close the channel, which is a very rare event. So most time, this check just query tx from CKB and do nothing.
Even in the case a channel is force shutdown by remote, the user do not need to immediately know it.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -4002,6 +4009,8 @@ bitflags! {
         const ABANDONED = 1 << 2;
         /// Channel is closed because of aborted funding.
         const FUNDING_ABORTED = 1 << 3;
+        /// Indicates the close is initiated by remote, only meaningful when UNCOOPERATIVE is set
+        const CLOSE_BY_REMOTE = 1 << 4;
```
quake: suggest to rename `UNCOOPERATIVE` to `UNCOOPERATIVE_LOCAL` and rename this name to `UNCOOPERATIVE_REMOTE`

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2087,6 +2110,77 @@ where
         Ok(())
     }
 
+    async fn check_channel_shutdown(
```
quake, line 2127: we already have a channel checking job, may be better to move this fn invoking to that job also:

```rust
        myself.send_interval(CHECK_CHANNELS_INTERVAL, || {
            NetworkActorMessage::new_command(NetworkActorCommand::CheckChannels)
        });
```

quake (approved):

PR #839, opened 2025-08-14
perf: refactor retryable task in channel actor to add delay with retry_count

chenyukang:
- The delay for retrying task from forward tlc and remove tlc is `retry_count` seconds.
- WaitingAck error's retry delay is `1.max(retry_count / 120)` seconds.


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1762,183 +1767,158 @@ where
...
+                            reason: reason.clone(),
+                        },
+                    )
+                    .await
+                {
+                    Ok(_) | Err(ProcessingChannelError::RepeatedProcessing(_)) => false,
+                    Err(ProcessingChannelError::WaitingTlcAck) => {
+                        // don't retry too many times, but in most cases we need to retry immediately for WaitingTlcAck
+                        // here we use 2 minutes as decreasing unit
+                        retry_delay = 1.max(retry_count / 120);
```
jjyr: how to understand "use 2 minutes as decreasing unit".
if retry_count is 130, or 150, we still get retry_delay = 1, if retry_count reach 240, then we can get 2. how to understand retry_delay? how is it related to 2 minuts?

jjyr: I figured it out, but still think retry_delay is confused, retry_delay is a intermediate state between retry_count and actually delay, may be we can use a helper function to remove it? 

```
retry_delay(retry_count: usize, waiting_tlc_ack: bool) -> Duration;
```

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -4475,16 +4474,22 @@ impl ChannelActorState {
     fn trigger_retryable_tasks(
         &mut self,
         myself: &ActorRef<ChannelActorMessage>,
-        run_job: Option<(RetryableTlcOperation, u32)>,
+        run_job: Option<(RetryableTlcOperation, u32, bool)>,
```
quake: The design of `trigger_retryable_tasks` function is confusing. When `run_job` is `None`, it actually retries all tasks, but this overloaded behavior makes the function purpose unclear.

I suggest splitting this into two separate functions for better readability:

1. `trigger_single_retryable_task(operation, waiting_ack)` - for retrying a specific task
2. `trigger_all_retryable_tasks()` - for retrying all pending tasks (used during reestablishment)

After splitting `trigger_retryable_tasks` into two separate functions, the `update_retryable_task_retry_count` function can be merged into `trigger_single_retryable_task` for better encapsulation.

Since `update_retryable_task_retry_count` is only used when retrying a single task (not when retrying all tasks), it logically belongs inside the single task retry function.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2761,8 +2743,8 @@ where
         myself: ActorRef<Self::Msg>,
         state: &mut Self::State,
     ) -> Result<(), ActorProcessingErr> {
-        if state.tlc_state.has_pending_operations() && !state.reestablishing {
-            state.trigger_retryable_tasks(&myself, false);
+        if state.has_pending_operations() && !state.reestablishing {
```
quake: If there are pending operations, that means reestablishing must be true, so I think the three lines of code are unnecessary.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -3121,8 +3103,28 @@ impl From<TlcInfo> for TlcNotifyInfo {
...
+            RetryableTlcOperation::RelayRemoveTlc(_, tlc_id, ..) => {
+                "relay_remove_tlc".hash(state);
+                tlc_id.hash(state);
+            }
+            RetryableTlcOperation::ForwardTlc(_, tlc_id, ..) => {
+                "forward_tlc".hash(state);
+                tlc_id.hash(state);
+            }
+        }
+    }
```
quake:
```suggestion
impl std::hash::Hash for RetryableTlcOperation {
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
        std::mem::discriminant(self).hash(state);
        match self {
            RetryableTlcOperation::RemoveTlc(tlc_id, ..)
            | RetryableTlcOperation::RelayRemoveTlc(_, tlc_id, ..)
            | RetryableTlcOperation::ForwardTlc(_, tlc_id, ..) => {
                tlc_id.hash(state);
            }
        }
    }
}
```

crates/fiber-lib/src/fiber/channel.rs
chenyukang: `update_retryable_task_retry_count` is removed.

quake (approved):

PR #862, opened 2025-08-25
fix: check tlc expiry should considering pending_tlc_count

quake:


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -5695,7 +5695,15 @@ impl ChannelActorState {
         let delay_epoch = EpochNumberWithFraction::from_full_value(self.commitment_delay_epoch);
         let epoch_delay_milliseconds =
             (delay_epoch.number() as f64 * MILLI_SECONDS_PER_EPOCH as f64 * 2.0 / 3.0) as u64;
-        let expect_expiry = current_time + epoch_delay_milliseconds;
+        let pending_tlc_count = self
+            .tlc_state
+            .all_tlcs()
```
chenyukang, line 5667: is better to use `get_active_tlcs`?

quake: get_active_tlcs requires a `for_remote` param and it's logic is a little bit different.

chenyukang (approved):

PR #867, opened 2025-08-25
Fix update close transaction when actor exit; remove network signing command

jjyr:
1. Move `ChannelActor` field `private_key` to `ChannelActorState`.
2. Remove `NetworkActorCommand::SignMessage` and `call!` on network.
3. Fix `update_close_transaction_confirmed` hanging issue.
4. Fix force close disconnected peer's channel do not update the state.
5. Update e2e test case

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -345,6 +345,7 @@ pub enum ChannelInitializationOperation {
 pub struct ChannelInitializationParameter {
     pub operation: ChannelInitializationOperation,
     pub ephemeral_config: ChannelEphemeralConfig,
+    pub private_key: Privkey,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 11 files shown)


chenyukang (approved):

crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -380,8 +380,10 @@ impl NetworkActorStateStore for Store {
 impl ChannelActorStateStore for Store {
     fn get_channel_actor_state(&self, id: &Hash256) -> Option<ChannelActorState> {
         let key = [&[CHANNEL_ACTOR_STATE_PREFIX], id.as_ref()].concat();
-        self.get(key)
-            .map(|v| deserialize_from(v.as_ref(), "ChannelActorState"))
+        let state: ChannelActorState = self
```
quake: Is here any difference between this PR and original code?

jjyr: No difference, I added some logging and forgot to revert the code here.

updated

quake (approved):

gpBlockchain:
lgtm

PR #879, opened 2025-09-03
graph_channels read from memory instead of from DB

### LINKED ISSUE #534: [Optimization]consider adding an API to query the channel synchronization progress.
When there are many channels, sending a-b-c paths requires waiting for the channel synchronization to be completed. At this time, it will be stuck for a long time. You may want to consider adding an API to query the channel synchronization progress.

Image

jjyr:
1. Using indexmap struct to store graph channels,  thus, we can support paginate in graph_channels.
2. Remove Db access in graph_channels and graph_nodes Rpc
3. Add total_count to result

~~Drawback:~~

~~Since our paginate based on return count, the query result may contains duplicates or missing for multiple pages. It is acceptable since graph_channels is mainly used for statistics, one way to workaround this problems is to set limit large enough to return all channels in once.~~

Also fixes #534

--------------------------

Updated the new commit use BTreeMap to store graph channels and nodes avoid the paginate drawback

### DIFF
--- a/crates/fiber-lib/src/rpc/graph.rs
+++ b/crates/fiber-lib/src/rpc/graph.rs
@@ -18,6 +18,7 @@ use ckb_types::H256;
 use jsonrpsee::proc_macros::rpc;
 use jsonrpsee::types::ErrorObjectOwned;
 
+use molecule::prelude::Entity;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use std::sync::Arc;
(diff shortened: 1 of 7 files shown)


crates/fiber-lib/src/rpc/graph.rs
```diff
@@ -357,20 +357,17 @@ where
...
+            u64::from_le_bytes(buf)
+        });
 
-        let channels = network_graph.get_channels_with_params(limit, cursor);
-        let last_cursor = channels
-            .last()
-            .map(|node| JsonBytes::from_vec(node.cursor().to_bytes().into()))
-            .unwrap_or_default();
+        let channels = network_graph.get_channels_with_params(limit, after);
+        let last_cursor = JsonBytes::from_vec(
```
chenyukang, line 379: prefer to use `page` and `per_page` as parameters.

chenyukang:
and to resolve https://github.com/nervosnetwork/fiber/issues/534,
please add a `total_count` field for `nodes` and `channels` graph rpc.

chenyukang:
please do similar refactoring for `graph_nodes` rpc.

crates/fiber-lib/src/rpc/graph.rs
jjyr: The param is still exposed as last_cursor for the external RPC, so even we change to `network_graph.get_channels_with_params(per_page, page * per_page);`  the extern user feels no difference, and some unneccesary complexity is introduced: need to encode two (page, per_page) instead of one param (after) into `last_cursor`.

jjyr:
> and to resolve #534, please add a `total_count` field for `nodes` and `channels` graph rpc.

The tricky part is we need to track private_channels_count or public_channels_count, since graph_channels interface only return public channels, updated in the commits

chenyukang (approved):

PR #899, opened 2025-09-16
Fix peer reconnect during CheckPeerInit

jjyr:
If a peer reconnect during CheckPeerInit check, the first issued CheckPeerInit command may disconnect the Peer. We add SessionId as argument to avoid this situation.

Example:

1. A <-> B
2. A set a timer of CheckPeerInit after 20s
3. 18s later B node restart and reconnect with A
4. A set a second CheckPeerInit after 20s
5. 2s later the first CheckPeerInit is triggerred. A disconnect B

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -262,7 +262,7 @@ pub enum NetworkActorCommand {
     // Settle MPP tlc set
     SettleMPPTlcSet(Hash256),
     // Check peer send us Init message in an expected time, otherwise disconnect with the peer.
-    CheckPeerInit(PeerId),
+    CheckPeerInit(PeerId, SessionId),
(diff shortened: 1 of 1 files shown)


chenyukang:
If this `18s later B node restart and reconnect with A` happened, there is no related `Disconnected` event happenend for node A?

I think this will return `None` for this scenario?
https://github.com/chenyukang/fiber/blob/be6750c6e7f5a359cf129f8361401459a7f45908/crates/fiber-lib/src/fiber/network.rs#L1433

jjyr:
> https://github.com/chenyukang/fiber/blob/be6750c6e7f5a359cf129f8361401459a7f45908/crates/fiber-lib/src/fiber/network.rs#L1433

B restart, it will disconnect from A, and then immediately connect to A. so B inserted in `peer_session_map` again.

quake (approved):

PR #901, opened 2025-09-17
feat: cch separate service

doitian:
Based on #615 and #898

A rework of #522

This PR allows running cch as a separate service, which talks to another fiber node via HTTP and WebSocket JSONRPC.


doitian:
Rework in #1165

PR #904, opened 2025-09-17
Perf: use BinaryHeap for retryable tasks

chenyukang:
use `BinaryHeap` to avoid too many ractor messages.


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1785,124 +1780,153 @@ where
...
-                    Ok(_) | Err(ProcessingChannelError::RepeatedProcessing(_)) => false,
-                    Err(ProcessingChannelError::WaitingTlcAck) => {
-                        waiting_ack = true;
-                        true
+        let mut tasks_to_retry = Vec::new();
+
+        // Process all tasks that are ready to execute
+        while let Some(_task) = state.retryable_tlc_operations.peek() {
+            // here we does not check the task next_retry_time with current time,
+            // because RelayRemoveTlc need a higher priority to be processed
```
doitian, line 1805: Why not set a lower interval for RelayRemoveTlc tasks?

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1785,124 +1780,153 @@ where
...
-                    .await
-                {
-                    Ok(_) | Err(ProcessingChannelError::RepeatedProcessing(_)) => false,
-                    Err(ProcessingChannelError::WaitingTlcAck) => {
-                        waiting_ack = true;
-                        true
+        let mut tasks_to_retry = Vec::new();
+
+        // Process all tasks that are ready to execute
+        while let Some(_task) = state.retryable_tlc_operations.peek() {
```
jjyr:
```suggestion
        while let Some(task) = state.retryable_tlc_operations.pop() {
```

doitian: The `peek` is for the commented-out code block.

crates/fiber-lib/src/fiber/channel.rs
chenyukang: I had a try to set RelayRemoveTlc to half of others interval , but seems not work.

PR #921, opened 2025-10-09
design: single-funded channel

doitian:
> [!TIP]
> [[PREVIEW]](https://github.com/doitian/fiber/blob/single-funded-channel-design/docs/specs/single-funded-channel.md)
> A copy is saved in [Discussion](https://github.com/nervosnetwork/fiber/discussions/1311)

## Actions

- [ ] Add a quick close case for the initiator when the non-initiator has no assets in the channel
- [ ] Save the non-initiator balance in lease lock script args for CKB channels

### DIFF
--- a/docs/specs/single-funded-channel.md
+++ b/docs/specs/single-funded-channel.md
@@ -1,0 +1,111 @@
+# Single-Funded Channel
+
+This feature allows the party who sends the `OpenChannelV2` message (the **initiator**) to provide all the required assets for opening a channel. The counterparty (the **non-initiator**) does not need to contribute funds during channel creation, but must refund the negotiated amount of assets to generate their settlement output cell when the channel is closed.
+
+## Feature Negotiation
+
+- Nodes that support single-funded channels MAY advertise the optional feature bit `ACCEPT_SINGLE_FUNDED_CHANNEL(5)`.
+- Nodes that require peers to accept single-funded channels MUST advertise the mandatory feature bit `ACCEPT_SINGLE_FUNDED_CHANNEL(4)`.
+
+The initiator must not send `OpenChannelV2` to peers that have not advertised either feature bit. The initiator may use OpenChannelV2 with peers that have advertised bit 4 or 5 to open either a single-funded or a dual-funded channel.
+
+## `OpenChannelV2` Schema
+
+```
+table OpenChannelV2 {
+    v1: OpenChannel,
+    refund_reserved_ckb_amount: Uint64,
+}
+
+union FiberMessage {
+    ...
+    AnnouncementSignatures,
+    // Append OpenChannelV2 after AnnouncementSignatures
+    OpenChannelV2,
+}
+```
+
+- `v1` embeds all fields from `OpenChannel`.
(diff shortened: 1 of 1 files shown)


docs/specs/single-funded-channel.md
```diff
@@ -0,0 +1,115 @@
...
+    OpenChannelV2,
+}
+```
+
+- `v1` embeds all fields from `OpenChannel`.
+- `refund_reserved_ckb_amount` specifies how many shannons (or SUDT/XUDT units) the non-initiator must refund at channel closure:
+    - Set to zero for dual-funded channels.
+    - Set to non-zero to open single-funded CKB or SUDT/XUDT channels.
+
+The initiator's nested field `v1.reserved_ckb_amount` must cover both the settlement outputs and fees. Given a commitment fee rate $F$, and maximum transaction size $N$, the initiator must reserve a fee of $3FN$.
```
chenyukang, line 33: please explain how this comes out.

doitian: From the code

docs/specs/single-funded-channel.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Non-Initiator Validation
+
+Upon receiving `OpenChannelV2`, the non-initiator must reject the request if:
+
+1. It has not enabled feature bit 4 or 5.
+2. The ratio `refund_reserved_ckb_amount / minimum_reserved_ckb_amount` falls outside acceptable bounds. Peers configure acceptable ratios in their config files.
```
jjyr, line 40: Please explain the popurse of this validation rule.

docs/specs/single-funded-channel.md
```diff
@@ -0,0 +1,115 @@
...
+1. The `reserved_ckb_amount` is not zero.
+
+## Settlement Transaction Construction
+
+Both dual- and single-funded channels use a similar settlement structure. Each party’s output cell contains:
+
+- **Balance** $B_1$ for the initiator, $B_2$ for the non-initiator.
+- **Reserved amount** $R_1$ for the initiator, $R_2$ for the non-initiator (excluding transaction fees)
+
+For single-funded channels, $R_2 = 0$. Let:
```
jjyr, line 62: R2 is never used in the following document.

Can R2 be a non-zero value in dual fund v2 channel?

docs/specs/single-funded-channel.md
```diff
@@ -0,0 +1,115 @@
...
+
+- $A$ = `refund_reserved_ckb_amount`
+- $C$ = `minimum_reserved_ckb_amount`
+
+### Case: $B_2 - A \ge 0$
```
jjyr, line 67: We have two constructions, and we need to check `B2 - A >= 0` to ensure which construction to use. It may cause confusion because the output's balance is not correspond to channel balance when `B2 -A >= 0` (but the balances are equals when `B2 - A < 0`).

However, we can remove the confusion it by only check `A == 0` to determine the construction.

1. If `A == 0`, both sides use shutdown script.
2. If `A != 0`, non-initiator always use lease script.

doitian:
Deprecated because the commitment lock script has updated: https://github.com/nervosnetwork/fiber-scripts/pull/22

PR #924, opened 2025-10-10
Support profiling

jjyr:
Add a new RPC dev_pprof to do profiling and output flamegraph.
To enable this Rpc make sure the binary is compiled with flamegraph feature and add rpc module prof in the config file.

Note: The dev rpc module is required debug build. But profiling suppose work under release build, to keep consistency we use a standalone RPC module for profiling.

Replace #923 

```
jjy@jdevmac:~/Workspace/fiber(profiling-rpc)$ curl -v -s \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"pprof","params": [{"duration_secs":15}]}' \
  http://127.0.0.1:21714
{"jsonrpc":"2.0","id":1,"result":{"path":"profiles/flamegraph-1760081967.svg"}}
```

```
CARGO_PROFILE_RELEASE_DEBUG=true cargo build --features pprof --release
```

### DIFF
--- a/crates/fiber-lib/src/rpc/mod.rs
+++ b/crates/fiber-lib/src/rpc/mod.rs
@@ -14,6 +14,8 @@ pub mod invoice;
 mod middleware;
 pub mod payment;
 pub mod peer;
+#[cfg(all(feature = "pprof", not(target_arch = "wasm32")))]
+pub mod prof;
 pub mod utils;
 pub mod watchtower;
 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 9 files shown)


crates/fiber-lib/src/rpc/mod.rs
```diff
@@ -308,6 +312,11 @@ pub mod server {
                     )
                     .unwrap();
             }
+
+            #[cfg(not(target_arch = "wasm32"))]
```
chenyukang: need to add cfg check here?

```rust
   #[cfg(feature = "pprof")]
```

crates/fiber-lib/src/rpc/mod.rs
```diff
@@ -14,6 +14,8 @@ pub mod invoice;
 mod middleware;
 pub mod payment;
 pub mod peer;
+#[cfg(not(target_arch = "wasm32"))]
```
chenyukang: #[cfg(all(feature = "pprof", not(target_arch = "wasm32")))]

chenyukang (approved):

PR #930, opened 2025-10-21
feat: record and notify store changes

doitian:
A rework of #736 based on the review https://github.com/nervosnetwork/fiber/pull/736#discussion_r2389703273

### DIFF
--- a/crates/fiber-lib/src/store/store_impl/mod.rs
+++ b/crates/fiber-lib/src/store/store_impl/mod.rs
@@ -1,7 +1,7 @@
 #[cfg(not(target_arch = "wasm32"))]
 mod native;
 #[cfg(not(target_arch = "wasm32"))]
-pub use native::{Batch, DbDirection, IteratorMode, Store};
+pub use native::{Batch, DbDirection, IteratorMode, Store, StoreChangeWatcher};
 
 #[cfg(all(target_arch = "wasm32", not(test)))]
 mod browser;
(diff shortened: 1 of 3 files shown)


chenyukang (approved):

crates/fiber-lib/src/store/store_impl/native.rs
```diff
@@ -100,9 +114,73 @@ impl Store {
     }
 }
 
+pub struct BatchWatcher {
+    inner: Arc<dyn StoreChangeWatcher>,
+    pending_changes: Vec<StoreChange>,
```
chenyukang, line 119: when remove changes from `pending_changes`.

doitian: Batch is destroyed after calling `commit`.

crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -238,6 +238,33 @@ pub enum KeyValue {
     HoldTlc((Hash256, Hash256, u64), u64),
 }
 
+/// Recorded store changes.
```
doitian, line 241: In long term, we could replace KeyValue with StoreChange. But I did not do this because of two reasons:

- We only need a small portion of store change events to implement multi-hop cch payments
- The KeyValue enum has many nested structs that have not implemented serd traits and serd is required for subscription over RPC.

PR #932, opened 2025-10-21
Avoid duplicated retryable tasks

chenyukang:
RemoveTlc with fullfill maybe duplicated.

It's may happen when RemoveCommand also invoked from cron tasks such as: 
https://github.com/chenyukang/fiber/blob/c963f5d98dd55901f58bbc4a8607db9b3b4b03d6/crates/fiber-lib/src/fiber/network.rs#L1544-L1565

### DIFF
--- a/crates/fiber-lib/src/fiber/tests/channel.rs
+++ b/crates/fiber-lib/src/fiber/tests/channel.rs
@@ -6062,3 +6062,50 @@ async fn test_funding_timeout() {
         .expect_event(|event| matches!(event, NetworkServiceEvent::ChannelFundingAborted(_)))
         .await;
 }
+
+#[tokio::test]
+async fn test_channel_one_peer_check_active_fail() {
+    init_tracing();
+
+    let (nodes, channels) =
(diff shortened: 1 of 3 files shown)


crates/fiber-lib/src/fiber/tests/channel.rs
```diff
@@ -6062,3 +6062,50 @@ async fn test_funding_timeout() {
         .expect_event(|event| matches!(event, NetworkServiceEvent::ChannelFundingAborted(_)))
         .await;
 }
+
+#[tokio::test]
+async fn test_channel_one_peer_check_active_fail() {
```
jjyr, line 6067: This test seems unrelated this change

chenyukang: yes, this test is testing another thing, 
the confirmed when one peer is shutdown channel forcely, another peer is still in ChannelReady.

chenyukang: the duplicated tasks issue can be triggered in several unit tests, so I added a debug assertion for it.
https://github.com/nervosnetwork/fiber/pull/932/commits/38f5bc0938b132fb7a81bd35cd0c07d2bc829725

jjyr (approved):

PR #933, opened 2025-10-22
refactor: update contract related code

quake:


crates/fiber-lib/src/fiber/tests/channel.rs
```diff
@@ -1706,7 +1708,7 @@ async fn test_network_send_payment_dry_run_will_not_create_payment_session() {
     init_tracing();
 
     let node_a_funding_amount = 100000000000;
-    let node_b_funding_amount = 6200000000;
+    let node_b_funding_amount = 11800000000;
```
chenyukang, line 1711: it's better to add a constant for minimal funding amount.

.github/workflows/e2e.yml
```diff
@@ -13,12 +13,12 @@ jobs:
           - open-use-close-a-channel
           - udt
           - reestablish
-          - cross-chain-hub
+          # - cross-chain-hub
```
chenyukang, line 16: why this is commented.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -5962,6 +5897,8 @@ impl ChannelActorState {
             attempt_id: None,
             expiry: message.expiry,
             hash_algorithm: message.hash_algorithm,
+            // only in unit test will it be true
+            is_last: message.onion_packet.is_none(),
```
chenyukang, line 5905: I think it's better to be default as `false`?
`apply_add_tlc_operation_with_peeled_onion_packet` will update it to true if proper onion packet contains next.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -6413,10 +6282,16 @@ impl ChannelActorState {
...
         }
 
-        let (commitment_tx, settlement_data) = self.verify_and_complete_tx(
-            commitment_signed.funding_tx_partial_signature,
-            commitment_signed.commitment_tx_partial_signature,
-        )?;
+        let (commitment_tx, settlement_data) =
+            self.verify_and_complete_tx(commitment_signed.funding_tx_partial_signature)?;
+
+        let tlcs = self.get_active_tlcs_for_settlement(false);
```
chenyukang: should only for debugging?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1582,15 +1595,31 @@ where
...
                             let expired_tlcs = actor_state
                                 .tlc_state
                                 .received_tlcs
                                 .get_committed_tlcs()
                                 .into_iter()
-                                .filter(|tlc| tlc.expiry < now)
+                                .filter(|tlc| {
+                                    tlc.is_last
+                                        && tlc.expiry < expect_expiry
+                                        && tlc.removed_confirmed_at.is_none()
```
chenyukang: it's better to use `tlc.remove_reason.is_none()`?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1582,15 +1595,31 @@ where
...
                             }
 
+                            let delay_epoch = EpochNumberWithFraction::from_full_value(
+                                actor_state.commitment_delay_epoch,
+                            );
+                            let epoch_delay_milliseconds = ((delay_epoch.number() as f64
+                                + delay_epoch.index() as f64 / delay_epoch.length() as f64)
+                                * MILLI_SECONDS_PER_EPOCH as f64
+                                * 2.0
+                                / 3.0)
```
chenyukang: I see serveral places using this, maybe a util function is better.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -3931,6 +3965,26 @@ where
...
+                                        self.network
+                                            .send_message(NetworkActorMessage::new_notification(
+                                                NetworkServiceEvent::PreimageCreated(
+                                                    payment_hash,
+                                                    payment_preimage,
+                                                ),
+                                            ))
+                                            .expect(ASSUME_NETWORK_ACTOR_ALIVE);
+                                    }
+                                }
```
chenyukang, line 3974: why do we need this part, I think this part of task will be done if we apply RemoveTlc from waiting queue?

crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -602,6 +604,21 @@ impl PreimageStore for Store {
...
+    fn get_preimage(&self, payment_hash: &Hash256) -> Option<Hash256> {
+        let key = [&[PREIMAGE_PREFIX], payment_hash.as_ref()].concat();
+        self.get(key)
+            .map(|v| deserialize_from(v.as_ref(), "Preimage"))
+            // Try to get the preimage from watchtower store
+            .or_else(|| {
+                let prefix = [&[WATCHTOWER_PREIMAGE_PREFIX], payment_hash.as_ref()].concat();
+                let mut iter = self.prefix_iterator(prefix.as_slice());
+                iter.next()
+                    .map(|(_key, value)| deserialize_from(value.as_ref(), "Preimage"))
```
chenyukang:
```suggestion
                    .map(|(_key, value)| deserialize_from(value.as_ref(), "Watchtower Preimage"))
```
to be different with error messages coming above.

chenyukang (approved):

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -8392,6 +7905,15 @@ impl InMemorySigner {
         }
     }
 
+    /*************  ✨ Windsurf Command ⭐  *************/
```
jjyr: Clean the document

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -7947,13 +7466,29 @@ impl ChannelActorState {
...
             funding_tx_partial_signature,
-            commitment_tx_partial_signature,
+            &compute_tx_message(&commitment_tx),
         )?;
-        self.complete_partially_signed_tx(&tx)
+
+        let completed_commitment_tx = {
+            let sign_ctx = self.get_funding_sign_context();
+            let signature = sign_ctx.sign_and_aggregate(
+                &compute_tx_message(&commitment_tx),
```
jjyr: https://github.com/nervosnetwork/fiber/pull/933/files#diff-0089cc1b51a74d30eb20e939ada0141f62d61c26d2754442900a2c7229aa027bR7474

We can reuse the message

``` rust
                &compute_tx_message(&commitment_tx),
```

.github/workflows/e2e.yml
quake: this e2e test is unstable, would like to fix it in sperate PR

crates/fiber-lib/src/fiber/channel.rs
quake: some unit test are using `AddTlc` message without onion packet directly, change it to false will break these unit tests

crates/fiber-lib/src/fiber/network.rs
quake: I think this condition can be removed since we are checking against committed tlcs only.

crates/fiber-lib/src/fiber/channel.rs
quake: sorry, I forgot to cleanup the debug log.

crates/fiber-lib/src/fiber/network.rs
quake: ChannelActor may be stopped, then retryable_tlc_operations won't insert the preimage to store in time.

PR #934, opened 2025-10-23
Settle payment with invoice in network actor

chenyukang:
Multiple nodes trying to send payment with the same invoice, the expected only one of them succeed.


jjyr (approved):

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1127,12 +1131,6 @@ where
                     );
                     return Err(ProcessingChannelError::FinalIncorrectPreimage);
                 }
-                // update invoice status to received only all the error checking passed
```
doitian, line 1130: If this block is removed, when the no-mpp invoice is updated to Received?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1754,6 +1767,11 @@ where
                             payment_preimage: preimage,
                         }),
                     };
+
+                    self.store
```
doitian: CCH needs to know whether the received tlcs total amount has reached the hold invoice amount. When the invoice is fulfilled, CCH will send payment to get the preimage. When CCH has to make the decision, the hold invoice has no preimage yet.

doitian (approved):

doitian:
Once merged, I will update #950 to rely on the invoice Received state.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1770,21 +1777,33 @@ where
...
                     }
+                } else if !fulfilled {
+                    return Ok(());
                 }
 
+                // if we have enough tlcs to fulfill the invoice, update invoice status to Received
+                // for hold invoice we may don't have preimages yet, so just update status here
+                self.store
+                    .update_invoice_status(&payment_hash, CkbInvoiceStatus::Received)
+                    .expect("update invoice status failed");
```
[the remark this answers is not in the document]
chenyukang: i think it's ok to update to Received even there is a tlc_fail.

quake (approved):

PR #937, opened 2025-10-27
Code refactor on channel actor state

chenyukang:
Use flags is easier to maintain.


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2941,6 +2945,15 @@ impl TlcStatus {
     }
 }
 
+bitflags! {
+    #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+    #[serde(transparent)]
+    pub struct AppliedFlags: u8 {
+        const APPLIED_ADD = 1;
```
jjyr: We can remove the duplicated prefix `APPLIED_`

jjyr (approved):

crates/fiber-lib/src/fiber/channel.rs
chenyukang: good idea.

quake (approved):

PR #945, opened 2025-11-05
feat: hold invoice

doitian:
Allow creating invoice without preimage and settle the invoice with the
preimage later via RPC.


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1140,6 +1140,20 @@ where
...
+                    .expiry_time()
+                    .map(|duration| duration.as_millis() as u64)
+                    .unwrap_or(DEFAULT_HOLD_TLC_TIMEOUT);
+                // Save TLC for the Hold Invoice
+                self.store.insert_payment_hold_tlc(
+                    payment_hash,
+                    HoldTlc {
+                        channel_id: add_tlc.channel_id,
+                        tlc_id: add_tlc.tlc_id.into(),
+                        hold_expire_at: now_timestamp_as_millis_u64() + expiry_time_millis,
```
jjyr: The invoice expire time should based on `invoice.timestamp` instead of current timestamp.
Should we calculate `hold_expire_at` from the `invoice.timestamp`?

doitian: I think so. The TLC should be hold when the invoice is open. You remind me that if we cancel the invoice, we should reject these TLCs as well.

doitian: To-dos:

- [x] Calculate TLC hold expire_at based on invoice timestamp.
- [x] Check that when invoice is canceled, the held TLCs are cleared.

doitian: I made a note on the second item and will check it later.

The first item is done in a new commit.

quake (approved):

PR #948, opened 2025-11-11
refactor(cch): refactor lnd trackers

doitian:
> [!IMPORTANT]
> This is a stacked PR:
> - #948 :point_left:
>     - #950
>         - #971

- Define payment and invoice events to be sent to CCH for both fiber and
  lnd.
- These events can be emitted from fiber store changes in a later PR.
- Extract lnd trackers into its own module.
- Limit concurrent lnd invoice trackers to 5, scheduling them at
  5-minute intervals using a round-robin strategy.

Actions:

- [x] Fix `Failed to spawn test actor: ActorAlreadyRegistered("test_lnd_tracker")`

PR #950, opened 2025-11-13
feat: monitor fiber store changes for cch

doitian:
> [!IMPORTANT]
> This is a stacked PR:
> - #948
>     - #950 :point_left:
>         - #971

- When an invoice state changes in Fiber (settled, failed, expired), the CCH order tracking system automatically detects and synchronizes the state without manual polling or intervention.
- When an outgoing payment state changes in Fiber (succeeded, failed, pending), the CCH order tracking system automatically detects and synchronizes the state without requiring manual status checks.
- CCH monitors invoice for the incoming payment. When CCH has received the incoming payment, it will send the outgoing payment to get the preimage.
- CCH monitors outgoing payment to get the preimage and use the obtained preimage the settle the incoming payment.

This PR will deprecate #615, which has been refactored and split into
- #930 
- #945 
- #942 
- #948 
- #950 :point_left: 

Actions:
- [x] Fix e2e test cross-chain-hub

### DIFF
--- a/crates/fiber-lib/src/cch/actor.rs
+++ b/crates/fiber-lib/src/cch/actor.rs
@@ -1,32 +1,33 @@
 use anyhow::{anyhow, Context, Result};
 use futures::StreamExt as _;
 use lightning_invoice::Bolt11Invoice;
-use lnd_grpc_tonic_client::{
-    create_invoices_client, create_router_client, invoicesrpc, lnrpc, routerrpc, InvoicesClient,
-    RouterClient, Uri,
-};
-use ractor::{call, RpcReplyPort};
+use lnd_grpc_tonic_client::{invoicesrpc, lnrpc, routerrpc, Uri};
+use ractor::{call, port::OutputPortSubscriberTrait as _, OutputPort, RpcReplyPort};
 use ractor::{Actor, ActorCell, ActorProcessingErr, ActorRef};
 use secp256k1::{PublicKey, Secp256k1, SecretKey};
 use serde::Deserialize;
 use std::str::FromStr;
+use std::sync::Arc;
 use tentacle::secio::SecioKeyPair;
-use tokio::{select, time::sleep};
+use tokio::select;
 use tokio_util::{sync::CancellationToken, task::TaskTracker};
 
-use crate::cch::order::CchInvoice;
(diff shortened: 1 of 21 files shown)


crates/fiber-lib/src/cch/actor.rs
```diff
@@ -478,15 +430,47 @@ impl CchState {
...
-    async fn receive_btc(
+    async fn handle_lnd_invoice_changed_event(
         &mut self,
-        myself: ActorRef<CchMessage>,
-        receive_btc: ReceiveBTC,
-    ) -> Result<CchOrder, CchError> {
+        mut order: CchOrder,
+        status: CchIncomingPaymentStatus,
+    ) -> Result<()> {
+        if !(order.is_incoming_invoice_lnd() && order.is_awaiting_invoice_event()) {
```
chenyukang: this check seems is used at least 3 times, use a single help function maybe more mantainable.

crates/fiber-lib/src/cch/order.rs
```diff
@@ -100,7 +74,35 @@ pub struct CchOrder {
...
+    pub fn is_incoming_invoice_lnd(&self) -> bool {
+        matches!(self.incoming_invoice, CchInvoice::Lightning(_))
+    }
+    pub fn is_outgoing_payment_fiber(&self) -> bool {
+        self.is_incoming_invoice_lnd()
+    }
+    pub fn is_outgoing_payment_lnd(&self) -> bool {
+        self.is_incoming_invoice_fiber()
+    }
+
```
chenyukang, line 89: the later 2 functions are just alias of the first 2 functions, 
proper function naming to keep only 2 function seems more maintainable.

doitian: The extra 2 functions make code easy to reason. In the place I want to ensure the order is sending outgoing payment to fiber, I can use `is_outgoing_payment_fiber`. It's more straightforward than `is_incoming_invoice_lnd` since I have to reason that when incoming invoice is from lnd, the outgoing payment is to fiber.

doitian:
Convert to Draft. Wait on #934

doitian:
#934  still does not update hold invoice to Received. Wait on #961

chenyukang (approved):

PR #954, opened 2025-11-21
feat: add pending_tlcs field to rpc list_channels

quake:


crates/fiber-lib/src/rpc/channel.rs
```diff
@@ -513,6 +535,23 @@ where
                         remote_balance: state.get_remote_balance(),
                         offered_tlc_balance: state.get_offered_tlc_balance(),
                         received_tlc_balance: state.get_received_tlc_balance(),
+                        pending_tlcs: state
+                            .tlc_state
```
chenyukang, line 539: do we need to return `Received` or `Offered`?

chenyukang (approved):

crates/fiber-lib/src/rpc/channel.rs
quake: the TlcStatus will return Outbound / Inbound

PR #986, opened 2025-12-06
Payment actor

jjyr:
## Intention

Move the send payment logic from network actor to a new payment actor.
The path finding of payment blocking the network actor thread, this change will fix this problem.

## potential issue

Since we move some payment logic into payment actor, we may encounter sychronize / lock issues.

## TODO

- [x] Handle resume payment actor after channel ready
- [x] TPS benchmark regression

A new metrics is added

``` rust
pub const INFLIGHT_PAYMENTS_COUNT: &str = "fiber.inflight_payments_count";
```

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -13,7 +13,6 @@ use ractor::{
 };
 use rand::seq::{IteratorRandom, SliceRandom};
 use rand::Rng;
-use secp256k1::Secp256k1;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
 use std::borrow::Cow;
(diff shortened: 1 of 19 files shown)


chenyukang (approved):

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2996,278 +2759,66 @@ where
...
-            return Err(Error::SendPaymentError(err.to_string()));
-        }
-        Ok(more_attempt)
-    }
-
     fn register_payment_retry(
         &self,
         myself: ActorRef<NetworkActorMessage>,
-        state: &mut NetworkActorState<S>,
+        _state: &mut NetworkActorState<S>,
```
chenyukang: could remove this paramter.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2996,278 +2759,66 @@ where
...
-        })?;
-
-        if !payment_data.dry_run && state.retry_send_payment_count >= MAX_RETRY_SEND_PAYMENTS {
-            return Err(Error::InvalidParameter(
-                "Too many pending retrying payment requests".to_string(),
-            ));
-        }
-        self.send_payment_with_payment_data(myself, state, payment_data)
-            .await
+        let _ = myself.send_message(NetworkActorMessage::new_event(
```
chenyukang: add debug message for error?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1978,20 +1895,29 @@ where
...
-                        myself,
-                        state,
-                        command.payment_hash,
-                        command.attempt_id,
-                        Some((ProcessingChannelError::TlcForwardingError(err.clone()), err)),
-                        command.previous_tlc,
-                    )
-                    .await;
+                    Ok(()) => {
+                        let _ = reply.send(Ok(()));
```
chenyukang, line 1956: the old code seems do not expect get reply from this point.

is there any code really waiting for this reply?
i remember we get a deadlock when some channel actor is waiting for the result from network actor, and network actor is waiting some result from channel actor.

jjyr: We write to DB in the failure path: `on_add_tlc_result_event` function. if we remove the reply, the DB status can not been guranteed for payment actor. (I try to keep changes as small as possible in this PR)

Currently, the message call is one way: `payment actor -> network actor` so we are ok for now.

PR #999, opened 2025-12-10
Fix the performance issue in path finding

chenyukang:
This PR try to reduce the call numbers of `find_path`.

- Add `payment_find_path_stats` to track the number of `inner_find_path`.
- `inner_find_path` now support to find the max single path for a target, so that we can avoid to do meanless searching when the result determined to be fail.
- When we get a router with `mid` amount, try to extend it to large amount with cheaper checking, to reduce the number of invocation of `inner_find_path` with exploring all the  whole graph.


jjyr (approved):

crates/fiber-lib/src/fiber/payment.rs
```diff
@@ -1323,19 +1332,29 @@ where
                             );
                         }
                     }
-                    remain_amount -= session_route.receiver_amount();
-                    target_amount = remain_amount;
+                    let current_amount = session_route.receiver_amount();
+                    remain_amount -= current_amount;
+                    target_amount = if let Some(single) = single_path_max {
```
jjyr, line 1449: What if we have two paths that can both pay `single_path_max` value like `A -> B -> C`, `A -> D -> C`.

After first iter, we use a smallter `target_amount`, but actually we can still pay `single_path_max`.

crates/fiber-lib/src/fiber/payment.rs
```diff
@@ -1323,19 +1332,29 @@ where
...
+                    } else {
+                        remain_amount
+                    };
                     if let Some(fee) = max_fee {
                         max_fee = Some(fee - session_route.fee());
                     }
                     result.push(attempt);
+                    if remain_amount > 0
+                        && remain_amount
+                            > current_amount * (session.max_parts() - result.len()) as u128
```
jjyr, line 1460: The `current_amount` is sum of multi-paths, shouldn't use it to estimate a single path amount.

chenyukang: `current_amount` is the current amount of the current single attempt route's amount?

```rust
                    let current_amount = session_route.receiver_amount();
```
it's expected as the max current amount router we can get from the network graph, so we expect the later amount will smaller than this value:

```rust
remain_amount > current_amount * (session.max_parts() - result.len()) as u128
``` 
means we can not find enough amount for `remain_amount`.

crates/fiber-lib/src/fiber/payment.rs
chenyukang: if we get a smaller `target_amount`, means we only need remain amount for the payment. 
this check is avoiding the scenario like this:

The payment have `remain_amount` of 10000, but actually the largest single path router we have is 90, so we should set up bound of search amount to 90, so we can avoid meanless search in binary search.

PR #1002, opened 2025-12-11
chore: remove unused function in store

chenyukang:
`get_range` is not used anymore.


doitian:
I noticed a warning that the delete function isn’t used. Perhaps it should be removed too.

chenyukang:
> I noticed a warning that the delete function isn’t used. Perhaps it should be removed too.

which function.

doitian:
> > I noticed a warning that the delete function isn’t used. Perhaps it should be removed too.
> 
> which function.

`delete`

chenyukang:
> > > I noticed a warning that the delete function isn’t used. Perhaps it should be removed too.
> > 
> > 
> > which function.
> 
> `delete`

it's used in `remove_watch_channel`.

jjyr (approved):

PR #1003, opened 2025-12-12
fix: handle funding transaction signing failure

doitian:
## Summary

This PR fixes a panic in the `SignFundingTx` handler when funding transaction signing fails. Instead of panicking, the code now properly handles the error by sending a `TxAbort` message to the peer and aborting the funding process.

### 1. Error Handling in SignFundingTx Handler
- **Before**:
    - Used `.expect("Signing succeeded")` which would panic if signing failed
    - Used `?` which would exit network actor if signing failed
- **After**: Properly handles signing errors with a `match` statement that:
  - Logs the error
  - Sends a `TxAbort` message to the peer with error details
  - Aborts funding and closes the channel gracefully
  - Returns early from the function

### 2. Code Refactoring
- Removed significant code duplication between the two branches (`Some(partial_witnesses)` and `None`)
- Consolidated error handling into a single location
- Simplified the code flow for better maintainability

### 3. Test Coverage
- Added `SignFundingTxFailureMockMiddleware` to simulate signing failures in tests
- Added `test_abort_funding_on_sign_funding_tx_failure` test case that:
  - Verifies that when signing fails, both nodes receive `ChannelFundingAborted` events
  - Confirms that `TxAbort` is sent and funding is aborted properly

## Impact

- **Breaking Changes**: None
- **Behavior Change**: Signing failures now result in graceful error handling instead of panics
- **Performance**: No performance impact
- **Security**: Improves system stability by preventing panics during channel funding

### DIFF
--- a/crates/fiber-lib/src/fiber/tests/network.rs
+++ b/crates/fiber-lib/src/fiber/tests/network.rs
@@ -1072,7 +1072,7 @@ impl MockChainActorMiddleware for CkbTxFailureMockMiddleware {
 
 #[derive(Clone, Debug)]
 struct SignFundingTxFailureMockMiddleware;
-#[cfg_attr(target_arch="wasm32",async_trait::async_trait(?Send))]
+#[cfg_attr(target_arch="wasm32", async_trait::async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl MockChainActorMiddleware for SignFundingTxFailureMockMiddleware {
     async fn handle(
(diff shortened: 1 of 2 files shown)


crates/fiber-lib/src/fiber/tests/network.rs
```diff
@@ -1069,6 +1070,37 @@ impl MockChainActorMiddleware for CkbTxFailureMockMiddleware {
     }
 }
 
+#[derive(Clone, Debug)]
+struct SignFundingTxFailureMockMiddleware;
+#[cfg_attr(target_arch="wasm32",async_trait::async_trait(?Send))]
```
chenyukang:
```suggestion
#[cfg_attr(target_arch="wasm32", async_trait::async_trait(?Send))]
```

chenyukang (approved):

PR #1006, opened 2025-12-15
fix: watchtower should mark settled tlc

quake:


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -7653,6 +7653,8 @@ pub trait ChannelActorStateStore {
     fn remove_payment_hold_tlc(&self, payment_hash: &Hash256, channel_id: &Hash256, tlc_id: u64);
     fn get_payment_hold_tlcs(&self, payment_hash: Hash256) -> Vec<HoldTlc>;
     fn get_node_hold_tlcs(&self) -> HashMap<Hash256, Vec<HoldTlc>>;
+    /// Check if a tlc is settled on chain
+    fn is_tlc_settled(&self, channel_id: &Hash256, payment_hash: &[u8]) -> bool;
```
chenyukang: all other api is using `payment_hash: &Hash256`, should we keep consistency here?

chenyukang (approved):

PR #1010, opened 2025-12-16
feat: save forwarding channel info

quake:
We need to track next hop tlc info to support watchtower monitor TLC.


crates/fiber-lib/src/rpc/channel.rs
```diff
@@ -293,6 +293,12 @@ pub struct Htlc {
     /// The expiry of the htlc
     #[serde_as(as = "U64Hex")]
     pub expiry: u64,
```
chenyukang, line 295: I think we do not need to add parameters for these, seems there is no testing scenario for it?

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -3019,14 +3023,7 @@ pub struct TlcInfo {
     pub created_at: CommitmentNumbers,
     pub removed_reason: Option<RemoveTlcReason>,
 
-    /// Note: `previous_tlc` is used to track the tlc chain for a multi-tlc payment,
-    ///       we need to know previous when removing tlc backwardly.
-    ///
-    /// Node A ---------> Node B ------------> Node C ----------> Node D
-    ///  tlc_1 <---> (tlc_1) (tlc_2) <---> (tlc_2) (tlc_3) <----> tlc_3
-    ///                ^^^^                 ^^^^
-    ///
```
chenyukang, line 3028: seems it's better to keep some comments for it.
```
 /// Node A ----ch1-----> Node B ------ch2-------> Node C ----ch3------> Node D
 ///  tlc_1 <-------> (tlc_1)(tlc_2) <-------> (tlc_2)(tlc_3) <----> tlc_3
 ///                   ^^^^   ^^^^              ^^^^    ^^^^
 ///                  
 ///  `ch1` use this field to track `tlc_2`, `ch2` use this field to track `tlc_1`
```

chenyukang: `ch1` use this field to track `tlc_2`, `ch2` use this field to track `tlc_1`, is my understanding right?

quake: updated code comment

chenyukang (approved):

PR #1017, opened 2025-12-20
perf: remove channel_id from TlcInfo

quake:
Remove channel_id from struct TlcInfo, slightly reduce storage overhead.


crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1734,7 +1734,7 @@ where
                                 {
                                     if self
                                         .store
-                                        .is_tlc_settled(&tlc.channel_id, &tlc.payment_hash)
+                                        .is_tlc_settled(&actor_state.get_id(), &tlc.payment_hash)
```
chenyukang: why not just use `&channel_id`.

chenyukang (approved):

PR #1022, opened 2025-12-24
refactor: improve hold TLC settlement logic

### LINKED ISSUE #965: [settle_invoice] two nodes can spend the same invoice.
version: 0.6.0

Two nodes send the same invoice at the same time, and then settle_invoice this invoice; both transactions will be successful.

Image

expected:
Only one transaction can be successfully sent

doitian:
Refactor hold TLC settlement logic to improve payment handling:

- Add invoice status validation (Expired, Cancelled, Paid states properly handled)
- Ensure `RemoveTlc` is called for invalid TLCs
- Reject overpaid TLCs for MPP invoices
- For non-MPP invoices, accept only one fulfilled TLC and reject all others

### Key Changes

1. **New `SettleTlcSetCommand` module**: Encapsulates TLC settlement logic with clear verification steps and settlement actions
2. **Separated command handling**: Split `SettleTlcSet` into two distinct commands:
   - `SettleTlcSet(payment_hash, tlc_ids)` - Settle TLCs by given list of `(channel_id, tlc_id)`
   - `SettleHoldTlcSet(payment_hash)` - Settle hold TLC set saved for a payment hash
3. **Improved channel actor**: Removed `is_mpp` field from `PendingNotifySettleTlc` and simplified the settlement flow
4. **Comprehensive test coverage**: Added unit tests for various scenarios including invoice status validation, MPP handling, and edge cases

### Behavior Notes

This refactoring ensures that, in the presence of multiple payments to the same invoice, at most one payment succeeds. However, due to the lack of database transaction support, there remains a small window where the total accepted amount may exceed the invoice amount.

Additionally, in the case of two concurrent MPP payments, it's possible to partially accept TLCs from both without either payment fully succeeding. This is an inherent limitation of MPP and cannot be avoided on the recipient side.

## Related Issues

Closes #965

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -935,15 +935,15 @@ where
                         // Not a hold invoice.
                         if is_mpp {
                             // Use the default expiry for mpp
-                            now_timestamp_as_millis_u64() + DEFAULT_HOLD_TLC_TIMEOUT
+                            Some(now_timestamp_as_millis_u64() + DEFAULT_HOLD_TLC_TIMEOUT)
                         } else {
                             // Use 0 to indicate to not create HoldTLC for single path payment
                             // with preimage.
-                            0
+                            None
                         }
                     } else {
                         // A hold invoice. Ensure the expiry is large enough for manual settlement via RPC.
-                        match invoice.expiry_time() {
+                        Some(match invoice.expiry_time() {
                             Some(invoice_expiry) => u64::try_from(
                                 invoice_expiry
                                     .as_millis()
@@ -952,7 +952,7 @@ where
                             )
                             .unwrap_or(u64::MAX),
                             None => tlc.expiry,
-                        }
+                        })
                     };
                     state
                         .pending_notify_settle_tlcs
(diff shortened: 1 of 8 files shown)


crates/fiber-lib/src/fiber/settle_tlc_set_command.rs
```diff
@@ -0,0 +1,262 @@
+use crate::{
```
chenyukang, line 1: tile file name is not following our current naming style, suggest something like `settle_tlcs`.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -2693,19 +2692,25 @@ where
...
-            is_mpp,
             hold_expire_at,
         } in pending_notify_tlcs
         {
             let channel_id = state.get_id();
             // Hold the tlc
             let expiry_duration =
                 Duration::from_millis(hold_expire_at.saturating_sub(now_timestamp_as_millis_u64()));
-            if !expiry_duration.is_zero() {
+            if expiry_duration.is_zero() {
```
chenyukang: it's better to use `hold_expire_at.is_zero()`, then 
```rust
 let expiry_duration =
          Duration::from_millis(hold_expire_at.saturating_sub(now_timestamp_as_millis_u64()));
self.network.send_after(expiry_duration, timeout_command);
```

in latter code.

chenyukang: I think even we add a new field for `is_hold` in `PendingNotifySettleTlc` seems more readable.

chenyukang (approved):

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -3125,6 +3135,21 @@ impl From<TlcInfo> for TlcNotifyInfo {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TlcInfoWithChannelId {
```
quake: We probably don’t need the `TlcInfoWithChannelId` structure. I’ve looked through the related code before, and from a logical perspective, only the `shared_secret` and `channel_id` are actually required to construct the error message.

doitian: Do you mean using a struct like `TlcErrorReplyContext`?

quake: > Do you mean using a struct like `TlcErrorReplyContext`?

yes

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -926,15 +926,15 @@ where
                         // Not a hold invoice.
                         if is_mpp {
                             // Use the default expiry for mpp
-                            now_timestamp_as_millis_u64() + DEFAULT_HOLD_TLC_TIMEOUT
+                            Some(now_timestamp_as_millis_u64() + DEFAULT_HOLD_TLC_TIMEOUT)
                         } else {
                             // Use 0 to indicate to not create HoldTLC for single path payment
```
quake: the comment should be updated also since we changed 0 to None.

doitian:
Going to change the logic a bit according to the feedback from @gpBlockchain 

Once we have added TLCs into the pending list, we will not reject them because of invoice status. An expired invoice will only reject new incoming TLCs.

PR #1025, opened 2025-12-26
Fix payment state on add or remove tlc

jjyr:
This PR move the `on_add_tlc_result_event` and `on_remove_tlc_result_event` code from **NetworkActor** to **PaymentActor**. The NetworkActor still handling the `SendPaymentOnionPacket` command, but after that, NetworkActor send a message to PaymentActor.

After this PR:

1. all the payment session updateing are maintained in PaymentActor.
2. resolve a payment status bug(we have fixed the bug in NetworkActor, but forget to fix it in PaymentActor, now we removed the duplicate function due to refactoring).

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -12,7 +12,6 @@ use ractor::{
     call_t, Actor, ActorCell, ActorProcessingErr, ActorRef, RpcReplyPort, SupervisionEvent,
 };
 use rand::seq::{IteratorRandom, SliceRandom};
-use rand::Rng;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
 use std::borrow::Cow;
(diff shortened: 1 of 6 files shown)


doitian (approved):

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -708,7 +708,9 @@ macro_rules! debug_event {
         #[cfg(any(debug_assertions, feature = "bench"))]
         $network
             .send_message(NetworkActorMessage::new_notification(
-                NetworkServiceEvent::DebugEvent(DebugEvent::Common($debug_event.to_string())),
+                $crate::fiber::network::NetworkServiceEvent::DebugEvent(
```
chenyukang, line 461: why change this.

jjyr: This macro is used in payment module, so we need to change it.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2876,131 +2751,75 @@ where
...
-
-    fn notify_payment_final(&self, state: &mut NetworkActorState<S>, payment_hash: Hash256) {
-        if let Some(actor) = state.inflight_payments.get(&payment_hash) {
-            if let Err(err) = actor.send_message(PaymentActorMessage::NotifyPaymentFinal) {
-                debug!(
-                    "CheckPayment message dropped because payment actor is likely stopping, error: {err}"
-                );
-            }
-        }
+        self.ensure_payment_command(
```
chenyukang: seems `ensure_payment_command` is not a proper function name, i can not get it's intention from the name.

jjyr: How about resume_payment_actor_and_send_command ?

crates/fiber-lib/src/fiber/network.rs
chenyukang: got it.

crates/fiber-lib/src/fiber/network.rs
chenyukang: yes, seems a better one.

chenyukang (approved):

chenyukang:
can we also move `SendPaymentData` to `payment.rs`?

PR #1081, opened 2026-01-28
Improve find_path performance with map

chenyukang:
Improved path-finding performance in routing.
Added a dedicated CI job to run payment benchmarks and report regressions.


quake:
I can confirm that during benchmarking, the main performance hotspot appears inside get_node_inbounds(). The reason is that `Pubkey` equality checks are expensive: internally, `PartialEq` for the `Pubkey` type calls secp256k1_ec_pubkey_serialize() in order to compare keys. This means every equality comparison performs a full serialization of the public key.

Because get_node_inbounds() iterates over all channels and performs up to two Pubkey equality checks per channel, a large number of costly serializations are triggered. As a result, the performance of the path-finding benchmark becomes dominated by cryptographic serialization work rather than the actual graph logic.

image

quake:
A more fundamental change may be to stop storing `Pubkey` objects inside `ChannelInfo` and instead store already-serialized compressed public keys (e.g. [u8; 33]). This removes the need for repeated serialization during equality checks, hashing, and lookups.

chenyukang:
> A more fundamental change may be to stop storing `Pubkey` objects inside `ChannelInfo` and instead store already-serialized compressed public keys (e.g. [u8; 33]). This removes the need for repeated serialization during equality checks, hashing, and lookups.

I will open another PR for this.

crates/fiber-lib/src/fiber/graph.rs
```diff
@@ -626,10 +630,28 @@ where
...
             }
             OwnedChannelUpdateEvent::Down(channel_outpoint) => {
+                // Remove from node_channels index
+                if let Some(channel) = self.channels.get(&channel_outpoint) {
+                    if let Some(set) = self.node_channels.get_mut(&channel.node1()) {
+                        set.remove(&channel_outpoint);
+                    }
+                    if let Some(set) = self.node_channels.get_mut(&channel.node2()) {
+                        set.remove(&channel_outpoint);
+                    }
```
quake, line 659: we need to remove entry if the set is empy.

crates/fiber-lib/src/fiber/graph.rs
quake: no need to collect, please use iterator directly:
```rust
        self.node_channels
            .get(&node_id)
            .map(|outpoints| {
                outpoints.iter().filter_map(move |outpoint| {
                    self.channels.get(outpoint).and_then(|channel| {
                        match channel.update_of_node1.as_ref() {
                            Some(info) if node_id == channel.node1() && info.enabled => {
                                return Some((channel.node2(), channel, info));
                            }
                            _ => {}
                        }
                        match channel.update_of_node2.as_ref() {
                            Some(info) if node_id == channel.node2() && info.enabled => {
                                return Some((channel.node1(), channel, info));
                            }
                            _ => {}
                        }
                        None
                    })
                })
            })
            .into_iter()
            .flatten()
```

quake, line 1070:
```rust
let mut channels: Vec<_> = self
            .node_channels
            .get(&node_id)
            .map(|outpoints| {
                outpoints.iter().filter_map(move |outpoint| {
                    self.channels.get(outpoint).and_then(|channel| {
                        match channel.update_of_node1.as_ref() {
                            Some(info) if node_id == channel.node2() && info.enabled => {
                                return Some((channel.node1(), channel.node2(), channel, info));
                            }
                            _ => {}
                        }
                        match channel.update_of_node2.as_ref() {
                            Some(info) if node_id == channel.node1() && info.enabled => {
                                return Some((channel.node2(), channel.node1(), channel, info));
                            }
                            _ => {}
                        }
                        None
                    })
                })
            })
            .into_iter()
            .flatten()
            .collect();
```

quake (approved):

PR #1083, opened 2026-01-29
feat: add external wallet signing for channel funding

quake:
Implement RPC endpoints that allow users to open payment channels by signing funding transactions with their own external wallets, instead of having the node automatically sign with its private key.

New RPC methods:
- open_channel_with_external_funding: negotiates channel with peer and returns unsigned funding tx for user to sign externally
- submit_signed_funding_tx: accepts signed tx and completes channel opening

Key changes:
- Add AwaitingExternalFunding channel state for channels waiting for user signature
- Add ExternalFundingTxBuilder to build unsigned funding transactions
- Add BuildUnsignedFundingTx message to CKB chain actor
- Add timeout mechanism for external funding (reuses existing funding timeout)
- Update RPC documentation with new endpoints

### DIFF
--- a/crates/fiber-lib/src/rpc/channel.rs
+++ b/crates/fiber-lib/src/rpc/channel.rs
@@ -1,5 +1,4 @@
 use crate::fiber::channel::{TLCId, TlcStatus};
-use crate::fiber::serde_utils::EntityHex;
 use crate::fiber::{
     channel::{
         AwaitingChannelReadyFlags, AwaitingTxSignaturesFlags, ChannelActorStateStore,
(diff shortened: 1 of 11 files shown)


ashuralyk:
Does this PR only allow using external wallet for channel Open phrase? How about the channel Close phrase?

Xcodes-chain:
pending_external_funding_replies memory leak risk:
If the peer rejects the channel (does not send back AcceptChannel), the reply port will remain in pending_external_funding_replies forever.
It is necessary to clean up this map when the channel fails, times out, or is abandoned.

crates/fiber-lib/src/rpc/channel.rs
```diff
@@ -638,4 +793,66 @@ where
...
+        handle_actor_call!(self.actor, message, params).map(|response| {
+            OpenChannelWithExternalFundingResult {
+                temporary_channel_id: response.channel_id,
+                unsigned_funding_tx: response.unsigned_funding_tx.into(),
+            }
+        })
+    }
+
+    /// Submits a signed funding transaction for an externally funded channel.
+    pub async fn submit_signed_funding_tx(
```
Xcodes-chain, line 840: SubmitSignedFundingTx Missing error handling after moving the reply port into the channel command

quake:
> Does this PR only allow using external wallet for channel Open phrase? How about the channel Close phrase?

we don't need external wallet signing for channel close

quake:
closed via #1120

PR #1084, opened 2026-01-29
Use new Pubkey for better performance

chenyukang:
Pubkey now stores the serialized form ([u8; 33]) directly for fast comparison and hashing.

Some benchmark for payment find path:

```console
find_path_single_call_large_graph/find_path_200n_2000c
                        time:   [1.3877 ms 1.4735 ms 1.5178 ms]
                        change: [-62.407% -60.892% -59.065%] (p = 0.00 < 0.05)
                        Performance has improved.
```


chenyukang:
`TPS slightly improved (+3.5%)` from benchmark test in CI.

gpBlockchain:
lgtm

crates/fiber-lib/src/fiber/types.rs
```diff
@@ -326,36 +328,50 @@ impl Privkey {
 }
 
 /// The public key for a Node
-#[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct Pubkey(pub PublicKey);
+/// It stores the serialized form ([u8; 33]) directly for fast comparison and hashing
+/// and uses SliceHex for human-readable formats (JSON), Pubkey33Bytes for bincode
+#[serde_as]
+#[derive(Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct Pubkey(#[serde_as(as = "IfIsHumanReadable<SliceHex, Pubkey33Bytes>")] pub [u8; 33]);
```
quake: I think  default tuple serialization here should be ok, no need to implement a customized `Pubkey33Bytes` type.
```suggestion
pub struct Pubkey(#[serde_as(as = "IfIsHumanReadable<SliceHex, [u8; 33]>")] pub [u8; 33]);
```

crates/fiber-lib/src/fiber/serde_utils.rs
```diff
@@ -11,13 +15,14 @@ where
...
     String::deserialize(deserializer)
         .and_then(|string| {
-            if string.len() < 2 || &string[..2].to_lowercase() != "0x" {
-                return Err(Error::custom(format!(
-                    "hex string does not start with 0x: {}",
-                    &string
-                )));
+            // Accept both "0x" prefixed and non-prefixed hex strings for backward compatibility
+            // with secp256k1::PublicKey's serde format which doesn't use "0x" prefix
+            let hex_str = if string.len() >= 2 && string[..2].to_lowercase() == "0x" {
```
jjyr: https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix

```
string.strip_prefix("0x")
```

crates/fiber-lib/src/fiber/types.rs
chenyukang: Need to be:
```rust
pub struct Pubkey(#[serde_as(as = "IfIsHumanReadable<SliceHex, [_; 33]>")] pub [u8; 33]);
```

but yeah, it simpler than customized new type.

quake (approved):

PR #1088, opened 2026-01-30
Refactory: Use LocalSigner to manage secretkeys.

joii2020:
This PR centralizes private key handling and signing behind `LocalSigner`, ensuring the `secret_key` is never passed around or accessed directly across the codebase. This significantly reduces the private key exposure surface and the risk of accidental misuse. 

In addition, it establishes a stable signing interface to prepare for the WASM build, where signing will be delegated to third-party wallets (signature-only, no private key input).

### DIFF
--- a/crates/fiber-lib/src/ckb/signer.rs
+++ b/crates/fiber-lib/src/ckb/signer.rs
@@ -1,0 +1,208 @@
+//! Local signer implementation for CKB transactions.
+//!
+//! This module provides a minimal encapsulation of the secret key used for signing
+//! CKB transactions. It caches derived values (public key, pubkey hash) to avoid
+//! repeated computation.
+
+use ckb_sdk::{
+    constants::SIGHASH_TYPE_HASH,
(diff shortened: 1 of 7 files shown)


crates/fiber-lib/src/ckb/signer.rs
```diff
@@ -0,0 +1,208 @@
...
+        rpc_url: String,
+    ) -> Result<FundingTx, FundingError> {
+        // Convert between different versions of secp256k1.
+        // This app requires 0.28 because of:
+        // ```
+        // #[derive(Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash, Serialize, Deserialize, Debug)]
+        // pub struct Signature(pub Secp256k1Signature);
+        // ```
+        //
+        // However, ckb-sdk-rust still uses 0.24.
```
quake: ckb-sdk-rust are use secp256k1 0.30, not 0.24, could you help to confirm it? https://github.com/nervosnetwork/ckb-sdk-rust/blob/master/Cargo.toml#L24C1-L24C10

quake (approved):

PR #1089, opened 2026-02-01
use forward! for fiber outgoing payments

### LINKED ISSUE #1071: Use `forward!` to handle response in a new incomming message when ready
> https://docs.rs/ractor/latest/ractor/macro.forward.html
> 
> 
> 
> We may use `forward!` to replace `call!` here, it won't block the current future. 

 _Originally posted by @jjyr in [#971](https://github.com/nervosnetwork/fiber/pull/971/changes#r2711025251)_

swananan:
fixes #1071

I’m unsure about this change: previously the result wasn’t always forwarded to cch actor (there is a fallback retry mechanism), but now (if I understand correctly) forward! will always relay the mapped message—only a fatal actor error would stop it. I added an oneshot to surface transient errors and a Noop message for cch actor to avoid unintended state changes. Please confirm this is acceptable.

### DIFF
--- a/crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
+++ b/crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
@@ -1,13 +1,13 @@
 use anyhow::{anyhow, Result};
 use futures::StreamExt as _;
 use lnd_grpc_tonic_client::routerrpc;
-use ractor::{call, ActorRef};
+use ractor::{forward, ActorRef};
 
 use crate::{
     cch::{
         actions::{
             backend_dispatchers::{dispatch_payment_handler, PaymentHandlerType},
-            ActionExecutor,
+            ActionExecutor, CchOrderAction,
         },
         actor::CchState,
         trackers::{map_lnd_payment_changed_event, CchTrackingEvent, LndConnectionInfo},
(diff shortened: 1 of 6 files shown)


crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
```diff
@@ -45,46 +52,55 @@ impl ActionExecutor for SendFiberOutgoingPaymentExecutor {
...
-                    return Err(anyhow!(failure_reason));
                 }
             }
-        };
+        })
+        .map_err(|err| anyhow!(err.to_string()))?;
 
-        self.cch_actor_ref
-            .send_message(CchMessage::TrackingEvent(event))?;
+        if let Ok(err) = retry_rx.try_recv() {
```
doitian: It makes no sense to use `forward!` but still wait and block here. To adopt `forward!`, we have to refactor the structure a bit to notify that CCH should retry the action later via message.

doitian (changes requested):

crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
swananan: Got it, forward! doesn’t await, so the oneshot isn’t meaningful here.

doitian (approved):

PR #1091, opened 2026-02-02
chore(deps): upgrade secp256k1 to 0.30 and musig2 to 0.2.4

quake:
This upgrade aligns all secp256k1 dependencies to version 0.30, eliminating the need for hacky hex string conversions between different secp256k1 versions.

Changes:
- secp256k1: 0.28.0 → 0.30.0
- musig2: 0.0.11 → 0.2.4 (now uses secp256k1 0.30)

API fixes for secp256k1 0.30:
- RecoveryId: to_i32() → i32::from(), from_i32() → try_from()
- Schnorr: sign/verify now take &[u8] instead of &Message
- Signature: serialize() → to_byte_array()
- Feature: rand-std → rand

Simplifies sign_funding_tx by removing hex conversion workaround.

### DIFF
--- a/migrate/src/main.rs
+++ b/migrate/src/main.rs
@@ -1,6 +1,5 @@
 use clap::Parser;
-use fiber::store::db_migrate::DbMigrate;
-use fiber::store::Store;
+use fiber_v061::store::{db_migrate::DbMigrate, Store};
 use fnn_migrate::migrations::*;
 use fnn_migrate::util::prompt;
 use std::path::Path;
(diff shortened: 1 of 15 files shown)


migrate/src/main.rs
```diff
@@ -1,6 +1,5 @@
 use clap::Parser;
-use fiber::store::db_migrate::DbMigrate;
-use fiber::store::Store;
+use fiber_v061::store::{db_migrate::DbMigrate, Store};
```
chenyukang, line 2: why change this to `v061`

chenyukang (approved):

migrate/src/main.rs
quake: I removed `fiber = { package = "fnn", path = "../crates/fiber-lib" }` to resolve the secp256k1 deps issue.

PR #1093, opened 2026-02-02
fix: avoid blocking NetworkActor on ChannelReady with attempt index

jjyr:
## Summary

When a channel reconnects, the `ChannelReady` handler was scanning all payment attempts to find those using this channel as the first hop. This full table scan could block the NetworkActor for minutes on large databases, causing payment actor timeouts with panic:

```
thread 'tokio-runtime-worker' panicked at payment.rs:1560:10:
network actor must be alive: Timeout
```

### Storage Layer
- Add `ATTEMPT_CHANNEL_INDEX_PREFIX` secondary index mapping `(channel_outpoint, payment_hash, attempt_id) -> ()`
- Add `get_pending_attempts_by_channel_outpoint()` for O(m) indexed lookup where m = attempts for the specific channel
- Only return pending attempts (Created/Retrying status) to avoid retrying already completed ones
- Add `clear_attempts_channel_index()` to clear only channel index entries (keep attempt records for query)
- Keep `delete_attempts()` for its original purpose (delete both main records and index when retrying failed payment)

### Payment Layer
- When payment reaches final status (Success or Failed), call `clear_attempts_channel_index()` to remove stale index entries
- This prevents completed payments from being retried on channel ready
- Attempt records are preserved for querying fee, routers, retry_times, attempts_count

### Migration
- Add migration `mig_20260203` to build channel index for non-final payments
- Iterates all payment sessions, skips final status (Success/Failed)
- For each non-final payment, creates index entries for all attempts with first hop channel outpoint

## Performance

- **Before:** O(n) scan of all attempts + filter by status + filter by channel (could take minutes)
- **After:** O(m) prefix scan where m = pending attempts for channel (fast)

## Test plan

- [x] `make fmt` passes
- [x] `make clippy` passes  
- [x] `make test` - all 581 tests pass
- [x] `make check-migrate` passes

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/fiber-lib/src/store/store_impl/mod.rs
+++ b/crates/fiber-lib/src/store/store_impl/mod.rs
@@ -831,20 +831,6 @@ impl NetworkGraphStateStore for Store {
             .collect()
     }
 
-    fn get_attempts_with_statuses(&self, status: &[AttemptStatus]) -> Vec<Attempt> {
-        let prefix = [ATTEMPT_PREFIX];
-        self.prefix_iterator(&prefix)
-            .filter_map(|(_key, value)| {
-                let attempt: Attempt = deserialize_from(value.as_ref(), "Attempt");
-                if status.contains(&attempt.status) {
-                    Some(attempt)
-                } else {
-                    None
-                }
-            })
-            .collect()
-    }
-
     fn insert_payment_history_result(
         &mut self,
         channel_outpoint: OutPoint,
--- a/crates/fiber-lib/src/store/store_impl/mod.rs
+++ b/crates/fiber-lib/src/store/store_impl/mod.rs
@@ -762,38 +762,60 @@ impl NetworkGraphStateStore for Store {
         batch.commit();
     }
 
-    fn get_attempts_by_channel_outpoint(&self, channel_outpoint: &OutPoint) -> Vec<(Hash256, u64)> {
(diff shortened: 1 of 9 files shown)


crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -704,12 +733,104 @@ impl NetworkGraphStateStore for Store {
...
+                ) {
+                    Some(attempt)
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
     fn get_attempts_with_statuses(&self, status: &[AttemptStatus]) -> Vec<Attempt> {
```
chenyukang: if we don't use it any more, we can delete it.

migrate/Cargo.toml
```diff
@@ -15,6 +15,7 @@ thiserror = "1.0.58"
 serde = { version = "1.0.197", features = ["derive"] }
 hex = "0.4.3"
 clap = { version = "4.0", features = ["derive"] }
+ckb-types = "0.202.0"
```
quake:
```suggestion
ckb-types = "1"
```

quake:
@jjyr need rebase to resolve conflicts

jjyr:
@quake resolved

quake (approved):

PR #1112, opened 2026-02-08
Fix v0.7.0 migration issues

chenyukang:
1. I have ported the patch to `migrate/v0.6.0` and `migrate/v0.6.1`, now we can should use latest crates.
2. Fix issues in trampoline routing related migrations
3. Add `sample` trait for types which involved storage, we may write more robust unit testing for migration in future versions.


quake (approved):

gpBlockchain:
lgtm

PR #1153, opened 2026-02-28
add watchtower querier trait

doitian:
feat: implement watchtower querier trait and integrate with network actor

Allow network/channel actor to query tlc status from watchtower no matter it is in the same process or runs as a standalone service.

Refs <https://github.com/nervosnetwork/fiber/discussions/1077>

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -1400,7 +1400,7 @@ where
                                             WatchtowerQueryActorMessage::QueryPreimage {
                                                 channel_id,
                                                 tlc_id: tlc_id.into(),
-                                                payment_hash: payment_hash,
+                                                payment_hash,
                                             },
                                         );
                                     }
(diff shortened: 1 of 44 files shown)


jjyr (approved):

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1347,9 +1350,20 @@ where
                                     }
                                 }
 
-                                let Some(payment_preimage) =
-                                    self.store.get_preimage(&tlc.payment_hash)
-                                else {
+                                let mut payment_preimage =
+                                    self.store.get_preimage(&tlc.payment_hash);
+                                if payment_preimage.is_none() {
+                                    if let Some(querier) = self.watchtower_querier.as_ref() {
```
jjyr: IMO we should spawn an async task to query and write back to store. The tlcs will be solved in the next CheckChannels.

If each query cost 500ms, and we have more than 60 channels. it will block actor 30s from execute other messages. It is enough to trigger a channel timeout force close.

doitian: Offloaded to a dedicated actor

jjyr (approved):

doitian:
Need to add e2e test

doitian:
Troubleshooting #1222

doitian:
After discussion with Quake, we prioritize fixing the issue when watchtower is running together with fiber node. I'll close this PR and create a new one based on that assumption.

PR #1165, opened 2026-03-03
cch: add separate service mode with fiber RPC and WebSocket

doitian:
Enables running the cross-chain hub in a separate process from the Fiber node. The CCH connects to Fiber via HTTP RPC for orders and subscribes to store changes over WebSocket so it can react to payment and invoice state without co-location.

- CchFiberAgentRef unifies in-process (NetworkActor) and RPC HTTP backends for add_invoice, send_payment, and settle_invoice.
- New pubsub RPC module exposes subscribe_store_changes for WebSocket clients; CCH in separate mode uses it with auto-reconnect.
- Replace CchFiberStoreWatcher with StoreChange port and map_store_change_to_events in the actor.
- Binary: optional fiber config (store path fallback), RPC and CCH can start without co-located Fiber; CchConfig gains fiber_rpc_url.
- Bruno e2e suite and test node layout for cross-chain-hub-separate flow.


crates/fiber-lib/src/cch/actor.rs
```diff
@@ -695,6 +764,53 @@ impl<S: CchOrderStore> CchState<S> {
...
+                failure_reason: None,
+            }],
+            StoreChange::PutPaymentSession {
+                payment_hash,
+                payment_session,
+            } => {
+                use fiber_types::payment::PaymentStatus;
+                let status = payment_session.status;
+                // For successful payments, we need the preimage. If it's not in the same
+                // store change batch, the PutPreimage event will follow.
```
chenyukang, line 787: can we depends on the `StoreChange` event order? is it possible these two change events comes in an unexpected order?

tests/deploy/udt-init/src/main.rs
```diff
@@ -249,14 +250,55 @@ fn create_node_config(
...
         config["services"]
             .as_sequence_mut()
             .unwrap()
             .push(serde_yaml::Value::String("cch".to_string()));
     }
 
+    // Node 4 (cch) is a standalone CCH service connecting to node 3 via RPC
+    if node_index == 4 {
+        // Remove fiber and ckb sections — CCH-only node doesn't run these
+        config.as_mapping_mut().unwrap().remove("fiber");
```
chenyukang, line 264: This PR is explicitly designed to allow a CCH-only node without a local Fiber service, but the main startup path still relies on parsed_fiber() to derive the currency. is it a risk?

crates/fiber-lib/src/cch/actor.rs
doitian: It does not depends on the event order, it just ignores the event that payment session changes the status to success.

tests/deploy/udt-init/src/main.rs
doitian: The CCH node requires currency to verify the RPC paramters to create CCH order. If we don't let user configure it for the CCH node, we have to read it from fiber, which will introduce a startup dependency that user must start fiber node first.

If we accept this operation inconvenient, we can verify that the CCH node and Fiber node are running on the same chain. However, there are still ricks that user can restart Fiber with a different chain, or run multiple Fiber nodes under load balance with different chains.

quake (approved):

PR #1191, opened 2026-03-11
refactor: unify storage layer with StorageBackend and add sqlite support

quake:
Unify the storage layer behind a StorageBackend trait in fiber-store, replacing the previous pattern where each backend (RocksDB, IndexedDB, BTreeMap) had its own Store with duplicated iterator logic scattered across fiber-lib.
Changes

New StorageBackend trait (fiber-store/src/backend.rs)
- Defines get, put, delete, batch, and collect_iterator as the unified KV interface
- All 4 backends implement this trait: native/RocksDB, SQLite (new), browser/IndexedDB, browser\_test/BTreeMap
SQLite backend (fiber-store/src/sqlite.rs)
- New backend behind sqlite feature flag, using WAL mode and lazy row iteration
- SQL LIMIT clause pushed into queries when callers pass a limit, avoiding unbounded scans
FiberStore trait (fiber-lib/src/store/store_trait.rs)
- Convenience iterator methods (prefix_iterator, prefix_iterator_with_limit, prefix_iterator_with_skip_while_and_start, etc.) built on top of StorageBackend::collect_iterator
- Domain store trait impls (ChannelActorStateStore, InvoiceStore, etc.) now depend on FiberStore instead of per-backend Store

Removed / simplified
- Deleted store_impl/native.rs, store_impl/browser.rs, store_impl/browser_test.rs — per-backend wrappers in fiber-lib are no longer needed
- Eliminated WatchedStore decorator pattern — event notification inlined into the Store wrapper
- Removed BoxedKVPair type alias ((Box<[u8]>, Box<[u8]>)), replaced with KVPair { key: Vec<u8>, value: Vec<u8> }
- Removed IteratorStart enum from both fiber-store and fiber-wasm-db-common, pass Vec<u8> start key directly
- Removed FilterMapFn parameter from collect_iterator
- Removed Batch::get from all backends (was only used in one place, replaced with store.get())

Bug fixes
- Fixed browser/IndexedDB backend to use per-row take_while IPC instead of fetching the entire DB then filtering
- Fixed reverse iterator to actually start from the end of the prefix range (was silently broken — IteratorMode::End with set_prefix_same_as_start always returned empty)
- Updated test_query_missing_broadcast_message to account for the now-correct reverse iterator behavior

CI
- Added fiber-store, fiber-types, fiber-json-types to NATIVE_PACKAGES clippy checks
- Added sqlite feature flag forwarding through fiber-lib and fiber-bin


Makefile
```diff
@@ -21,13 +21,15 @@ check:
...
+	cargo check --features sqlite $(NATIVE_PACKAGES)
 	rustup target add wasm32-unknown-unknown
 	cargo check --target wasm32-unknown-unknown -p fiber-types --all-features
 	cd migrate && cargo check --locked
 
 .PHONY: clippy
 clippy:
 	cargo clippy --all-targets --all-features $(NATIVE_PACKAGES) -- -D warnings
+	cargo clippy --features sqlite $(NATIVE_PACKAGES) -- -D warnings
 	cargo clippy $(WASM_PACKAGES) --target wasm32-unknown-unknown -- -D warnings
```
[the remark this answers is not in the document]
quake: This actually works. When --features is used with multiple -p packages, Cargo applies the feature only to packages that define it and ignores the rest — it only errors when none of the selected packages recognize the feature. Verified locally:

```
$ cargo check --features sqlite -p fiber-types
# error: the package 'fiber-types' does not contain this feature: sqlite
$ cargo check --features sqlite -p fiber-types -p fnn
# OK — sqlite applied to fnn only, fiber-types unaffected
```

crates/fiber-store/src/browser.rs
```diff
@@ -193,12 +205,32 @@ impl Batch {
...
+
+    fn commit(self) {
+        self.chan
+            .dispatch_database_command(DbCommandRequestWithTakeWhile::Delete { keys: self.delete })
             .expect("Failed to delete batch");
         self.chan
-            .dispatch_database_command(DbCommandRequestWithSkipWhileFunc::Put { kvs: self.puts })
+            .dispatch_database_command(DbCommandRequestWithTakeWhile::Put { kvs: self.puts })
             .expect("Failed to put batch");
     }
```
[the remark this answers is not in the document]
quake: This is a pre-existing pattern from the original browser backend, not introduced by this PR. That said, the practical risk is low — the WASM IPC is synchronous (Atomics::wait), and the DB worker processes commands sequentially, so no concurrent reader can interleave between the two calls. The proper fix would be adding a BatchCommit IPC command on the worker side that wraps both operations in a single IndexedDB transaction — but that's a separate improvement, not a regression here.

crates/fiber-store/src/sqlite.rs
```diff
@@ -0,0 +1,241 @@
...
+        .map_err(|e| e.to_string())?;
+
+        // Create the key-value table.
+        // Note: PRIMARY KEY on `key` already creates an implicit B-tree index,
+        // so no additional index is needed for prefix/range scans on `key`.
+        conn.execute_batch(
+            "CREATE TABLE IF NOT EXISTS kv_store (
+                 key BLOB PRIMARY KEY NOT NULL,
+                 value BLOB NOT NULL
+             );",
```
chenyukang: add `WITHOUT ROWID` to eliminate the "Hidden" overhead?

chenyukang:
```console
cargo tree -p fnn --features sqlite -i ckb-rocksdb
ckb-rocksdb v0.21.1
└── fiber-store v0.7.1 (/Users/yukang/code/fiber/crates/fiber-store)
    └── fnn v0.7.1 (/Users/yukang/code/fiber/crates/fiber-lib)
```

if we enable feature `sqlite` it should not depend on `rocksdb`?

crates/fiber-lib/src/store/store_trait.rs
```diff
@@ -0,0 +1,123 @@
...
+                // For reverse iteration, start past the prefix range.
+                let mut end_key = prefix.to_vec();
+                let last = end_key
+                    .last_mut()
+                    .expect("empty prefix with Reverse is not supported");
+                if *last < 0xFF {
+                    *last += 1;
+                } else {
+                    end_key.push(0xFF);
+                }
```
chenyukang, line 132: This does not correctly represent the exclusive upper bound for prefixes ending in 0xFF, so some keys that still share the prefix can be skipped. It does not look exploitable with the current schema prefixes, but the helper is generic, so the boundary logic is still incorrect.

crates/fiber-store/src/sqlite.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::backend::{BatchWriter, StorageBackend, TakeWhileFn};
+use crate::iterator::{IteratorDirection, KVPair};
+
+/// SQLite-backed key-value store.
+#[derive(Clone, Debug)]
+pub struct Store {
+    conn: Arc<Mutex<Connection>>,
+}
+
+impl Store {
```
doitian, line 15: Every backend (`native.rs`, `sqlite.rs`, `browser.rs`, `browser_test.rs`) duplicates the same code between inherent methods and `StorageBackend`/`BatchWriter` trait implementations. For example in `sqlite.rs`, `Batch` has identical `put`/`delete`/`commit` in both the inherent impl and the `BatchWriter` impl (lines ~209-241). Same story for `Store::get`/`put`/`delete`.

The trait impls just delegate to the inherent methods (`self.get(key)`), so the inherent methods serve only to let callers avoid importing the trait. But since the domain code goes through `FiberStore`/`StorageBackend` anyway, the inherent methods seem unnecessary.

crates/fiber-store/src/sqlite.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    pub fn delete<K: AsRef<[u8]>>(&mut self, key: K) {
+        self.operations.push(BatchOp::Delete {
+            key: key.as_ref().to_vec(),
+        });
+    }
+
+    pub fn commit(self) {
+        let conn = self.conn.lock().expect("lock poisoned");
+        conn.execute_batch("BEGIN").expect("begin transaction");
```
doitian: If any operation panics between BEGIN and COMMIT, the connection is left in a dirty state. Rusqlite provides `Connection::transaction()` which gives RAII-based rollback-on-drop safety.

doitian:
Suggest to add the test `cargo nextest run --features sqlite` at least when pushing to develop.

crates/fiber-lib/src/store/store_trait.rs
```diff
@@ -0,0 +1,123 @@
...
+pub type SkipWhileFn = Box<dyn Fn(&[u8]) -> bool + Send + 'static>;
+
+/// Trait that extends `StorageBackend` with convenience iterator methods
+/// used by the domain store trait implementations.
+///
+/// The `fiber_store::Store` type (which cfg-selects between RocksDB, IndexedDB,
+/// BTreeMap, or SQLite backends) implements this trait, and the domain trait
+/// implementations (`ChannelActorStateStore`, `InvoiceStore`, etc.) are written
+/// in terms of these convenience methods.
+pub trait FiberStore: StorageBackend + Clone + std::fmt::Debug {
```
doitian, line 82: All these functions will create a Vec first, then call `into_iter` on the Vec. Why not just return the Vec? What's the design decision here?

doitian, line 82: I would prefer a uniform `prefix_iterator(&self, prefix: &[u8], options: PrefixOptions)` interface and use a builder for `PrefixOptions` like `std::fs::OpenOptions`. At least it can avoid the awkward name `prefix_iterator_with_skip_while_and_start_and_limit`.

crates/fiber-store/src/sqlite.rs
```diff
@@ -0,0 +1,241 @@
...
+        // The take_while predicate may stop iteration before the SQL limit is
+        // reached, but having the SQL limit prevents unbounded scans when the
+        // caller already knows how many rows it needs.
+        let sql = match (direction, limit) {
+            (IteratorDirection::Forward, 0) => {
+                "SELECT key, value FROM kv_store WHERE key >= ?1 ORDER BY key ASC".to_string()
+            }
+            (IteratorDirection::Forward, n) => {
+                format!(
+                    "SELECT key, value FROM kv_store WHERE key >= ?1 ORDER BY key ASC LIMIT {n}"
```
doitian: It's cache friendly to use `LIMIT ?2` instead of interpolating the limit into the SQL.

crates/fiber-bin/src/main.rs
```diff
@@ -495,116 +498,6 @@ fn forward_event_to_actor(
     }
 }
 
-async fn forward_event_to_client<T: WatchtowerRpcClient + Sync>(
```
doitian, line 498: Good refactoring :+1:

crates/fiber-lib/src/fiber/tests/network.rs
```diff
@@ -459,14 +459,17 @@ async fn test_query_missing_broadcast_message() {
...
     let channel_announcement = channel_context.channel_announcement.clone();
-    // A timestamp so large that the other node will unlikely try to send GetBroadcastMessages
-    // with timestamp smaller than this one.
-    let long_long_time_ago = now_timestamp_as_millis_u64() - 30 * 24 * 3600 * 100;
+    // A timestamp in the recent past. It needs to be recent enough to pass the
+    // broadcast subscription filter (which is latest_cursor - MAX_MISSING_BROADCAST_MESSAGE_TIMESTAMP_DRIFT,
+    // i.e. ~2 hours), but old enough to be distinct from the current timestamp.
+    // Node2 won't sync this during active syncing because it hasn't submitted the funding tx yet,
+    // not because of the timestamp.
+    let not_long_ago = now_timestamp_as_millis_u64() - 60 * 60 * 1000;
```
doitian, line 467: Why the change? It seems tricky.

crates/fiber-lib/src/store/store_trait.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// Start iteration from the given key instead of the prefix boundary.
+    pub fn start_key(mut self, key: &'a [u8]) -> Self {
+        self.start_key = Some(key);
+        self
+    }
+
+    /// Skip entries at the start while the predicate returns true.
+    /// Once a key fails the predicate, all subsequent entries are kept.
+    pub fn skip_while(mut self, f: SkipWhileFn) -> Self {
```
jjyr, line 57: Please remove the space

jjyr (approved):

crates/fiber-lib/src/fiber/tests/network.rs
quake: this unit test original version is a flaky test, this change fixed it, the reason has been wroten in the comment

chenyukang (approved):

crates/fiber-lib/src/store/store_trait.rs
```diff
@@ -12,82 +80,55 @@ pub type SkipWhileFn = Box<dyn Fn(&[u8]) -> bool + Send + 'static>;
...
-    /// Like `prefix_iterator`, but caps the number of entries returned.
-    ///
-    /// `limit` = 0 means no limit.
-    fn prefix_iterator_with_limit(
-        &self,
-        prefix: &[u8],
-        limit: usize,
-    ) -> std::vec::IntoIter<KVPair> {
+    /// Collect all entries whose key starts with `prefix`, in forward order.
+    fn prefix_iter(&self, prefix: &[u8]) -> Vec<KVPair> {
```
doitian: Trivial suggestion: consider rename

1. `prefix_iter` -> `collect_by_prefix`
2. `prefix_iter_with` -> `collect_by_prefix_with`

doitian (approved):

PR #1227, opened 2026-03-22
Fix gossip startup sync cursor pollution

jjyr:
Our node used the latest broadcast cursor in the local store to decide where to start syncing gossip, but that cursor mixed two different sources of state: gossip learned from remote peers and announcements generated by the local node itself.

When auto_announce was enabled, the node saved its own NodeAnnouncement as soon as a peer connected. That immediately pushed the “latest broadcast cursor” close to the current time.

## Fix
- store the last remote broadcast cursor and use it for startup gossip sync watermarks
- update the remote cursor only when newly verified gossip learned from peers is saved

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -176,6 +176,10 @@ pub trait GossipMessageStore {
 
     fn get_latest_broadcast_message_cursor(&self) -> Option<Cursor>;
 
+    fn get_latest_remote_broadcast_message_cursor(&self) -> Option<Cursor>;
+
+    fn update_latest_remote_broadcast_message_cursor(&self, cursor: Cursor);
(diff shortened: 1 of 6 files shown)


quake:
good catch, and the fix looks good to me, but I would like to suggest a simpler alternative:

Keep an in-memory last_remote_cursor on GossipActorState,
1. Initialize it from store.get_latest_broadcast_message_cursor() in GossipActor::pre_start, at this point the store has not yet been polluted because auto_announce only fires later in on_peer_connected
2. Update it in prune_messages_to_be_saved() (the only path for verified remote messages) when remote messages are actually stored
3. Use it in get_safe_cursor_to_start_syncing() instead of get_latest_cursor()

The advantage: no schema change, no new persistent key, no migration concern.

quake (approved):

PR #1217, opened 2026-03-18
channel: keep actors alive across disconnects

jjyr:
## Purpose

The Network actor is becoming single point of the system.

This PR keeps the channel actor alive even peer is disconnected, so we can offload channel's bussiness logic from network actor to channel actors.

## Summary

- Keep channel actors alive across peer disconnects instead of stopping and recreating them.
- Introduce explicit channel connectivity handling (`Offline -> Syncing -> Online`) and let the channel actor drive reestablish on reconnect.
- Add official `ChannelOnline` / `ChannelOffline` service events so reconnect tests can wait on channel connectivity instead of actor recreation or fixed sleeps.
- Offload channel tlcs timeout check from network actor to channel actors.

## Secondary changes
- Rework reconnect bookkeeping around `Pubkey` so requested disconnects and backoff reconnects stay aligned with long-lived actors.
- Preserve retryable remove-TLC preimages during reestablish and avoid settling forwarding TLCs that are still waiting on downstream results.

PR #1228, opened 2026-03-23
network: onion & socks5 support for fiber

Officeyutong:
This PR adds onion & socks5 proxy support for fiber, basically based on https://github.com/nervosnetwork/ckb/pull/4733
This enables Fiber nodes to:
- Route all outbound P2P connections through a SOCKS5 proxy (e.g. Tor)
- Connect to .onion addresses via a dedicated Tor SOCKS5 proxy
- Host a Tor onion hidden service, making the node reachable via a .onion address


gpBlockchain:
lgtm

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -816,6 +816,153 @@ where
...
+        &self,
+        config: &FiberConfig,
+        listening_addrs: &[MultiAddr],
+        my_peer_id: &tentacle::secio::PeerId,
+        tracker: &tokio_util::task::TaskTracker,
+        myself: ActorRef<NetworkActorMessage>,
+    ) -> Result<Option<(MultiAddr, tokio_util::sync::CancellationToken)>, String> {
+        use std::net::{Ipv4Addr, SocketAddr};
+
+        if config.onion.onion_server.is_none() && config.proxy.proxy_url.is_none() {
```
chenyukang: This guard makes `listen_on_onion` depend on either `onion_server` or `proxy.proxy_url`, but creating a Tor hidden service only requires the control port plus the local forwarding address. As written, a node that only wants to expose an inbound `.onion` address will silently skip hidden-service setup unless it also configures an outbound SOCKS5 proxy. That seems inconsistent with the config docs for `listen_on_onion` and `tor_controller`. Could we remove this check, or restrict it to the outbound `.onion` dialing path instead?

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -816,6 +816,153 @@ where
...
+        let peer_id_str = my_peer_id.to_base58();
+        let (onion_service, onion_addr) =
+            super::onion_service::OnionService::new(onion_config, &peer_id_str)?;
+
+        let cancel_token = tokio_util::sync::CancellationToken::new();
+        let token_clone = cancel_token.clone();
+        let (reconnect_tx, mut reconnect_rx) = tokio::sync::mpsc::unbounded_channel();
+
+        tracker.spawn(async move {
+            if let Err(err) = onion_service.start(token_clone, reconnect_tx).await {
```
chenyukang: The Tor auth, bootstrap wait, and `ADD_ONION` call all happen asynchronously inside `OnionService::start()` / `launch_onion_service()`, after `start_onion_service()` has already returned the derived onion address.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -5258,6 +5448,38 @@ where
...
+                    &listening_addr,
+                    &my_peer_id,
+                    &tracker,
+                    myself.clone(),
+                )
+                .await
+            {
+                Ok(Some((addr, token))) => {
+                    info!("Onion service address: {}", addr);
+                    announced_addrs.push(addr);
```
chenyukang, line 5503: `start_onion_service()` does not wait for the hidden service to be successfully registered with Tor. At this point we only have a derived onion address plus a background task that will try to authenticate, bootstrap, and call `ADD_ONION` later. Pushing `addr` into `announced_addrs` here means we may advertise an unreachable onion address if Tor setup fails. Could we delay announcing it until the first successful hidden-service registration completes?

crates/fiber-lib/src/fiber/network.rs
Officeyutong: This line has been removed now. Note: Onion implementation in CKB also has this line, maybe we also need to remove that.

crates/fiber-lib/src/fiber/network.rs
Officeyutong: The race condition was fixed now: There is a oneshot channel which will be triggered when `launch_onion_service` succeeds or fails. Other logic of `start_onion_service` will wait for this channel returns.

crates/fiber-lib/src/fiber/network.rs
Officeyutong: Also fixed by the oneshot channel

config/mainnet/config.yml
```diff
@@ -14,6 +14,32 @@ fiber:
...
+  #   ## If not set, it is derived from listening_addr. e.g. "127.0.0.1:8228"
+  #   p2p_listen_address: "127.0.0.1:8228"
+  #   ## Path to store the onion service private key. [default: $BASE_DIR/fiber/onion_private_key]
+  #   onion_private_key_path: ""
+  #   ## Tor controller address. [default: 127.0.0.1:9051]
+  #   tor_controller: "127.0.0.1:9051"
+  #   ## Tor controller hashed password (if HashedControlPassword is set in torrc).
+  #   tor_password: ""
+  #   ## The external port exposed by the onion service. [default: 8115]
+  #   onion_external_port: 8115
```
eval-exec: CKB's `onion_external_port` choose 8115, because ckb's default p2p port number is 8115.
What's fiber's default p2p port number?

crates/fiber-lib/src/fiber/onion_service.rs
```diff
@@ -0,0 +1,422 @@
...
+            format!(
+                "Failed to parse onion address {} to multiaddr: {:?}",
+                onion_multi_addr_str, err
+            )
+        })?;
+
+        Ok((OnionService { config, key }, onion_multi_addr))
+    }
+
+    /// Start the onion service. This will loop and reconnect if the Tor connection drops.
```
eval-exec:
```suggestion
    /// Start the onion service. This will loop and reconnect if the connection to Tor Server drops.
```

crates/fiber-lib/src/fiber/onion_service.rs
```diff
@@ -0,0 +1,422 @@
...
+                    if let Some(tx) = ready_tx.take() {
+                        let _ = tx.send(Ok(()));
+                    }
+                    if !first_start {
+                        // Tor has reconnected after a failure, notify the network actor
+                        // to re-establish peer connections that were dropped.
+                        info!("Tor reconnected, notifying network to reconnect peers");
+                        let _ = reconnect_notify.send(());
+                    }
+                    first_start = false;
```
eval-exec, line 156: Suggest move `first_start = false` to bellow the entier match pattern.

config/mainnet/config.yml
Officeyutong: It's `8228`. Fixed now

crates/fiber-lib/src/fiber/network.rs
eval-exec: Your check is different with CKB:
https://github.com/nervosnetwork/ckb/blob/a45a0943048688d8bdaef3a8adbd608b645d2152/util/launcher/src/lib.rs#L277-L291


CKB check `self.args.config.network.onion.listen_on_onion` first, if listen_on_onion is true, then check `if onion_config.onion_server.is_none() && self.args.config.network.proxy.proxy_url.is_none()`

This means that when listen_on_onion = true but onion_config.onion_server is not set, CKB will fall back to using config.network.proxy.proxy_url as the Tor server endpoint.

Bitcoin has a similar behavior: https://github.com/bitcoin/bitcoin/blob/master/doc/tor.md?utm_source=chatgpt.com#1-run-bitcoin-core-behind-a-tor-proxy

crates/fiber-lib/src/fiber/onion_service.rs
Officeyutong: Fixed

eval-exec (approved):

.gitignore
```diff
@@ -34,3 +34,4 @@ target
 baseline.json
 current.json
 comparison_report.txt
+/test-config
```
quake: unrelated change?

crates/fiber-lib/src/fiber/config.rs
```diff
@@ -401,6 +401,83 @@ pub struct FiberConfig {
         help = "Address for metrics endpoint (e.g., 127.0.0.1:9090). Requires binary to be compiled with RUSTFLAGS=\"--cfg tokio_unstable\" and metrics feature enabled"
     )]
     pub metrics_addr: Option<String>,
+
+    /// SOCKS5 proxy configuration
+    #[arg(skip)]
+    #[serde(default)]
+    pub proxy: ProxyConfig,
```
quake, line 411: suggest to add a none wasm cfg condition for these two options

.gitignore
Officeyutong: I used this folder to contain config files that enables tor on my local development environment

quake: please use local exclude info:
```
echo test-config>> .git/info/exclude
```

Officeyutong: Fixed now

crates/fiber-lib/src/fiber/config.rs
Officeyutong: Added `#[cfg(not(target_arch = "wasm32"))]` for these two config section

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -816,6 +816,167 @@ where
...
+            if let Err(err) = onion_service
+                .start(token_clone, reconnect_tx, ready_tx)
+                .await
+            {
+                error!("Onion service stopped with error: {}", err);
+            }
+        });
+
+        // Wait for the onion service to successfully register with Tor before
+        // returning the address, so callers don't advertise an unreachable address.
```
chenyukang, line 948: do we need to add a timeout for waiting?
a Tor instance that is reachable on the control port but never reaches Done can prevent the whole node from starting now.
waiting with timeout error should report with proper error and exit.

crates/fiber-lib/src/fiber/onion_service.rs
```diff
@@ -0,0 +1,420 @@
...
+                    if !first_start {
+                        // Tor has reconnected after a failure, notify the network actor
+                        // to re-establish peer connections that were dropped.
+                        info!("Tor reconnected, notifying network to reconnect peers");
+                        let _ = reconnect_notify.send(());
+                    }
+                }
+                Err(err) => {
+                    error!("Failed to start onion service: {}", err);
+                    if let Some(tx) = ready_tx.take() {
```
chenyukang: do we need to retry for the first time error? since I see there is a variable `first_start` here, but we will return without retry for the first time.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -816,6 +816,167 @@ where
...
+        tracker.spawn(async move {
+            loop {
+                tokio::select! {
+                    msg = reconnect_rx.recv() => {
+                        if msg.is_none() {
+                            break;
+                        }
+                        info!("Tor reconnected, triggering MaintainConnections");
+                        let _ = myself.send_message(NetworkActorMessage::new_command(
+                            NetworkActorCommand::MaintainConnections,
```
jjyr, line 989: We may have a potential time race here.

Can we ensure the DisconnectPeer messages are handled before the trigger of `MaintainConnections`?

If `MaintainConnnection` is fired before DisconnectPeer, the peer will stay disconnected until the next time.

crates/fiber-lib/src/fiber/config.rs
jjyr: We can move 

`ProxyConfig` -> `proxy.rs`
`OnionConfig` -> `onion_service.rs`

The two config structure will be gated with the two modules.

crates/fiber-lib/src/fiber/onion_service.rs
Officeyutong: Fixed now

crates/fiber-lib/src/fiber/network.rs
Officeyutong: Fixed now

crates/fiber-lib/src/fiber/network.rs
Officeyutong: Fixed. Added a delay of 3 seconds to ensure `DisconnectPeer` was handled properly

jjyr: https://github.com/nervosnetwork/fiber/pull/1217 
Introduces an auto reconnect mechanism can matigate this issue.
So IMO we can simply skip this issue in the current PR

chenyukang (approved):

crates/fiber-lib/src/fiber/network.rs
Officeyutong: Got it

PR #1235, opened 2026-03-27
RPC method not found should not return unauthorized when auth is disabled

jjyr:
When biscuit auth is disabled, the rpc method not found should not return unauthorized error.

### DIFF
--- a/crates/fiber-lib/src/rpc/middleware.rs
+++ b/crates/fiber-lib/src/rpc/middleware.rs
@@ -105,8 +105,9 @@ impl<S> BiscuitAuthMiddleware<S> {
                     return true;
                 }
                 Err(err) => {
-                    tracing::debug!("Failed check_permission #{err:?}");
-                    return false;
+                    tracing::debug!("Failed get_rule #{err:?}");
+                    // no auth rule, but allow local rpc to proceed.
+                    return true;
                 }
             }
         }


crates/fiber-lib/src/rpc/middleware.rs
```diff
@@ -105,8 +105,9 @@ impl<S> BiscuitAuthMiddleware<S> {
                     return true;
                 }
                 Err(err) => {
-                    tracing::debug!("Failed check_permission #{err:?}");
-                    return false;
+                    tracing::debug!("Failed get_rule #{err:?}");
+                    // no auth rule, but allow local rpc to proceed.
+                    return true;
```
chenyukang, line 110: I found an issue that if some dev add a new rpc, but forget to config auth rules, I remember the error is "method not found", it's better to report it's not authorized?

it's a separate issue with this one.

jjyr: I am not sure if this issue still exist, you can check the middleware module, the only error we return from middleware is Unauthorized

https://github.com/jjyr/fiber/blob/099bd414d3f26783b3ad540a3fa8f2eaa9739fab/crates/fiber-lib/src/rpc/middleware.rs#L202C27-L202C43

quake (approved):

PR #1237, opened 2026-03-28
feat: update macOS build configuration for portability

chenyukang:
Add release target for `aarch64-apple-darwin`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04, ubuntu-22.04-arm, macos-15-intel, windows-2022 ]
+        os: [ ubuntu-22.04, ubuntu-24.04, ubuntu-22.04-arm, macos-15, macos-15-intel, windows-2022 ]
         include:
           - os: ubuntu-22.04
             bin_suffix:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -115,11 +118,17 @@ jobs:
...
+        echo /opt/homebrew/bin >> $GITHUB_PATH
+        echo /opt/homebrew/sbin >> $GITHUB_PATH
+    - if: startsWith(matrix.os, 'macos-15')
       name: Build macos portable
       run: |
-        export OPENSSL_LIB_DIR=/usr/local/opt/openssl@1.1/lib
-        export OPENSSL_INCLUDE_DIR=/usr/local/opt/openssl@1.1/include
+        export OPENSSL_PREFIX="$(brew --prefix openssl@1.1)"
+        export OPENSSL_LIB_DIR="${OPENSSL_PREFIX}/lib"
+        export OPENSSL_INCLUDE_DIR="${OPENSSL_PREFIX}/include"
```
eval-exec: Since tentacle-secio provides the openssl-vendor feature, we could statically link OpenSSL into the Fiber binary by enabling this feature.
https://github.com/nervosnetwork/tentacle/blob/87ef6d9bd659012bb1394f5f3e8ccd4f8e615197/secio/Cargo.toml#L53

chenyukang: the feature is already enabled at https://github.com/chenyukang/fiber/blob/f4e82901dd7d0741360fe1f8cf15786214b1809c/crates/fiber-lib/Cargo.toml#L67-L74

so I will just remove these env variables.

PR #1238, opened 2026-03-30
Optimize Fiber startup by limiting gossip broadcast message loading

joii2020:
## Summary

This improves Fiber startup time, with the most noticeable impact on `fiber-wasm`.

The change avoids materializing all remaining gossip broadcast messages when only the next page is needed during startup replay. Instead, it applies the requested limit at the storage layer.

## Impact

Using [fiber-wallet](https://github.com/joii2020/fiber-wallet/blob/master/apps/src/services/fiber-wasm.ts#L55), after a channel is successfully established and the page is refreshed again:

- before: startup took about `58s`
- after: startup takes about `3s`

This is especially beneficial for `fiber-wasm`, where the extra store iteration / IPC / serialization cost is much more visible than on native.

### DIFF
--- a/crates/fiber-lib/src/store/store_impl/mod.rs
+++ b/crates/fiber-lib/src/store/store_impl/mod.rs
@@ -1290,6 +1290,39 @@ impl WatchtowerStore for Store {
 }
 
 impl GossipMessageStore for Store {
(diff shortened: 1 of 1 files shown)


quake:
close via #1240

PR #1240, opened 2026-03-31
fix: add limit to broadcast message queries and use loop-based pagination

quake:
Replace unbounded get_broadcast_messages_iter with limit-aware get_broadcast_messages(cursor, limit) to avoid loading all broadcast messages into memory. Callers like get_nodes_with_params and get_channels_with_params now use loop-based pagination to correctly handle filtering (e.g. skipping closed channels) while respecting the requested limit.

### DIFF
--- a/crates/fiber-lib/src/fiber/graph.rs
+++ b/crates/fiber-lib/src/fiber/graph.rs
@@ -656,7 +656,7 @@ where
     // Process them and set nodes and channels accordingly.
     pub(crate) fn load_from_store(&mut self) {
         loop {
-            let messages = self.store.get_broadcast_messages(&self.latest_cursor, None);
+            let messages = self.store.get_broadcast_messages(&self.latest_cursor, 0);
             if messages.is_empty() {
                 break;
             }
@@ -887,18 +887,27 @@ where
     }
 
     pub fn get_nodes_with_params(&self, limit: usize, after: Option<Cursor>) -> Vec<NodeInfo> {
-        let cursor = after.unwrap_or_default();
-        self.store
-            .get_broadcast_messages_iter(&cursor)
-            .into_iter()
-            .filter_map(|message| match message {
-                BroadcastMessageWithTimestamp::NodeAnnouncement(node_announcement) => {
-                    Some(NodeInfo::from(node_announcement))
+        let mut cursor = after.unwrap_or_default();
+        let mut result = Vec::new();
+        loop {
+            let batch = self
+                .store
+                .get_broadcast_messages(&cursor, if limit > 0 { limit - result.len() } else { 0 });
+            if batch.is_empty() {
+                break;
+            }
+            for message in batch {
+                cursor = message.cursor();
+                if let BroadcastMessageWithTimestamp::NodeAnnouncement(node_announcement) = message
+                {
+                    result.push(NodeInfo::from(node_announcement));
+                    if limit > 0 && result.len() >= limit {
+                        return result;
+                    }
                 }
-                _ => None,
-            })
-            .take(limit)
-            .collect()
+            }
+        }
+        result
     }
 
     pub fn get_node(&self, node_id: &Pubkey) -> Option<&NodeInfo> {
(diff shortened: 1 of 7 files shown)


crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -1290,38 +1290,40 @@ impl WatchtowerStore for Store {
...
-            (message, timestamp).into()
-        })
-        .collect::<Vec<_>>()
+        let mut options = PrefixIterOptions::new()
+            .start_key(&start)
+            .skip_while(Box::new(move |key: &[u8]| key == start_cloned));
+        if limit > 0 {
+            options = options.limit(limit);
+        }
+        self.collect_by_prefix_with(&prefix, options)
```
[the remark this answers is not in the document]
quake: let's resolve this issue in other PR

crates/fiber-lib/src/fiber/graph.rs
```diff
@@ -887,18 +887,27 @@ where
...
+                break;
+            }
+            for message in batch {
+                cursor = message.cursor();
+                if let BroadcastMessageWithTimestamp::NodeAnnouncement(node_announcement) = message
+                {
+                    result.push(NodeInfo::from(node_announcement));
+                    if limit > 0 && result.len() >= limit {
+                        return result;
+                    }
```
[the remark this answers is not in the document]
quake: Good point, but in practice the hit rate is ~1/3 so it converges quickly, and these are low-frequency RPC endpoints, happy to optimize as a follow-up if needed.

crates/fiber-lib/src/fiber/graph.rs
```diff
@@ -949,31 +958,36 @@ where
...
-
-                    // assuming channel is closed if disabled from the both side
                     let is_closed = channel_info.update_of_node1.is_some_and(|u| !u.enabled)
                         && channel_info.update_of_node2.is_some_and(|u| !u.enabled);
                     if !is_closed {
-                        Some(channel_info)
-                    } else {
-                        None
+                        result.push(channel_info);
+                        if limit > 0 && result.len() >= limit {
```
chenyukang, line 983: if limit is 0, old implementation will return zero items, now seems we will not stop.

quake: yes, 0 means no limit now, and the checking has been moved to GossipMessage::GetBroadcastMessages.

crates/fiber-lib/src/store/store_impl/mod.rs
```diff
@@ -1290,38 +1286,39 @@ impl WatchtowerStore for Store {
...
-            (message, timestamp).into()
-        })
-        .collect::<Vec<_>>()
+        let mut options = PrefixIterOptions::new()
+            .start_key(&start)
+            .start_key_exclusive();
+        if limit > 0 {
+            options = options.limit(limit);
+        }
+        self.collect_by_prefix_with(&prefix, options)
```
chenyukang, line 1303: similar issue, how about limit == 0, the meaning seems changed.

crates/fiber-lib/src/fiber/graph.rs
chenyukang: prefer to use `Option<usize>` to limit, so `None` means no limit, `Some(0)` will be empty.

crates/fiber-lib/src/store/store_trait.rs
```diff
@@ -134,32 +129,48 @@ pub trait FiberStore: StorageBackend + Clone + std::fmt::Debug {
...
-                            if predicate(&kv.key) {
-                                return false;
-                            }
-                            skipping = false;
-                        }
-                        true
-                    })
-                    .collect()
+        // Strip the cursor entry itself when exclusive.
+        if let Some(ref cursor_key) = start_key_owned {
```
chenyukang, line 156:
```rust
#[cfg_attr(not(target_arch = "wasm32"), test)]
#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
fn test_store_get_broadcast_messages_reverse_excludes_cursor() {
    let (store, _dir) = generate_store();
    let first_timestamp = now_timestamp_as_millis_u64();
    let first_channel_announcement = mock_channel();
    let first_outpoint = first_channel_announcement.out_point().clone();
    store.save_channel_announcement(first_timestamp, first_channel_announcement.clone());

    let second_timestamp = first_timestamp + 1;
    let second_channel_announcement = mock_channel();
    let second_outpoint = second_channel_announcement.out_point().clone();
    store.save_channel_announcement(second_timestamp, second_channel_announcement.clone());

    let latest = store.get_broadcast_messages_reverse(None, 1);
    assert_eq!(
        latest,
        vec![BroadcastMessageWithTimestamp::ChannelAnnouncement(
            second_timestamp,
            second_channel_announcement.clone(),
        )],
    );

    let before_cursor = Cursor::new(
        second_timestamp,
        BroadcastMessageID::ChannelAnnouncement(second_outpoint),
    );
    let previous = store.get_broadcast_messages_reverse(Some(&before_cursor), 1);
    assert_eq!(
        previous,
        vec![BroadcastMessageWithTimestamp::ChannelAnnouncement(
            first_timestamp,
            first_channel_announcement,
        )],
    );

    let first_cursor = Cursor::new(
        first_timestamp,
        BroadcastMessageID::ChannelAnnouncement(first_outpoint),
    );
    assert!(store
        .get_broadcast_messages_reverse(Some(&first_cursor), 1)
        .is_empty());
}
```

get_broadcast_messages_reverse(Some(cursor), limit) is intended to treat cursor as an exclusive upper bound, meaning the result should start from the message before that cursor. The test inserts two broadcast messages, then asks for one message before the second cursor. The expected result is the first message.

The current implementation of start_key_exclusive() removes the wrong element for reverse iteration. All current storage backends return the start_key as the first item in a reverse scan, but the code checks results.last() and removes the tail. That means the cursor entry is kept, and a real previous entry may be dropped. In practice, get_broadcast_messages_reverse(Some(cursor), 1) returns the cursor message again instead of the previous message.

crates/fiber-lib/src/store/store_impl/mod.rs
quake: 0 = no limit in current dev branch:
https://github.com/nervosnetwork/fiber/blob/0b28bcd8a1af3d0e76a05f4744dd234e6c9675bb/crates/fiber-store/src/backend.rs#L43

if you want to change collect_iterator fn to use `limit: Option<usize>` also, let's do it in sperate PR

chenyukang (approved):

PR #1242, opened 2026-04-01
Add CHANGELOG, migration guide, and data migrations for v0.8.0-rc1

quake:
- CHANGELOG.md: comprehensive changelog covering all breaking changes, features, bug fixes, and refactoring between v0.7.1 and v0.8.0-rc1
- docs/notes/v0.8.0-migration-guide.md: user-facing guide for in-place database migration via fnn-migrate (no channel closing required)
- mig_20260302_channel_replay_fields: append pending_replay_updates and last_was_revoke default values to channel actor state (prefix 0)
- mig_20260302_channel_open_record: delete ephemeral channel open records (prefix 201) incompatible due to PeerId-to-Pubkey field change
- refactor: use local fiber-store for migration infra, move check_validate to fnn binary

### DIFF
--- a/crates/fiber-bin/src/main.rs
+++ b/crates/fiber-bin/src/main.rs
@@ -72,6 +72,21 @@ pub async fn main() -> Result<(), ExitMessage> {
     let _span = info_span!("node", node = fnn::get_node_prefix()).entered();
 
     let config = Config::parse();
+
+    if config.check_validate {
+        let parsed_fiber_config = config
+            .parsed_fiber()
+            .ok_or(ExitMessage("fiber config must be set".to_string()))?;
+        let store_path = parsed_fiber_config.store_path();
+        if let Err(err) = fnn::store::check_validate(&store_path) {
+            eprintln!("db validate failed:\n{}", err);
+            std::process::exit(1);
+        } else {
+            println!("db validate success");
+            std::process::exit(0);
+        }
+    }
+
     let parsed_fiber_config = config
         .parsed_fiber()
         .ok_or(ExitMessage("fiber config must be set".to_string()))?;
--- a/crates/fiber-bin/src/main.rs
+++ b/crates/fiber-bin/src/main.rs
@@ -77,7 +77,13 @@ pub async fn main() -> Result<(), ExitMessage> {
         let parsed_fiber_config = config
             .parsed_fiber()
             .ok_or(ExitMessage("fiber config must be set".to_string()))?;
-        let store_path = parsed_fiber_config.store_path();
+        let store_path = parsed_fiber_config.base_dir().join("store");
+        if !store_path.exists() {
+            return Err(ExitMessage(format!(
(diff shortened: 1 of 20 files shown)


migrate/src/migrations/mig_20260302_channel_open_record.rs
```diff
@@ -0,0 +1,71 @@
...
+        &self,
+        db: &'a Store,
+        _pb: Arc<dyn Fn(u64) -> ProgressBar + Send + Sync>,
+    ) -> Result<&'a Store, Error> {
+        info!(
+            "MigrationObj::migrate to {} - clearing channel open record entries ...",
+            MIGRATION_DB_VERSION
+        );
+
+        let prefix = vec![CHANNEL_OPEN_RECORD_PREFIX];
```
doitian, line 46: Is it possible to use `DeleteRange`?

quake: no, we are using store trait now, not rocksdb impl.

docs/notes/v0.8.0-migration-guide.md
```diff
@@ -0,0 +1,143 @@
...
+Or download the pre-built binaries from the
+[releases page](https://github.com/nervosnetwork/fiber/releases).
+
+## Step 4: Run the migration tool
+
+```bash
+./fnn-migrate -p /path/to/fiber-dir/store
+```
+
+The `-p` / `--path` flag takes the path to the RocksDB store directory (typically
```
chenyukang: do we need to change migrate `-p` to `-d`? keep consistent with fnn node.

docs/notes/v0.8.0-migration-guide.md
```diff
@@ -0,0 +1,143 @@
...
+tar -zcvf backup-fiber-dir.tar.gz fiber-dir
+```
+
+## Step 3: Build or download v0.8.0
+
+Build from source:
+
+```bash
+git checkout v0.8.0-rc1
+cargo build --release -p fnn -p fnn-migrate
```
chenyukang: `fnn-migrate` is not a package of fiber, it's standalone project, so this command will fail.

crates/fiber-bin/src/main.rs
```diff
@@ -72,6 +72,27 @@ pub async fn main() -> Result<(), ExitMessage> {
     let _span = info_span!("node", node = fnn::get_node_prefix()).entered();
 
     let config = Config::parse();
+
+    if config.check_validate {
```
chenyukang, line 76: This removes the old `fnn-migrate --check-validate -p <store>` flow. Now validation only works for `<fiber base dir>/store` after config parsing, so scripts and offline checks for other store paths will break. We should either keep this mode in `fnn-migrate` or add a path flag to `fnn`?

chenyukang (approved):

PR #1249, opened 2026-04-04
Fix fnn-migrate error messages and update README

gpBlockchain:


chenyukang (approved):

crates/fiber-store/src/db_migrate.rs
```diff
@@ -87,7 +87,8 @@ impl<'a> DbMigrate<'a> {
                                 .to_string(),
                         );
                     }
-                    return Err(format!("Fiber need to run some database migrations, please run `fnn-migrate -p {}` to start migrations.", path.as_ref().display()));
+                    let dir = path.as_ref().parent().unwrap_or(path.as_ref());
```
eval-exec, line 92: Evaluating `.parent()` at here feels bad, how about pass fiber's base directory path from upper level?

quake: @gpBlockchain please check this review comment

gpBlockchain: It seems that modifying upwards also requires calling parent, adding comments here would be simpler

```suggestion
                    // The path points to the store directory (e.g. <base_dir>/store),
                    // use .parent() to get the fiber base directory for the fnn-migrate -d flag.
                    let dir = path.as_ref().parent().unwrap_or(path.as_ref());
```

PR #1252, opened 2026-04-07
Persist external funding state across restarts

joii2020:
## Summary

This PR fixes a restart-recovery issue in the external funding flow.

In the `fiber-wasm` scenario, cross-origin restrictions, especially on many mobile browsers, mean JoyID signing often has to be completed through a redirect flow. In practice, this can cause `fiber-wasm` to restart, which previously caused the runtime state created by `open_channel_with_external_funding` to be lost. As a result, the subsequent `submit_signed_funding_tx` call would fail.

## Changes

This PR adds persisted recovery state for the external funding flow so the following scenarios can continue correctly after a restart:

- `submit_signed_funding_tx` can still be executed after restart
- the post-submission channel handshake can resume after restart
- external funding timeout behavior still applies after restart
- persisted external funding recovery state is cleared once the flow reaches a terminal path

Specifically, this PR:

- adds a dedicated persisted recovery state for external funding
- restores external-funding-related runtime state from the store during reload / reestablish
- allows `submit_signed_funding_tx` to recover the missing channel state before continuing the flow
- updates the store schema and related store tests
- adds restart-focused tests and an e2e restart script for the external funding flow

## Test Coverage

This PR adds or updates coverage for:

- submitting `submit_signed_funding_tx` after restart
- resuming the external funding handshake after initiator restart
- resuming the external funding handshake after acceptor restart
- ensuring external funding timeout still aborts correctly after restart
- ensuring external funding recovery state is cleared after the channel reaches a terminal path
- adding an e2e test script at `tests/bruno/e2e/external-funding-open/run-restart-test.sh`

The restart behavior is added on top of the existing Bruno e2e flow. Since node restart is difficult to express cleanly in the Bruno workflow itself, this part is covered by a dedicated script.

## Notes

This PR is focused on restart recovery for the external funding flow and does not change RPC semantics.


quake:
Instead of introducing a new store prefix and ExternalFundingRecoveryState with separate get/insert/delete methods, we can add a single Option field to the existing ChannelActorData:

```rust
#[serde_as]
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct ExternalFundingPersistState {
    #[serde_as(as = "EntityHex")]
    pub funding_lock_script: Script,
    #[serde_as(as = "Vec<EntityHex>")]
    pub funding_lock_script_cell_deps: Vec<packed::CellDep>,
    #[serde_as(as = "EntityHex")]
    pub unsigned_funding_tx: Transaction,
    pub started_at_ms: u64,
    pub signed_submitted: bool,
}
```

Append to ChannelActorData:
```rust
#[serde(default)]
pub external_funding: Option<ExternalFundingPersistState>,
```

When an external funding flow begins, populate external_funding = Some(...) and persist via the normal insert_channel_actor_state path. On deserialization (restart/reload), rebuild ephemeral_config.external_funding from the persisted core.external_funding field. When the channel reaches ChannelReady or Closed, set it back to None.

This eliminates: No new store key prefix or schema change, no new ExternalFundingRecoveryState type with separate store operations, no new trait methods on ChannelActorStateStore (no mock/wasm changes needed). And the should_persist_channel_state special case can be removed, channel state is always persisted normally.

Migration is simple: append 1 byte (0u8, the bincode encoding of Option::None) to existing serialized ChannelActorState entries.

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -659,6 +686,19 @@ where
                     return Ok(());
                 }
 
+                if state.ephemeral_config.external_funding.enabled
+                    && state.ephemeral_config.external_funding.signed_submitted
```
chenyukang, line 701: In the external-funding path where should_local_send_tx_signatures_first() is true, the peer’s TxSignatures reply is routed into handle_tx_signatures(Some(...)), which falls through to SignFundingTx. do_sign_funding_tx() then replaces the stored externally-signed witnesses with the peer-provided list and re-signs with the node key. That drops the user’s external signature and breaks any external-funded channel whose signature order happens to put the initiator first.

chenyukang:
```console
External Wallet        Node A(channel)           Node A(network)         Chain Actor
     |                      |                          |                     |
     | sign unsigned tx     |                          |                     |
     |--------------------->|                          |                     |
     |                      | state.funding_tx =       |                     |
     |                      | externally signed tx     |                     |
     |                      |                          |                     |
     |                      | send TxSignatures        |                     |
     |                      |------------------------->|  send to peer       |
     |                      |                          |                     |
     |                      | receive peer reply       |                     |
     |                      | TxSignatures             |                     |
     |                      |                          |                     |
     |                      | handle_tx_signatures(    |                     |
     |                      |   Some(peer_witnesses))  |                     |
     |                      |                          |                     |
     |                      | falls through normal     |                     |
     |                      | SignFundingTx path       |                     |
     |                      |------------------------->|                     |
     |                      |                          | replace tx          |
     |                      |                          | witnesses with      |
     |                      |                          | peer_witnesses      |
     |                      |                          |-------------------->|
     |                      |                          |                     |
     |                      |                          | ask chain actor     |
     |                      |                          | to sign again       |
     |                      |                          |-------------------->|
     |                      |                          |                     |

```

chenyukang: There the `replace tx witnesses` should be something like merge ..

here is a possbile patch for your reference:

```diff
diff --git a/crates/fiber-lib/src/fiber/channel.rs b/crates/fiber-lib/src/fiber/channel.rs
index f2c64f0b1..39639e3c2 100644
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -695,7 +695,43 @@ where
                         .is_some()
                     && state.should_local_send_tx_signatures_first()
                 {
-                    state.handle_tx_signatures(Some(tx_signatures.witnesses))?;
+                    let Some(funding_tx) = state.funding_tx.clone() else {
+                        return Err(ProcessingChannelError::InvalidState(format!(
+                            "Received TxSignatures message, but the channel's funding tx is none {:?}",
+                            state.state
+                        )));
+                    };
+                    let new_witnesses: Vec<_> = tx_signatures
+                        .witnesses
+                        .into_iter()
+                        .map(|x| x.pack())
+                        .collect();
+                    debug!(
+                        "Merging external funding tx witnesses of {:?} with {:?}",
+                        funding_tx.calc_tx_hash(),
+                        new_witnesses.iter().map(|x| hex::encode(x.as_slice()))
+                    );
+                    let merged_witnesses =
+                        merge_external_funding_witnesses(funding_tx.witnesses(), new_witnesses);
+                    state.funding_tx = Some(
+                        funding_tx
+                            .as_advanced_builder()
+                            .set_witnesses(merged_witnesses)
+                            .build()
+                            .data(),
+                    );
+                    self.network
+                        .send_message(NetworkActorMessage::new_event(
+                            NetworkActorEvent::FundingTransactionPending(
+                                state.must_get_funding_transaction().clone(),
+                                state.must_get_funding_transaction_outpoint(),
+                                state.get_id(),
+                            ),
+                        ))
+                        .expect(ASSUME_NETWORK_ACTOR_ALIVE);
+                    state.update_state(ChannelState::AwaitingChannelReady(
+                        AwaitingChannelReadyFlags::empty(),
+                    ));
                     return Ok(());
                 }
 
diff --git a/crates/fiber-lib/src/fiber/tests/channel.rs b/crates/fiber-lib/src/fiber/tests/channel.rs
index efa12053c..662b1ea67 100644
--- a/crates/fiber-lib/src/fiber/tests/channel.rs
+++ b/crates/fiber-lib/src/fiber/tests/channel.rs
@@ -7432,6 +7432,17 @@ fn mock_sign_external_funding_tx(unsigned_tx: &Transaction) -> Transaction {
         .data()
 }
 
+fn mock_sign_external_funding_tx_with_witness(
+    unsigned_tx: &Transaction,
+    witness: Bytes,
+) -> Transaction {
+    unsigned_tx
+        .as_advanced_builder()
+        .set_witnesses(vec![witness])
+        .build()
+        .data()
+}
+
 #[test]
 fn test_external_funding_witness_merge_preserves_existing_signatures() {
     let external_signature: Bytes = [1u8; 65].pack();
@@ -8377,6 +8388,70 @@ async fn test_external_funding_initiator_restart_after_signed_submit_resumes_han
         .await;
 }
 
+#[tokio::test]
+async fn test_external_funding_initiator_send_tx_signatures_first_preserves_external_witness() {
+    init_tracing();
+
+    let [node_a, node_b] = NetworkNode::new_n_interconnected_nodes_with_config(2, |i| {
+        let mut builder = NetworkNodeConfigBuilder::new()
+            .node_name(Some(format!("ext-fund-send-first-node-{}", i)))
+            .base_dir_prefix(&format!("test-ext-fund-send-first-node-{}-", i));
+        if i == 1 {
+            builder = builder.fiber_config_updater(|config| {
+                config.auto_accept_channel_ckb_funding_amount = Some(150_000_000_000);
+            });
+        }
+        builder.build()
+    })
+    .await
+    .try_into()
+    .expect("2 nodes");
+
+    let (channel_id, unsigned_tx) =
+        open_external_funding_channel(&node_a, &node_b, 100_000_000_000).await;
+    let external_witness: Bytes = [9u8; 65].pack();
+    let signed_tx =
+        mock_sign_external_funding_tx_with_witness(&unsigned_tx, external_witness.clone());
+
+    let submit_message = |rpc_reply| {
+        NetworkActorMessage::Command(NetworkActorCommand::SubmitSignedFundingTx {
+            channel_id,
+            signed_tx: signed_tx.clone(),
+            reply: rpc_reply,
+        })
+    };
+    let submit_result = call!(node_a.network_actor, submit_message)
+        .expect("node_a alive")
+        .expect("submit signed funding tx success");
+
+    let expected_tx_hash: Hash256 = signed_tx.clone().into_view().hash().into();
+    assert_eq!(submit_result, expected_tx_hash);
+
+    let state = wait_for_external_funding_post_submit_progress(&node_a, channel_id).await;
+    assert!(
+        matches!(
+            state.state,
+            ChannelState::AwaitingChannelReady(_) | ChannelState::ChannelReady
+        ),
+        "initiator should complete tx_signatures handshake without re-signing, got {:?}",
+        state.state
+    );
+
+    let funding_tx = state
+        .funding_tx
+        .clone()
+        .expect("funding tx should remain available after tx_signatures");
+    assert_eq!(
+        funding_tx
+            .witnesses()
+            .get(0)
+            .expect("first witness exists")
+            .raw_data(),
+        external_witness.raw_data(),
+        "initiator should preserve the external witness when it sends tx_signatures first"
+    );
+}
+
 #[tokio::test]
 async fn test_external_funding_acceptor_restart_after_signed_submit_resumes_handshake() {
     init_tracing();

```

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -4961,6 +5043,94 @@ impl ChannelActorState {
...
+        self.ephemeral_config.external_funding = ExternalFundingRuntime {
+            enabled: true,
+            funding_lock_script: Some(external_funding.funding_lock_script),
+            funding_lock_script_cell_deps: external_funding.funding_lock_script_cell_deps,
+            unsigned_funding_tx: Some(external_funding.unsigned_funding_tx),
+            started_at: Some(UNIX_EPOCH + Duration::from_millis(external_funding.started_at_ms)),
+            signed_submitted: external_funding.signed_submitted,
+        };
+        if restore_signing_commitment {
+            self.state = ChannelState::SigningCommitment(
```
chenyukang, line 5517: One follow-up here: this restore path can move a pre-submit external-funding channel into `SigningCommitment(THEIR_COMMITMENT_SIGNED_SENT)` before the signed funding tx is ever submitted. After a restart, `post_start()` only re-arms the external-funding timeout for `state.state.is_awaiting_external_funding()`, so this branch falls back to the normal funding timeout (`created_at` + `funding_timeout_seconds`). `has_funding_timeout_elapsed()` would still treat the channel as waiting for external funding, but the check itself will now fire much later.

I think `post_start()` should key off `external_funding.enabled && !external_funding.signed_submitted` instead, e.g.

```rust
if state.ephemeral_config.external_funding.enabled
    && !state.ephemeral_config.external_funding.signed_submitted
{
    self.schedule_external_funding_timeout_check(&myself, state);
} else {
    self.schedule_timeout_event(
        &myself,
        state.created_at,
        state.ephemeral_config.funding_timeout_seconds,
    );
}
```

tests/bruno/e2e/external-funding-open/README.md
```diff
@@ -74,6 +74,47 @@ cd tests/bruno
 npm exec -- @usebruno/cli@1.20.0 run e2e/external-funding-open -r --env test
 ```
 
+### Restart regression script
+
+[`run-restart-test.sh`](tests/bruno/e2e/external-funding-open/run-restart-test.sh) is a restart regression test for the external funding flow. It specifically verifies that after node1 has already:
```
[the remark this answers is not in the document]
joii2020: I think using this path here would be more intuitive.

PR #1257, opened 2026-04-08
feat!(cch): CCH multi-asset swap

doitian:
## Summary

This PR delivers **CCH multi-asset swap**: a protocol specification plus the implementation that follows it. The cross-chain hub now swaps Lightning BTC against native CKB and allowlisted UDTs (not just a single wrapped-BTC UDT), and the operator-mediated proposal flow is **asynchronous / fire-and-forget**.

**BREAKING**: This PR contains database migration for CCH orders. Once merged and a new version is released, the migration should depend on the new version instead of the version in the workspace.

## Specification

`docs/specs/cch-multi-asset-swap.md`:

- Scope: HTLC flow unchanged; Fiber leg covers native CKB (shannons) and UDTs with type script; Lightning remains BTC-denominated.
- Economics: swap rate from invoice pair + hub fees (BTC-side fee basis); no hard-coded 1:1 wrapped-BTC mapping.
- Hub policy: asset allowlist and per-UDT decimals/metadata.
- Swap acceptor: WebSocket JSON-RPC subscription with binary accept/reject, configurable timeout, explicit RPC errors on reject/timeout.
- Async proposal model: client submits and polls; workflow resumes on the operator's response.

### Multi-asset CchOrder

- Generalized config and actor beyond a single wrapped-BTC UDT to native CKB and allowlisted UDTs (`cch/config.rs`).
- `CchOrder` reshaped (`fiber-types`, `fiber-json-types`): `fiber_type_script: Option<Script>`, separate BTC and Fiber amount fields (`btc_amount_msat`, `btc_fee_msat`, `fiber_amount_smallest_unit`). The counterparty invoice (`incoming_invoice: CchInvoice`) is always present — the order is only created once the counterparty leg has been priced and its invoice minted.
- `CchPendingProposal` is a **separate store record** (not a `CchOrderStatus` variant). It holds the `SwapProposal` plus the scaffolding needed to build a `CchOrder` when the operator accepts. No `CchOrder` exists during the proposal phase.

### Async / fire-and-forget proposal flow

- `send_btc` / `receive_btc` validate, persist a `CchPendingProposal` record, broadcast the `SwapProposal` to operator subscribers, and **return immediately** with a `NewOrderResult::PendingProposal(SwapProposal)`. The client polls `get_cch_order`.
- The workflow **resumes** when the operator calls `submit_swap_proposal_response`: the actor mints the counterparty invoice, **creates** a new `CchOrder` with status `Pending`, deletes the `CchPendingProposal` record, and starts the action flow. Reject/timeout **deletes** the proposal record — no `CchOrder` is ever created.
- On restart, every persisted `CchPendingProposal` record is re-broadcast and its timeout rescheduled to its deadline. Orders are resumed as usual via `ResumeOrder`.
- Removed the `Arc<Mutex<HashSet<Hash256>>>` inflight-hash mutex (the tokio shared-state antipattern). All order creation runs inline in the single-threaded actor mailbox, which naturally serializes same-hash requests. The swap acceptor is reduced to a pure pub/sub broadcaster; blocking `call_t!` is replaced with fire-and-forget message passing.
- Fiber-leg outgoing payment fee budgets are converted from the BTC-denominated fee to the Fiber asset's smallest unit using the order's **net** exchange rate (excluding the hub fee), so the invariant "outgoing route fee never exceeds collected fee" holds regardless of which leg carries the outgoing payment.

### RPC

- `cch.rs`: async/poll doc semantics; `submit_swap_proposal_response` is now the resume path; reverted the inflated proposal-timeout mailbox window.
- Regenerated `rpc/README.md`.

### Migration

- `crates/fiber-store/src/migrations/mig_20260610_cch_multi_asset.rs`: migrates legacy single-asset `CchOrder` to the multi-asset shape in the **new** migration system (auto-registered via `build.rs`). Legacy orders are never `PendingProposal` (that variant doesn't exist in `CchOrderStatus`).

### Tests / docs

- Reworked `cch` actor/dispatcher/scheduler/state-machine tests to the async model (immediate `PendingProposal` return, operator response, poll for resolution, timeout → deleted, restart re-broadcast).
- Regenerated `store/.schema.json`.

## Validation

- `cargo build -p fnn -p fiber-bin --features rocksdb` ✓
- `cargo check --target wasm32-unknown-unknown -p fiber-store` ✓
- `fiber-store` tests, `cch` tests, `make check-migrate`, fmt/clippy/typos ✓


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `3bbf5ea0ed7debd83a707b5f28264bee2fd7371f..528c50a074a4ecde3df5335daccaa685d93ecb07`

Validation:
- Performed a diff-focused review of the CCH multi-asset flow, RPC authorization, CCH migration, watchtower change, and in-flight CKB transaction handling.

Notes:
- `crates/fiber-lib/src/fiber/in_flight_ckb_tx_actor.rs` now treats CKB RPC `-301 TransactionFailedToResolve` as a permanent rejection and reports `ReportRejected` without first checking whether the tx is already committed. The previous guard documented that rebroadcasting an already-committed tx can return `-301` because its inputs are spent. For funding txs, this can flow to `FundingTransactionFailed` and `abort_funding`, so an already-committed funding tx can be locally abandoned before the committed trace is processed.
- Suggested fix: before converting `-301` into rejection, query `get_transaction(tx_hash)` and defer to the tracer or report committed status when the tx is already committed.

chainTe:
## Security review

Reviewed `9bffae4bfa05fbf70c061626bad14da411861855..8f6064a118cd201e03861f009fc91172e91302c1` and found one Medium issue:

**The CCH migration reconstructs legacy invoice amounts without considering swap direction.** In `crates/fiber-store/src/migrations/mig_20260610_cch_multi_asset.rs:55`, `lightning_invoice_amount` is always set to `old.amount_sats * 1000`. For a legacy `ReceiveBTC` order, however, `amount_sats` is the fee-exclusive outgoing Fiber principal, while the stored incoming Lightning hold invoice is fee-inclusive. When the order resumes, `outgoing_fee_budget_fiber_smallest_unit` subtracts `btc_fee_msat` again, which can inflate the Fiber routing-fee budget beyond the value of the fee collected for that order.

Suggested fix: branch on the legacy `incoming_invoice` variant and derive each leg's exact amount from its encoded invoice. Add nonzero-fee migration tests for both `SendBTC` and `ReceiveBTC`, asserting that `lightning_invoice_amount - btc_fee_msat` is the fee-exclusive BTC principal and that the resumed outgoing fee budget cannot exceed the converted collected fee.

Validation performed:

- SQLite-backed CCH migration tests: 2 passed.
- Focused outgoing-fee-budget tests: 9 passed.
- RocksDB-backed migration compilation was blocked by the runner's missing `stdbool.h` header.
- The full existing test suite was not run locally.

PR #1261, opened 2026-04-08
fix: push limit to DB layer in list_payments to avoid unbounded memory usage

quake:
get_payment_sessions_with_limit previously called collect_by_prefix without a limit, loading ALL payment session KV pairs into memory before applying .filter_map().take(limit) as iterator adapters.

Rewrite get_payment_sessions_with_limit as a simple iterator chain using the lazy batched PrefixIterator, which fetches only 100 entries at a time and stops as soon as enough results are collected.


gpBlockchain:
lgtm

PR #1263, opened 2026-04-09
 Add gossip ban and rate limit policy

jjyr:
This PR require #1217 

  ## Summary

  This PR adds a gossip policy that temporarily bans misbehaving non-channel peers and rate-
  limits gossip traffic to reduce spam, protect sync, and limit abuse.

  ## Changes

  - add `GossipPolicyConfig` under `FiberConfig` as `gossip_policy`
  - add in-memory gossip scoring and 10-minute temporary bans for non-channel peers
  - add outbound gossip rate limiting at both global and per-peer levels
  - add inbound channel update rate limiting
  - add delayed outbound gossip delivery with per-peer queues
  - add tests for ban tracking, rate limiting, delayed sends, and related gossip flows

[View diff](https://github.com/nervosnetwork/fiber/pull/1263/changes/bb976fa84a3f1627bd0084c2d30bad3dd364123a)


crates/fiber-lib/src/fiber/gossip.rs
```diff
@@ -2250,17 +2751,20 @@ async fn get_channel_tx(
...
+            tx_status: TxStatus::Committed(_, block_hash, _),
         }) => Ok((tx, block_hash.into())),
-        Err(err) => Err(Error::InvalidParameter(format!(
-            "Channel announcement transaction {:?} not found or not confirmed, result is: {:?}",
-            &outpoint.tx_hash(),
-            err
-        ))),
-        _ => Err(Error::InvalidParameter(format!(
-            "Channel announcement transaction {:?} not found or not confirmed, the reason is unknown",
+        Ok(response) => Err(Error::InvalidParameter(format!(
```
chenyukang: CkbRpcClient::get_transaction returns a successful response with tx_status Unknown, Pending, or Proposed when the funding tx is not committed yet. This branch still maps those statuses to InvalidParameter, so an honest peer can be scored or banned whenever our local chain view lags behind theirs. The defer path only handles transport errors; non-committed statuses should be deferred too.

crates/fiber-lib/src/fiber/gossip.rs
```diff
@@ -1751,10 +2010,23 @@ impl<S: GossipMessageStore + Send + Sync + 'static, C: CkbChainClient + Send + S
             }
 
             ExtendedGossipMessageStoreMessage::SaveMessages(peer, messages) => {
+                let now_ms = now_timestamp_as_millis_u64();
                 for message in messages {
```
chenyukang, line 2099: for BroadcastMessage like ChannelUpdate, seems save latest message is more meanningful than old ones.
here when we reached a limit, we drop those new ones?
maybe it's not a big issue, since we add limit here for extreme scenarios....

crates/fiber-lib/src/fiber/config.rs
```diff
@@ -258,6 +259,15 @@ pub struct FiberConfig {
     )]
     pub(crate) min_outbound_peers: Option<usize>,
 
+    /// Whether to automatically schedule reconnect backoff for peers with active channels. [default: true]
+    #[arg(
+        name = "FIBER_ENABLE_PEER_RECONNECT_BACKOFF",
+        long = "fiber-enable-peer-reconnect-backoff",
```
chenyukang, line 267: add this to the sample config in `testnet/config.yml` and `mainnet/config.yml`, but keep in commented?

jjyr: ok

please note the part of change is belongs to #1217, I will update in #1217

Please review this PR through the link in the PR description

crates/fiber-lib/src/fiber/gossip.rs
jjyr: Right, it is for preventing the milicious attack, the regular gossip message shouldn't reach the limitation.

quake (approved):

chainTe:
## Security Review

Reviewed range: `2d89d7647ca3d0bf1e8d546aa2da16636beb9662..d8f466175fb02db352a8ee573640b39d02d7903c`

Validation:
- Reviewed the gossip ban/rate-limit policy, inbound channel-update throttling, outbound delayed response queue, deferred channel-announcement verification, and active-channel-aware ban disconnect behavior.
- Report format validation and HTML rendering passed for the private review artifacts.

Notes:
- No source changes were made.

chainTe:
## Security Review

Result: 1 confirmed High availability finding after correction. The delayed outbound gossip queue introduced by this PR is item-count capped but not byte-capped while retaining full gossip response payloads, so remote peers can force large delayed `GetBroadcastMessagesResult` / `QueryBroadcastMessagesResult` payloads to accumulate in memory.

Reviewed range: `2d89d7647ca3d0bf1e8d546aa2da16636beb9662..d8f466175fb02db352a8ee573640b39d02d7903c`

Validation:
- Static source-to-sink re-review confirmed reachability from remote Get/Query gossip request handlers to delayed full-response retention.

Notes:
- Recommended fix: add per-peer and global queued-byte budgets before storing delayed outbound payloads, decrement those budgets on send/drop/disconnect, and add inbound request backpressure for large Get/Query response paths.

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `2d89d7647ca3d0bf1e8d546aa2da16636beb9662..d8f466175fb02db352a8ee573640b39d02d7903c`

Validation:
- Static review traced remote gossip ingress through the new inbound channel-update limiter and pending-message insertion path.

Notes:
- The new inbound channel-update limiter is keyed by `(peer, channel_outpoint, is_node1)`, but a remote peer controls `channel_outpoint`; once the tracked-key cap is reached, the limiter evicts an old key and admits the new one. A peer can therefore send unique bogus outpoints and continue feeding pending gossip work despite the intended rate limit. Consider rejecting new keys when the limiter is full or adding per-peer/global pending-message bounds, and count limiter drops as policy violations so sustained abuse reaches the ban path.
- The delayed outbound queue is item-capped per peer but not byte-capped; adding a byte budget would make the resource bound easier to reason about.

jjyr:
@chainTe 

Fixed in https://github.com/nervosnetwork/fiber/pull/1485

PR #1265, opened 2026-04-10
Fix gitignore

chenyukang:
`.agents` is openx's directory to store local skills

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ target
 baseline.json
 current.json
 comparison_report.txt
+.agents
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ members = [
     "crates/fiber-json-types",
     "crates/fiber-cli",
 ]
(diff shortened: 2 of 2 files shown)


quake:
each coding agent has the convention to load local skills, for example:

```
opencode
Project config: .opencode/skills/<name>/SKILL.md

claude
Project-Local: .claude/skills/<skill-name>/ or .agents/skills/<skill-name>/
```

I don't think add extra prompt in the AGENT.md is a good idea which will increase the token usage.

chenyukang:
> each coding agent has the convention to load local skills, for example:
> 
> ```
> opencode
> Project config: .opencode/skills/<name>/SKILL.md
> 
> claude
> Project-Local: .claude/skills/<skill-name>/ or .agents/skills/<skill-name>/
> ```
> 
> I don't think add extra prompt in the AGENT.md is a good idea which will increase the token usage.

update the change to only add a directory to gitignore.

chenyukang:
I will add it into `.git/info/exclude`.

PR #1270, opened 2026-04-13
feat: add addr_type parameter to connect_peer RPC for transport type filtering

quake:
When connect_peer is called with only a pubkey, the node randomly selects an address from the peer's known addresses. In WASM environments, this may select a non-WSS address that is unsupported. Add an optional addr_type parameter (tcp/ws/wss) to allow callers to filter addresses by transport type before random selection.

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -24,8 +24,8 @@ use std::sync::Arc;
 use strum::AsRefStr;
 use tentacle::multiaddr::{MultiAddr, Protocol};
 use tentacle::service::SessionType;
+use tentacle::utils::extract_peer_id;
 use tentacle::utils::TransportType;
-use tentacle::utils::{extract_peer_id, is_reachable, multiaddr_to_socketaddr};
 use tentacle::{
     async_trait,
     builder::{MetaBuilder, ServiceBuilder},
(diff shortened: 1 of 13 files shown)


crates/fiber-lib/src/fiber/network.rs
```diff
@@ -1447,11 +1452,16 @@ where
...
+                    addresses
+                        .into_iter()
+                        .filter(|addr| matches_addr_type(addr, addr_type))
+                        .choose(&mut rand::thread_rng())
+                } else {
+                    addresses.into_iter().choose(&mut rand::thread_rng())
+                };
                 let Some(addr) = address else {
                     let _ = reply.send(Err(Error::PeerNotFound(pubkey).to_string()));
                     return Ok(());
```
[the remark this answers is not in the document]
chenyukang: good catch here

chenyukang (approved):

crates/fiber-lib/src/utils/mod.rs
```diff
@@ -2,3 +2,27 @@ pub mod actor;
...
+use tentacle::utils::{is_reachable, multiaddr_to_socketaddr};
+
+/// Check whether a multiaddr is publicly reachable.
+///
+/// For IP-based addresses (`Ip4`/`Ip6`), this delegates to tentacle's
+/// `multiaddr_to_socketaddr` + `is_reachable` check.
+///
+/// For DNS-based addresses (`Dns4`/`Dns6`), we treat them as always reachable
+/// because a DNS name implies a publicly resolvable endpoint.
+pub(crate) fn is_addr_reachable(addr: &Multiaddr) -> bool {
```
chenyukang, line 16: it there a possibility that private hostnames and malformed DNS multiaddrs can now propagate

jjyr (approved):

crates/fiber-lib/src/utils/mod.rs
quake: it's possible, but it is normal all gossip messages, an unreachable ip4/ipv6 address can be broadast also.

PR #1277, opened 2026-04-16
Fix: enhance payment routing with wait_until_route_available

chenyukang:
Fixes https://github.com/nervosnetwork/fiber/actions/runs/24484064507/job/71554890264

The benchmark workflow was running tests/perf with FIBER_GOSSIP_NETWORK_MAINTENANCE_INTERVAL_MS=60000 and FIBER_GOSSIP_STORE_MAINTENANCE_INTERVAL_MS=20000, but the integration warm-up sent a keysend payment immediately after channel setup. Under slower gossip
  maintenance, the route graph was not yet converged, so the warm-up failed with `PathFind error: no path found`.

This change makes the warm-up wait for route discovery with dry_run polling before sending the real payment, and then verifies the payment reaches Success. This keeps the benchmark behavior unchanged while removing the race that caused release-time benchmark CI failures.

cc @swananan


jjyr (approved):

swananan:
Thanks for the fix, I missed that impact of the env changes - dry_run polling looks clean. And I will work on the next step of gossip optimzation this weekend.

PR #1280, opened 2026-04-16
Fix docker image building slow issues

chenyukang:
Docker image building is too slow because of `qemu`.
we should build on native.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -184,6 +184,7 @@ jobs:
           ghcr.io/${{ github.repository }}
           nervos/fiber
         flavor: |
+          # Keep auto latest disabled; the explicit latest tag below is only for stable releases.
           latest=false
         tags: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -178,22 +186,71 @@ jobs:
...
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+    - name: Extract Docker metadata
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: |
+          ghcr.io/${{ github.repository }}
+          nervos/fiber
+        flavor: |
+          latest=false
```
eval-exec, line 236: latest should be true?

eval-exec (approved):

.github/workflows/release.yml
chenyukang: we keep `latest=false` intentionally to disable metadata-action's automatic latest tag. The workflow adds `latest` explicitly below only for stable release tags, so prerelease tags like `vX.Y.Z-rcN` do not update `latest`.

PR #1282, opened 2026-04-16
Fiber WSS Configuration Manual

jiangxianliang007:
This PR demonstrates how to connect a Fiber node via WASM.

### DIFF
--- a/docs/fiber-node-wss.md
+++ b/docs/fiber-node-wss.md
@@ -158,21 +158,56 @@ http {
 
 Add an A record for the domain fiber.example.com to your Fiber node's IP address in your domain registrar.
 
-## Get Your node's addresses
 
+## Connect to Your Node via WSS
+
+**Option A — Connect by explicit WSS address:**
 Prerequisites:
 
-- Fiber node config.yml announced_addrs add public IP
-- Your node's RPC port is 8227
+- In your Fiber node's `config.yml`, add the WSS multiaddr to `announced_addrs` so that
+  browser/WASM clients can discover and connect to your node via the P2P graph:
 
+```yaml
+announced_addrs:
+    - /ip4/<your-public-ip>/tcp/8228
+    - /dns4/fiber.example.com/tcp/443/wss
 ```
-curl -s -X POST http://127.0.0.1:8227 -H "Content-Type: application/json" -d '{
+
+
+```sh
+curl -s -X POST http://127.0.0.1:8227 \
+  -H "Content-Type: application/json" \
+  -d '{
     "id": 42,
     "jsonrpc": "2.0",
-    "method": "node_info",
-    "params": []
-}' |jq '.result.addresses[0]' |awk -F '/p2p/' '{print $2}'
+    "method": "connect_peer",
+    "params": [{ "address": "/dns4/fiber.example.com/tcp/443/wss" }]
+  }'
 ```
 
-## Connect via the WSS address
-/dns4/{Your-Domain-Name}/tcp/443/wss/p2p/{Your-Fiber-Node-Addresses}
+**Option B — Connect by pubkey (requires WSS address to be propagated via gossip):**
+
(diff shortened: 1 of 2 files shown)


docs/fiber-node-wss.md
```diff
@@ -0,0 +1,178 @@
...
+- Fiber node config.yml announced_addrs add public IP
+- Your node's RPC port is 8227
+
+```
+curl -s -X POST http://127.0.0.1:8227 -H "Content-Type: application/json" -d '{
+    "id": 42,
+    "jsonrpc": "2.0",
+    "method": "node_info",
+    "params": []
+}' |jq '.result.addresses[0]' |awk -F '/p2p/' '{print $2}'
```
chenyukang: This should use `jq -r`; otherwise the extracted peer id can include the JSON quotes. It also assumes `addresses[0]` is the right `/p2p/` address. A safer version would filter for an address containing `/p2p/`, for example:

```sh
jq -r '.result.addresses[] | select(contains("/p2p/")) | split("/p2p/")[1]'
```

docs/fiber-node-wss.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Add Domain DNS Resolution
+
+Add an A record for the domain fiber.example.com to your Fiber node's IP address in your domain registrar.
+
+## Get Your node's addresses
+
+Prerequisites:
+
+- Fiber node config.yml announced_addrs add public IP
```
chenyukang: For browser/WASM clients, adding only a public TCP/IP address is not enough if the client resolves peers from graph data. The node should advertise a browser-usable WSS multiaddr, e.g. `/dns4/fiber.example.com/tcp/443/wss`, or this section should make clear that the WSS address is only for explicit `connect_peer(address=...)` usage.

chenyukang (commented):
Reviewed the WSS configuration guide. I left a few inline comments around the parts that could mislead users when they try to connect from WASM/browser environments.

docs/fiber-node-wss.md
```diff
@@ -0,0 +1,178 @@
...
+    server {
+        listen 8443 ssl;
+        server_name fiber.example.com;
+
+        ssl_certificate /usr/local/nginx/ssl/fullchain.pem;
+        ssl_certificate_key /usr/local/nginx/ssl/privkey.pem;
+        ssl_protocols TLSv1.2 TLSv1.3;
+
+        location / {
+            proxy_pass http://127.0.0.1:8228;
```
chenyukang, line 138: This proxy target relies on Fiber accepting WebSocket traffic on the P2P port. Since that depends on `fiber.reuse_port_for_websocket` staying enabled, the guide should mention that requirement near the prerequisites or this proxy example.

docs/fiber-node-wss.md
```diff
@@ -0,0 +1,178 @@
...
+curl -s -X POST http://127.0.0.1:8227 -H "Content-Type: application/json" -d '{
+    "id": 42,
+    "jsonrpc": "2.0",
+    "method": "node_info",
+    "params": []
+}' |jq '.result.addresses[0]' |awk -F '/p2p/' '{print $2}'
+```
+
+## Connect via the WSS address
+/dns4/{Your-Domain-Name}/tcp/443/wss/p2p/{Your-Fiber-Node-Addresses}
```
chenyukang: This gives the final multiaddr, but it would be more actionable to show how to use it, for example with `connect_peer` using the `address` parameter. If the intended flow is pubkey-based connection, please also mention `addr_type: "wss"` and the need for graph data to contain a WSS address.

docs/fiber-node-wss.md
jiangxianliang007: Fiber listens for WebSocket connections on the same TCP port as P2P by default, The rest has been updated, and a pubkey connect_peer has been added.

PR #1289, opened 2026-04-20
Enhance hold TLC settlement handling for received invoices

### LINKED ISSUE #1161: Flaky test: test_trampoline_routing_race_same_invoice
# Flaky test: `test_trampoline_routing_race_same_invoice`

## Summary

The test `fiber::tests::trampoline::test_trampoline_routing_race_same_invoice` is flaky. It sometimes fails with both payments succeeding when the invariant is that exactly one must succeed and one must fail.

## Test behavior

- **Setup:** Four nodes A, T (trampoline), B, C. A and B both pay the **same** invoice (created by C) via T concurrently.
- **Expected:** Exactly one of the two payments succeeds; the other fails (e.g. invoice already taken).
- **Observed (when flaky):** Both payments can succeed, so the assertion `assert_eq!(success, 1)` fails with `success == 2`.

## Root cause

When two payers race on the same invoice, the payee C can receive two TLCs (one from each path). Settlement is triggered by `SettleHoldTlcSet(payment_hash)`. If the two TLCs commit in **separate** channel updates, C sends **two** `SettleHoldTlcSet(payment_hash)` messages:

1. First run: one TLC, invoice is Open → fulfill that TLC, set invoice to **Received**.
2. Second run: the other TLC, invoice is **Received**. In `SettleTlcSetCommand`, the hold path explicitly allows `Received` (so that hold invoices can settle when the preimage arrives after the invoice was already marked Received). So the second run also calls `try_settle_all()` and fulfills the second TLC → both payments succeed.

So the flakiness comes from allowing fulfillment when the invoice is already Received in the hold path.

## Why we don't fix it by rejecting when Received

Rejecting all TLCs in `SettleTlcSetCommand` when `invoice_status == Received` would stabilize this test but would **break hold invoices**. For hold invoices, the preimage often becomes available only **after** the invoice has been marked Received (e.g. from a watchtower or external system). We must allow settlement when the invoice is already Received in that case. So we cannot add a blanket "reject when Received" rule in `SettleTlcSetCommand`.

## Mitigation: add nextest retries (optional)

Because the failure is timing-dependent, re-running the test often yields a pass (only one run "wins" the race). You can reduce CI flakiness by adding a nextest override that retries this test.

**How to add it:** In `.config/nextest.toml`, append:

```toml
# Flaky race test: two senders paying same invoice; one must succeed, one must fail.
# Retries allow a passing run when scheduling varies.
[[profile.default.overrides]]
filter = 'test(test_trampoline_routing_race_same_invoice)'
retries = { backoff = "fixed", count = 3, delay = "1s" }
```

**Why it helps:** The test only fails when both payments succeed. That happens only when two `SettleHoldTlcSet` runs are processed in an order that allows both TLCs to be fulfilled. Retrying a few times with a short delay makes it likely that at least one run will hit the intended outcome (one success, one failure), so the suite passes without changing protocol behavior or breaking hold invoices.

## Possible directions for a proper fix (not implemented)

- Serialize settlement per `payment_hash` (e.g. a lock or single-actor processing) so only one `SettleHoldTlcSet` run can fulfill for a given invoice.
- Or make the store transition Open → Received atomic (e.g. compare-and-swap) and have a second run see Received and reject only when it would be a true "second payer" case, without breaking the hold-invoice "preimage after Received" flow.

## Location

- Test: `crates/fiber-lib/src/fiber/tests/trampoline.rs` — `test_trampoline_routing_race_same_invoice`
- Settlement: `crates/fiber-lib/src/fiber/settle_tlc_set_command.rs` (hold path allows `Received` in `verify_invoice_status`)

chenyukang:
Fixes #1161

Fix flaky `test_trampoline_routing_race_same_invoice` .

- Add a dedicated `SettleReceivedHoldTlcSet` path for hold invoices that are already `Received` and later get a preimage.
- Keep `SettleHoldTlcSet` for newly-arrived hold TLCs, and reject them when the invoice is already `Received` to avoid duplicate same-invoice fulfillment.
- Route invoice TLCs found by `CheckChannels` through `SettleTlcSetCommand` instead of directly fulfilling them with a known preimage.
- Add unit coverage for both duplicate `Received` hold TLC rejection and valid post-preimage hold invoice settlement.

For reviwer:
The first commit is bugfix, the second commit is code refactoring.


chenyukang:
cc @doitian

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2573,6 +2437,221 @@ where
...
+
+    async fn handle_committed_received_tlcs(
+        &self,
+        state: &mut NetworkActorState<S, C>,
+        channel_id: Hash256,
+        committed_tlcs: Vec<(TLCId, u64, Hash256, bool)>,
+        invoice_tlc_sets: &mut HashMap<Hash256, Vec<(Hash256, u64)>>,
+    ) {
+        for (_tlc_id, id, payment_hash, is_waiting_forward_result) in committed_tlcs {
+            if self.store.get_invoice(&payment_hash).is_some() {
```
jjyr: In the new `handle_committed_received_tlcs` function, we completely skipped the `is_invoice_fulfilled` check.

chenyukang: The fulfilled check was moved into SettleTlcSetCommand. CheckChannels now only groups open invoice TLCs by payment_hash and passes them to settle_tlc_set; `SettleTlcSetCommand::leave_just_fulfilled_tlcs` performs the amount/MPP fulfillment check before creating fulfill settlements. 

This avoids the old one-by-one direct fulfill path and lets MPP be evaluated as a set.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -2138,32 +2021,13 @@ where
...
-                                ))
-                                .expect(ASSUME_NETWORK_MYSELF_ALIVE);
-                        }
-                    }
-                }
+                self.retry_hold_tlc_sets(&myself);
             }
             NetworkActorCommand::SettleHoldTlcSet(payment_hash) => {
-                self.settle_hold_tlc_set(state, payment_hash).await;
+                self.settle_hold_tlc_set(state, payment_hash, false).await;
```
doitian: Bool parameter is hard to read. It's better to create two versions of the functions instead.

doitian (approved):

doitian (approved):

Issue #1256, opened 2026-04-08
Unified Migration System Design


quake:
# Unified Migration System Design

**Date**: 2026-04-08
**Status**: Draft
**Scope**: fiber-store migration framework, cross-platform (native + WASM)

---

## 1. Background & Motivation

The store layer has been unified behind the `StorageBackend` trait with four backend implementations (RocksDB, SQLite, IndexedDB, BTreeMap). However, the migration system has not kept pace:

- **Native-only**: The current `migrate/` crate is a standalone workspace that only supports RocksDB.
- **Historical baggage**: Migrations depend on old git branches (`migrate/v0.6.0`, `v0.6.1`, `v0.7.0`) to deserialize legacy data formats.
- **No WASM migration**: The browser/IndexedDB backend has zero migration support.
- **Fragmented code**: Active migrations live in `migrate/src/migrations/`, archived ones in `migrate/src/migrations_backup/`, and the migration framework itself is split between `fiber-store` and `migrate/`.

Since the store is now unified, migrations can be written purely against `StorageBackend`, making cross-platform migration both possible and straightforward.

---

## 2. Design Goals

1. **Archive all existing migrations** — start a new epoch, sever dependency on historical git branches.
2. **Unified migration crate** — migration code lives in `fiber-store`, compiles to both native and WASM.
3. **Auto-migrate at startup** — with user confirmation and backup reminder, works on both platforms.
4. **Preserve breaking change mechanism** — breaking migrations require explicit user confirmation.
5. **Schema check unchanged** — `migration-check` + `.schema.json` continues to guard type changes in CI.

---

### 3.1 New Directory Structure

```
fiber/
├── migrate_archive/                # Archived: v0.7.0 and earlier migrations (reference only)
│   ├── README.md                   # Explains this is archived, not compiled
│   ├── src/migrations/             # Former active migrations
│   └── src/migrations_backup/      # Even older migrations
├── crates/
│   └── fiber-store/
│       ├── build.rs                # Scans src/migrations/ for LATEST_DB_VERSION
│       └── src/
│           ├── backend.rs          # StorageBackend trait (unchanged)
│           ├── migration.rs        # Migration trait + Migrations registry (refactored)
│           ├── db_migrate.rs       # DbMigrate coordinator (refactored)
│           ├── migrations/         # New: all migration implementations
│           │   ├── mod.rs          # Auto-registers all migrations
│           │   └── mig_YYYYMMDD_xxx.rs
[...]

PR #1323, opened 2026-05-05
feat: Unified Migration System Design

quake:
Impl of https://github.com/nervosnetwork/fiber/issues/1256


quake:
@copilot apply changes based on the comments in [this thread](https://github.com/nervosnetwork/fiber/pull/1323#pullrequestreview-4225165560)

crates/fiber-store/src/migration.rs
```diff
@@ -32,205 +123,156 @@ impl Migrations {
...
-                pb
-            };
-            db = m.migrate(db, Arc::new(pb))?;
-            db.put(MIGRATION_VERSION_KEY, m.version());
-        }
-        mpb.clear().expect("MultiProgress join");
-        Ok(db)
+            .filter(|(v, _)| v.as_str() > current_version)
+            .map(|(_, m)| m)
+            .collect()
```
Zhangcy0x3, line 191: ### (Nit) Suggestion: Use .range() for better efficiency 
I noticed that pending_migrations is currently using `.iter().filter()`. Since we are utilizing a `BTreeMap `which is already ordered by the version string, we could leverage the `.range()` method here.
Using `.range((Bound::Excluded(current_version.to_string()), Bound::Unbounded))` would allow us to locate the starting point in $O(\log n)$ time instead of $O(n)$. While the current number of migrations is small, using `.range()` is more idiomatic for ordered containers and improves the clarity of our intent to fetch a specific version subset.
```suggestion
        self.migrations
            .range((std::ops::Bound::Excluded(current_version.to_string()), std::ops::Bound::Unbounded))
            .map(|(_, m)| m)
            .collect()
```

crates/fiber-store/src/migration.rs
```diff
@@ -32,205 +123,156 @@ impl Migrations {
...
-                String::from_utf8(version_bytes.to_vec()).expect("version bytes to utf8")
-            }
-            None => {
-                return Ordering::Less;
-            }
-        };
+    pub fn get_db_version(&self, store: &Store) -> Option<String> {
+        store
+            .get(MIGRATION_VERSION_KEY)
+            .map(|v| String::from_utf8(v).expect("version bytes to utf8"))
```
Zhangcy0x3: ### (Question) Error Handling Strategy: Panic vs. Result in WASM

Regarding the use of `.expect` here: while version corruption is indeed a fatal error, there's a technical nuance to consider for our WASM/WebWorker implementation.

A `panic` (triggered by `expect`) will lead to a WASM trap that bypasses the external `.map_err()` logic in startup flow. This means the intended `ExitMessage `might not be gracefully propagated to the JavaScript side.

To stay consistent with the `MigrateError `flow defined in the design document, should we return a `Result `instead? This would allow the error to be captured and handled by our `exit_to_js `logic.

Zhangcy0x3 (commented):
I've left a couple of technical suggestions regarding WASM error handling and some minor API improvements for BTreeMap. Please take a look when you have time.

migrate_archive/src/migrations_backup/mig_20250114.rs
chenyukang, line 1: This PR moves migrate/ to migrate_archive/, but the release workflow still cd migrate, builds it, and packages migrate/target/release/fnn-migrate. Release builds will fail immediately because migrate/Cargo.toml no longer exists.

crates/fiber-store/src/migration.rs
```diff
@@ -1,26 +1,117 @@
...
+pub type MigrateProgressFn = Box<dyn Fn(MigrationProgress)>;
+
+// --- Migration trait ---
+
+pub trait Migration: Send + Sync {
+    /// Timestamp-based version string: "YYYYMMDDHHMMSS"
+    fn version(&self) -> &str;
+
+    /// Execute migration using the store backend.
+    fn migrate(&self, store: &Store) -> Result<(), String>;
```
chenyukang: here should be `StorageBackend`?

chenyukang: I have a try following this way.

https://github.com/quake/fiber/pull/10

quake: resolved in https://github.com/nervosnetwork/fiber/pull/1323/commits/5f9ebc237d0f0f1376d7c1a29064610f446979f1 by introcuded a new trait MigrationStore

docs/plans/2026-04-09-unified-migration-system.md
```diff
@@ -0,0 +1,1069 @@
+# Unified Migration System — Implementation Plan
+
+> **For Claude:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task.
```
jjyr: Should we keep the plan doc in the project?

jjyr (approved):

docs/plans/2026-04-09-unified-migration-system.md
chenyukang: I think it's better to remove or convert it to dev notes.

PR #1335, opened 2026-05-09
Fix preimage retention for MPP during force-close scenarios

### LINKED ISSUE #1332: MPP force close hangs when only one split TLC is on-chain
commit: https://github.com/nervosnetwork/fiber/commit/0aa8665169f0e95caf62e04bd43f467264ce1cf6
## Title

MPP force close hangs when only one split TLC is on-chain

## Summary

MPP payment is split across two A-B channels. If only one A-B channel is force-closed, then payee calls `settle_invoice`, the test keeps waiting for commit cells to be consumed.

Flow:

1. Open two A-B channels and one B-C channel.
2. Send one MPP payment from A to C with `max_parts = 2`.
3. Wait payment `Inflight`, invoice `Received`.
4. Force-close only one A-B channel.
5. C calls `settle_invoice`.
6. Off-chain TLC settled successfully, while on-chain TLC remains unsettled permanently

## Expected

The force-closed TLC should be settled/consumed, and the payment should reach a final state.

## Actual

Commit cells are not fully consumed. The test hangs waiting for on-chain settlement.

chenyukang:
Fixes #1332

cc @15168316096


15168316096:
> Fixes #1332
> 
> cc @15168316096

@chenyukang The bugfix was validated and the behavior met PR expectations.

Xcodes-chain:
There are still some issues with the on-chain part of the case; let's double-check them. @chenyukang

chenyukang:
> @15168316096 wrote:
> There are still some issues with the on-chain part of the case; let's double-check them. @chenyukang

it's caused by the latest commit?

Xcodes-chain:
> > @15168316096 wrote:
> > There are still some issues with the on-chain part of the case; let's double-check them. @chenyukang
> 
> it's caused by the latest commit?

@chenyukang The issue is still being investigated; a new case was found to have failed to run.
Let's take a look at the latest committed FNN binary.

gpBlockchain:
lgtm

Xcodes-chain:
@chenyukang The FNN test corresponding to the new local regression PR passed.

PR #1378, opened 2026-05-26
Fix hop hint path capacity args

jjyr:
- Fix `inner_find_path` hop hint handling so `eval_and_update` receives `channel_capacity` and
`tlc_min_value` in the correct order.
- Add a regression test for private hop hints to ensure hinted private channels remain routable.

### DIFF
--- a/crates/fiber-lib/src/fiber/graph.rs
+++ b/crates/fiber-lib/src/fiber/graph.rs
@@ -2181,8 +2181,8 @@ where
                 let tlc_min_val = 0;
                 self.eval_and_update(
                     &hint.channel_outpoint,
-                    tlc_min_val,
                     sufficiently_large_capacity,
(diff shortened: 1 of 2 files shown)


chenyukang (approved):

gpBlockchain:
Tested on the latest develop after this PR.

I still observed two issues:

1. When using `dry_run` to send an amount smaller than `tlc_min_value`, the request returns success.

2. When sending an amount smaller than `tlc_min_value` without `dry_run`, the send request succeeds, but `get_payment` later shows the payment as `failed`. After that, sending an amount greater than `tlc_min_value` returns `no path found`.

Could you help confirm whether these behaviors are expected, or whether we should add extra validation for `tlc_min_value` before sending?
cc @chenyukang

PR #1379, opened 2026-05-26
Reject cooperative shutdown when peer is offline

jjyr:
## Purpose

Reject cooperative channel shutdown when the peer is offline, while still allowing force shutdown.

## Summary

- Move the offline-peer check into `ChannelActor::handle_shutdown_command` for non-force shutdown commands.
- Keep force shutdown behavior unchanged so it can still proceed while the peer is offline.


gpBlockchain:
lgtm

quake (approved):

PR #1387, opened 2026-06-01
fix: mock-chain UDT funding context

Zhangcy0x3:
## Summary

This PR fixes a bug that caused trampoline routing tests with UDT
channels to fail on `upstream/develop` after commit https://github.com/nervosnetwork/fiber/commit/311e90ed3c34a13c0460e3111b696c0f0f82a59a.

---

### Mock Chain: UDT type script missing from funding output

**Root cause**<br>
Commit `311e90ed` added validation in `ChannelActorState::is_tx_final` and
`FundingTx::update_for_peer` that checks the funding transaction's first output
type script against the negotiated UDT type script. The real `CkbChainActor`
correctly builds UDT funding outputs via `FundingTxBuilder::build_funding_cell`.
However, the **test-only** `MockChainActor` has its own simplified `Fund` handler
that builds outputs manually — and it had two bugs:

1. **`None` branch (first funding call)**: when the funding tx had no existing
   outputs, the mock always built a CKB-only output (no type script), even for
   UDT channels.
2. **`Some` branch (second call, UDT)**: the UDT output was built via
   `CellOutput::new_builder()` but forgot `.lock(request.script.clone())`,
   producing an output with an empty lock script.

Additionally, `CkbChainState::build_funding_context` was hardcoding
`funding_udt_type_script: None`, so the `FundingContext` never carried the
correct UDT type script even when one was provided.

**Fix**
- Modified `build_funding_context` to accept `funding_udt_type_script` as a
  parameter and pass it through at both call sites.
- Rewrote the `MockChainActor::Fund` handler's `None` and `Some` branches to
  correctly build UDT outputs (with both type script and lock script) and to
  produce the correct `outputs_data` (local-only amount for the first call,
  combined amount for the second).
- Modified `MockChainActor::SendTx` to skip `context.verify_tx()` for
  transactions with no inputs (funding txs). The mock chain cannot resolve the
  dummy UDT type script's code hash (all zeros), causing verification failures
  that would abort the channel via `FundingTransactionFailed`.

---

### Files changed

| File | Change |
|------|--------|
| `ckb/actor.rs` | Pass `funding_udt_type_script` through `build_funding_context` |
| `ckb/tests/test_utils.rs` | Fix `MockChainActor::Fund` UDT output building (both `None` and `Some` branches); skip `verify_tx` for input-less funding txs  |

### Test results

Summary [ 162.609s] 1021 tests run: 1021 passed, 8 skipped (previously 4 failed)

### DIFF
--- a/crates/fiber-lib/src/ckb/tests/test_utils.rs
+++ b/crates/fiber-lib/src/ckb/tests/test_utils.rs
@@ -408,7 +408,7 @@ impl Actor for MockChainActor {
                     return Ok(());
                 }
 
-                let outputs = match outputs.get(0) {
+                let (outputs, outputs_data): (packed::CellOutputVec, packed::BytesVec) = match outputs.get(0) {
                     Some(output) => {
                         if output.lock() != request.script {
                             error!(
@@ -424,11 +424,19 @@ impl Actor for MockChainActor {
                             let udt_output = packed::CellOutput::new_builder()
                                 .capacity(Capacity::shannons(ckb_amount).pack())
                                 .type_(Some(udt_script.clone()).pack())
+                                .lock(request.script.clone())
                                 .build();
 
                             let mut outputs_builder = outputs.as_builder();
                             outputs_builder.replace(0, udt_output);
-                            outputs_builder.build()
+                            let outputs: packed::CellOutputVec = outputs_builder.build();
+
+                            let udt_amount = request.local_amount + request.remote_amount;
+                            let mut data = BytesMut::with_capacity(16);
+                            data.put(&udt_amount.to_le_bytes()[..]);
+                            let outputs_data: packed::BytesVec = vec![data.freeze().pack()].pack();
+
+                            (outputs, outputs_data)
                         } else {
                             let current_capacity: u64 = output.capacity().unpack();
                             capacity += current_capacity as u128;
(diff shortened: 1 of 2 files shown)


chenyukang:
I'm not sure what's the `bug 2` in your description.

Seems the tests are all failing because of mock test framework does not handle UDT script well.

After you first commit [d2f8fbc](https://github.com/nervosnetwork/fiber/pull/1387/commits/d2f8fbce9633b5429068fc6f2f26259072450137), all the tests already passed.

---

The `Peer Pubkey(...)'s feature not found, waiting for peer to send Init message` is caused by `WARN fnn::fiber::gossip: Malicious gossip peer found Pubkey(0252f505cc4360fe41852db086fb13bef78268f0a02c4ecf1d8ca1c129daa98847): InvalidBroadcastMessage`

The failure is caused by the stricter UDT channel announcement validation introduced in [77bf7c802](https://github.com/nervosnetwork/fiber/commit/77bf7c802e278dbeb6b66e2dfe80859e0c1a6a0e).

This test creates a public UDT channel, B --(UDT)--> C, so the channel announcement contains `udt_type_script: Some(...)`. However, the mock chain’s `Fund` handler used to build the initial funding output without setting `type_`, while still writing a 16-byte UDT amount into `outputs_data`. As a result, the mock funding transaction looked like a UDT funding cell in data, but its output type script was `null`.

After 77bf7c802, `verify_channel_announcement` checks that a UDT announcement matches the on-chain funding output: the output must have the same UDT type script, and the output data must contain the announced UDT amount. Therefore, gossip rejects the announcement with `UDT channel announcement but output has no type script`.

That rejection is treated as an `InvalidBroadcastMessage`, causing the related peers to be banned and disconnected. The later visible test failure, `Peer ...'s feature not found, waiting for peer to send Init message`, is only a downstream symptom of that earlier gossip ban.

Zhangcy0x3:
> I'm not sure what's the `bug 2` in your description.
> 
> Seems the tests are all failing because of mock test framework does not handle UDT script well.
> 
> After you first commit [d2f8fbc](https://github.com/nervosnetwork/fiber/pull/1387/commits/d2f8fbce9633b5429068fc6f2f26259072450137), all the tests already passed.
> 
> The `Peer Pubkey(...)'s feature not found, waiting for peer to send Init message` is caused by `WARN fnn::fiber::gossip: Malicious gossip peer found Pubkey(0252f505cc4360fe41852db086fb13bef78268f0a02c4ecf1d8ca1c129daa98847): InvalidBroadcastMessage`
> 
> The failure is caused by the stricter UDT channel announcement validation introduced in [77bf7c802](https://github.com/nervosnetwork/fiber/commit/77bf7c802e278dbeb6b66e2dfe80859e0c1a6a0e).
> 
> This test creates a public UDT channel, B --(UDT)--> C, so the channel announcement contains `udt_type_script: Some(...)`. However, the mock chain’s `Fund` handler used to build the initial funding output without setting `type_`, while still writing a 16-byte UDT amount into `outputs_data`. As a result, the mock funding transaction looked like a UDT funding cell in data, but its output type script was `null`.
> 
> After [77bf7c8](https://github.com/nervosnetwork/fiber/commit/77bf7c802e278dbeb6b66e2dfe80859e0c1a6a0e), `verify_channel_announcement` checks that a UDT announcement matches the on-chain funding output: the output must have the same UDT type script, and the output data must contain the announced UDT amount. Therefore, gossip rejects the announcement with `UDT channel announcement but output has no type script`.
> 
> That rejection is treated as an `InvalidBroadcastMessage`, causing the related peers to be banned and disconnected. The later visible test failure, `Peer ...'s feature not found, waiting for peer to send Init message`, is only a downstream symptom of that earlier gossip ban.

Thanks for the detailed analysis — that makes perfect sense. I had traced the symptom (`feature not found` / `peer_session_map` missing entries) but missed the root cause that it was a downstream effect of the gossip ban triggered by the UDT channel announcement mismatch.

I've dropped the second commit — [d2f8fbc ](https://github.com/nervosnetwork/fiber/pull/1387/changes/d2f8fbce9633b5429068fc6f2f26259072450137)alone resolves all test failures. The `PeerFeaturesNotFound` variant and `PendingOutboundChannel` retry infrastructure from the second commit can be revisited separately as an architectural enhancement if needed in the future, but they are not required for this fix.

I have also updated the PR title and description to accurately reflect this root cause and removed the sections related to the dropped commit. Let me know if anything else is needed for merge!

quake (approved):

crates/fiber-lib/src/ckb/tests/test_utils.rs
```diff
@@ -545,10 +581,26 @@ impl Actor for MockChainActor {
                         }
                     }
 
+                    // If the transaction has no inputs, it's a funding transaction.
+                    // Skip script verification since the type script on the output
+                    // will be verified when it is consumed later.
```
chenyukang, line 586:
```suggestion
                    // The inputs will always not empty for production enviornment, but it may empty for test enviornment, 
                    // `MockChainActor` only used in testing, so we skip VM verification for these synthetic transactions
```

PR #1389, opened 2026-06-02
fix: enforce private filesystem permissions on store directories and files

quake:
Store directory and database files were created with default filesystem permissions inherited from the process umask, potentially exposing channel signer keys, payment preimages, and watchtower secrets to other local users on shared hosts.

Changes:
- Set `0o700` on base directory and store directory after creation (and on existing dirs for upgrade)
- Set `0o700` on RocksDB directory, `0o600` on all RocksDB files
- Set `0o700` on SQLite directory, `0o600` on SQLite database file


quake:
@copilot rebase and fix CI errors

PR #1391, opened 2026-06-02
fix: clean stale funding exclusions and stabilize external-funding handshake CI

quake:
When a peer TxUpdate replaces the local funding transaction, `do_update_channel_funding` creates a fresh `FundingTx` and re-adds inputs to the `LiveCellsExclusionMap`. However, the old funding tx hash was never removed, leaving stale exclusion entries that could lock otherwise-usable cells after channel abort.

This fix looks up the previous funding tx hash from the channel actor state before rebuilding, and sends `RemoveFundingTx` to the CKB chain actor after the new funding tx is built successfully.

Additionally, this PR includes the external-funding handshake idempotency adjustment needed to resolve CI failures after rebasing, so post-submit duplicate `TxComplete` handling no longer causes the external-funding flow to stall.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -586,7 +586,6 @@ where
             }
             FiberChannelMessage::TxUpdate(tx) => {
                 if state.ephemeral_config.external_funding.enabled
-                    && state.ephemeral_config.external_funding.signed_submitted
                     && state.state.is_awaiting_external_funding()
                 {
                     self.handle_external_funding_tx_sync(myself, state, tx.tx)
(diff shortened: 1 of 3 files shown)


quake:
@copilot rebase and fix CI errors

chainTe:
## Security Review

Result: 1 confirmed finding (medium)
Reviewed range: `bd834319a863e6569844eb8737bff66e7037e1e2..d75916d3d67239bafd88ea9017ec7203e1c1c490`

Validation:
- Static trace covered the changed external-funding state machine, stale funding exclusion cleanup, CKB chain actor exclusion handling, and adjacent external-funding tests.

Notes:
- In the external-funding `AWAITING_EXTERNAL_FUNDING` state, a peer `TxUpdate` with the same raw transaction body can now drive the local node into `signed_submitted = true` before local operator submission is observed. Because the validation checks raw transaction equality and funding-output finality, but not local submission intent or witness authenticity, this can disable the external-funding submission timeout and leave the operator-funded opening wedged until recovery. The peer cannot redirect funds through this path because the raw transaction body must match.
- Suggested fix: keep this sync path gated on local operator submission, or require strong signed-transaction evidence before setting `signed_submitted` and clearing the submission timeout. Add a regression where the peer echoes the unsigned transaction before `SubmitSignedFundingTx` and assert the channel remains in waiting-for-external-funding-submission timeout coverage.

PR #1392, opened 2026-06-02
Deduct outer route fee from trampoline budget in routing logic

chenyukang:
When building a trampoline-routed payment, Fiber uses the payer's full max_fee_amount both as part of the amount sent to the first trampoline hop and as the budget allocated to inner trampoline forwarding. The route fee needed to reach the first trampoline is not deducted from the user's fee cap.


chenyukang:
@copilot

PR #1397, opened 2026-06-03
fix: prevent watchtower RPC errors from killing event-processing task

quake:
## Summary
- Fix GHSA-49ph-wf39-5prx: standalone watchtower RPC errors can permanently stop all event forwarding

## Problem
`forward_event_to_client` used `.expect()` on all 6 watchtower RPC calls. Any transient failure (network error, watchtower downtime, token expiry) would panic, killing the entire event-processing tokio task. Once dead, no further events were forwarded to either the standalone watchtower or the built-in watchtower actor.

## Fix
- `forward_event_to_client` return type changed from `()` to `Result<(), String>`
- 6 `.expect()` calls replaced with `.map_err(|e| format!(...))?`
- Native (`main.rs`) and WASM (`lib.rs`) callers log the error and continue the event loop

## Changes
| File | Change |
|------|--------|
| `crates/fiber-lib/src/event_handler.rs` | Return `Result`, `?` instead of `.expect()` |
| `crates/fiber-bin/src/main.rs` | Log error, don't block built-in watchtower actor |
| `crates/fiber-wasm/src/lib.rs` | Log error, keep event loop alive |

## Risk
Low — error path only triggers what was previously a panic. The built-in watchtower actor (native) is now independent of the standalone client outcome.

### DIFF
--- a/crates/fiber-bin/src/main.rs
+++ b/crates/fiber-bin/src/main.rs
@@ -340,7 +340,15 @@ async fn run_node(
                                         }
                                     }
                                     if let Some(watchtower_client) = watchtower_client.as_ref() {
-                                        forward_event_to_client(event.clone(), watchtower_client).await;
+                                        if let Err(err) =
+                                            forward_event_to_client(event.clone(), watchtower_client)
+                                                .await
+                                        {
+                                            error!(
+                                                "Failed to forward event to standalone watchtower: {}",
+                                                err
+                                            );
+                                        }
                                     }
                                     if let Some(watchtower_actor) = watchtower_actor.as_ref() {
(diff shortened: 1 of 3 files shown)


chainTe:
## Security Review

Reviewed range: `4ee1dafe686fd0ea6ee704673862be36f8e15960..51ef0c177af5f7601a5aef8653ef16ec8c548e0c`

Validation:
- Reviewed all 3 changed files and directly supporting watchtower RPC/actor paths.
- Ran a cheap diff whitespace check for the reviewed range.

Notes:
- The reviewed change replaces standalone watchtower RPC panic handling with logged errors while keeping the event-processing task alive.

PR #1400, opened 2026-06-03
fix: validate chain_hash on individual gossip messages before persisting

quake:
## Summary
- Fix GHSA-p557-4332-x4c8: per-message chain_hash validation for gossip messages

## Problem
Gossip request wrappers (`BroadcastMessagesFilter`, `GetBroadcastMessages`, `QueryBroadcastMessages`) validated `chain_hash`, but individual messages within results were not checked. A malicious peer could inject self-signed `NodeAnnouncement` messages from the wrong chain, causing:
1. Persistence of wrong-chain gossip in the local store
2. Rebroadcast to other peers
3. Poisoned sync cursors (cursor advanced before chain_hash check)

## Fix (3 changes)

| File | Change |
|------|--------|
| `gossip.rs:insert_message_to_be_saved_list` | Reject message before queueing if `chain_hash` doesn't match |
| `gossip.rs:verify_and_save_broadcast_message` | Reject message before persisting to store |
| `graph.rs:update_for_messages` | Move `update_latest_cursor` AFTER chain_hash check |

## Risk
Low — wrong-chain messages were already rejected by the graph layer; this just moves rejection earlier (before store writes and cursor advancement).

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -68,6 +68,7 @@ use fiber_types::{
     NodeAnnouncement, Pubkey,
 };
 
+
 // The maximum duration drift between the broadcast message timestamp and latest cursor in store.
 pub(crate) const MAX_MISSING_BROADCAST_MESSAGE_TIMESTAMP_DRIFT: Duration =
     Duration::from_secs(60 * 60 * 2);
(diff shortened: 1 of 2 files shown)


chainTe:
## Security Review

Reviewed range: `7caa70a6d016379259d3de2b7c37d36594af3c58..9430c6e541c4b7e31b5fdbc300cfa7d0a7ae7d13`

Validation:
- Reviewed the changed gossip persistence and graph cursor paths plus directly supporting message helper paths.
- Confirmed individual gossip `chain_hash` values are rejected before pending queue insertion, store persistence, or graph cursor advancement.

Notes:
- No new privilege boundary, funds-safety issue, payment integrity issue, secret exposure, resource-exhaustion path, or gossip poisoning bypass was identified in the reviewed range.

gpBlockchain:
Follow-up on the wrong-chain gossip persistence fix: the PR correctly prevents newly received wrong-chain gossip messages from being queued or saved, but the merged PR-head behavior appears to introduce an upgrade/restart availability regression for stores that already contain wrong-chain records.

At PR head, `NetworkGraph::update_for_messages` skips messages whose embedded `chain_hash` does not match before advancing `latest_cursor`. `NetworkGraph::load_from_store` repeatedly calls `get_broadcast_messages(&self.latest_cursor, 0)` until the store returns no messages, and the store query is exclusive after the supplied cursor. If an upgraded node already has a wrong-chain stored gossip record after `latest_cursor`, and no later valid record advances the graph cursor past it, each load iteration can return the same wrong-chain record, skip it, leave `latest_cursor` unchanged, and loop forever during graph initialization.

This matters because the issue fixed here previously allowed malicious peers to persist wrong-chain gossip records, so an attacked node may carry exactly this contaminated state into the upgraded version.

Suggested fix: either delete/quarantine mismatched stored records during graph/store loading, or use a separate store-scan cursor that advances for every returned row while keeping the graph's latest valid cursor unpoisoned for accepted same-chain messages.

PR #1401, opened 2026-06-03
fix: prevent RPC panic on invalid watchtower private keys

quake:
## Summary
- Fix GHSA-xwjm-m8pp-4x8q: watchtower RPC panics on invalid secp256k1 private keys

## Problem
`Privkey::from_slice` used `.expect("Invalid secret key")` on `SecretKey::from_slice()`. Watchtower RPC endpoints accept `Privkey` fields (e.g. `local_settlement_key`, `settlement_data.tlcs[].local_key`), and a caller could submit a 32-byte hex-valid but secp256k1-invalid scalar (e.g. all zeroes), triggering a panic instead of a structured RPC error.

## Fix
- Add `Privkey::try_from_slice` that returns `Result<Self, secp256k1::Error>`
- Update `TryFrom<JsonPrivkey> for fiber_types::Privkey` to use `try_from_slice`, returning a validation error string instead of panicking

## Changes
| File | Change |
|------|--------|
| `fiber-types/src/primitives.rs` | Add `try_from_slice` method |
| `fiber-json-types/src/convert.rs` | Use `try_from_slice` in `TryFrom` |

## Risk
Low — `from_slice` is preserved for internal callers with known-good keys. Only the JSON RPC conversion path uses the new fallible method.

### DIFF
--- a/crates/fiber-json-types/src/convert.rs
+++ b/crates/fiber-json-types/src/convert.rs
@@ -96,7 +96,8 @@ impl TryFrom<JsonPrivkey> for fiber_types::Privkey {
     type Error = String;
 
     fn try_from(jp: JsonPrivkey) -> Result<Self, Self::Error> {
-        Ok(fiber_types::Privkey::from_slice(jp.as_bytes()))
+        fiber_types::Privkey::try_from_slice(jp.as_bytes())
+            .map_err(|e| format!("Invalid private key: {e}"))
     }
 }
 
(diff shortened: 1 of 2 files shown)


chainTe:
## Security Review

Reviewed range: `a4007aa16358372f13c99dccacc7326a8862bc01..03c9c1e2449c629742659020f7714a6a85f3817d`

Validation:
- `cargo test -p fiber-json-types --features conversion,watchtower --lib privkey -- --nocapture` completed successfully; no matching unit tests were present, but the relevant conversion/watchtower feature set compiled and linked.

Notes:
- Invalid watchtower private-key scalars now return request-level validation errors instead of panicking during JSON-to-internal conversion.

PR #1402, opened 2026-06-03
fix: validate shutdown fee against the actual peer close_script

quake:
## Summary
- Fix GHSA-74ch-xcgm-jm6g: shutdown fee validated against stale peer script

## Problem
`check_shutdown_fee_valid` used `self.remote_shutdown_script` (the negotiated script from channel open) to calculate the remote shutdown fee and occupied capacity. But `build_shutdown_tx` later uses `remote_shutdown_info.close_script` (the script from the incoming `Shutdown` message).

A peer could send a `Shutdown` with a large/expensive `close_script` and a `fee_rate` that passes validation against the small old script, then force the victim to sign a cooperative close with fees computed on the oversized actual script.

## Fix
Add `remote_close_script` parameter to `check_shutdown_fee_valid` so it validates against the incoming `Shutdown.close_script` directly, matching what `build_shutdown_tx` will later use.

## Change
`channel.rs` — 4 lines changed in `check_shutdown_fee_valid` signature and body, 1 line at call site.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -1799,7 +1799,7 @@ where
             }
         };
 
-        if !state.check_shutdown_fee_valid(shutdown.fee_rate.as_u64()) {
+        if !state.check_shutdown_fee_valid(&shutdown.close_script, shutdown.fee_rate.as_u64()) {
             return Err(ProcessingChannelError::InvalidParameter(
                 "Shutdown fee is invalid".to_string(),
             ));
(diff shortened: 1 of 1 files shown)


chainTe:
## Security Review

Reviewed range: `8da4014b1775ef8f0da4f17f83512b4c64a3248e..c003b06f50f88bef3b15aee80cc5846bacfad38f`

Validation:
- Diff-scoped source review covered `crates/fiber-lib/src/fiber/channel.rs` and direct shutdown fee/transaction helpers.
- Targeted local compile/test attempt was blocked by the local environment because `ckb-librocksdb-sys` could not find `stdbool.h` while generating RocksDB bindings.

Notes:
- The reviewed change validates the peer-supplied `Shutdown.close_script` before storing it and later using it for cooperative close transaction construction.

PR #1403, opened 2026-06-03
fix: reject NodeAnnouncement addresses with mismatched /p2p peer ID

quake:
## Summary
- Fix GHSA-xjfm-j93j-6qq5: validate `/p2p` peer ID in NodeAnnouncement addresses matches the announced node_id

## Problem
`NodeAnnouncement` addresses with `/p2p` components were not validated against the announcement's `node_id`. A malicious (but otherwise valid) node could advertise addresses pointing to unrelated peers, causing victims to misdirect connections and waste outbound peer slots.

## Fix
Two-layer validation:

| Layer | File | Behavior |
|-------|------|----------|
| Gossip ingestion | `gossip.rs:insert_message_to_be_saved_list` | Reject entire message if any address has mismatched `/p2p` |
| Graph processing | `graph.rs:process_node_announcement` | Strip mismatched addresses before storing |

Addresses without `/p2p` are unaffected (backward compatible).

## Risk
Low — the fiber node itself always appends `/p2p` matching its own peer ID to all announced addresses. Legitimate peers following the same convention are unaffected.

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -27,6 +27,8 @@ use tentacle::{
     builder::MetaBuilder,
     bytes::Bytes,
     context::{ProtocolContext, ProtocolContextMutRef, SessionContext},
+    multiaddr::Protocol,
+    secio::PeerId,
     service::{ProtocolHandle, ProtocolMeta, ServiceAsyncControl, SessionType},
     traits::ServiceProtocol,
     SessionId,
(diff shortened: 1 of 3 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `a4007aa16358372f13c99dccacc7326a8862bc01..fc7ae55d1582667759073e5b35efdbcab5fd85f7`

Validation:
- Reviewed the changed gossip/graph/test files plus supporting graph RPC and reconnect-address consumers.

Notes:
- The fix rejects fresh mismatched `/p2p` addresses and strips them in the in-memory graph path, but pre-upgrade poisoned `NodeAnnouncement` rows can remain in the gossip store. Raw store read paths such as graph node pagination and reconnect-by-pubkey should apply the same `/p2p` peer ID consistency filter, or contaminated rows should be migrated/quarantined/deleted, so upgraded nodes cannot keep exposing or dialing stale mismatched addresses.

PR #1407, opened 2026-06-03
fix: validate hash_algorithm and payment_hash consistency for forwarded TLCs

quake:
## Summary
- Fix GHSA-vg8f-mwh8-j6xp: forwarded TLC hash_algorithm can mismatch upstream

## Problem
`check_add_tlc_consistent` only validated UDT type between forwarding channels. A malicious payer could construct an incoming TLC with `hash_algorithm=SHA256` while the onion packet for the next hop specified `hash_algorithm=BLAKE2B`. The forwarding node creates the downstream TLC with mismatched hash semantics — the downstream peer can fulfill with a preimage validated against one algorithm, while the upstream TLC is validated against a different algorithm, causing the forwarding node to pay downstream but fail to claim upstream.

## Fix
In `check_add_tlc_consistent`, look up the previous channel's TLC info and validate that the forwarded `AddTlcCommand` has the same `hash_algorithm` and `payment_hash`.

## Changes
`crates/fiber-lib/src/fiber/channel.rs` — ~20 lines added to `check_add_tlc_consistent`

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -1420,6 +1420,21 @@ where
                     .map_err(|err| ProcessingChannelError::PeelingOnionPacketError(err.to_string()))
                     .map_err(ProcessingChannelError::without_shared_secret)?;
                 let shared_secret = peeled.shared_secret;
+
+                // The onion payload is encrypted but the hash_algorithm field is
+                // not cryptographically bound to the onion. A malicious sender
+                // can set a different hash_algorithm on the wire AddTlc than in
+                // the onion hop data, causing the forwarding node to produce an
+                // outgoing TLC that a downstream peer can fulfill while the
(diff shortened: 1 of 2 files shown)


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -1420,6 +1420,21 @@ where
                     .map_err(|err| ProcessingChannelError::PeelingOnionPacketError(err.to_string()))
                     .map_err(ProcessingChannelError::without_shared_secret)?;
                 let shared_secret = peeled.shared_secret;
+
+                // The onion payload is encrypted but the hash_algorithm field is
+                // not cryptographically bound to the onion. A malicious sender
+                // can set a different hash_algorithm on the wire AddTlc than in
```
chenyukang, line 1426: add a new test case in tests?

chenyukang:
seems this not cover `GHSA-mhg3-439x-2fch`, I will create a seperated PR.

PR #1408, opened 2026-06-03
Fix: enhance TLC error packet decoding to reject spoofed nodes

chenyukang:
Malicious node may use other node's pubkey as node_id in tlc error, we should check this.

### DIFF
--- a/crates/fiber-lib/src/fiber/tests/types.rs
+++ b/crates/fiber-lib/src/fiber/tests/types.rs
@@ -689,6 +689,27 @@ fn test_tlc_err_packet_encryption() {
             .expect("decrypted");
         assert_eq!(decrypted_tlc_fail_detail, tlc_fail_detail);
     }
+
+    {
+        // The error packet authenticates the second hop as reporter, so a payload that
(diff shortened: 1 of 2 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: d4833cc87956a2d3978b91370e35b3ceb4e3c1e0^..d4833cc87956a2d3978b91370e35b3ceb4e3c1e0

Validation:
- Focused test passed: cargo test -p fnn --features sqlite test_tlc_err_packet_encryption -- --nocapture.
- Static trace confirmed the encrypted-path reporter check is still bypassed by the existing plaintext fast path.

Notes:
- TlcErrPacket::decode returns plaintext zero-HMAC packets before comparing error_node_id with the authenticated hop reporter. A peer can still send a plaintext RemoveTlcFail that names a different route node/channel, so spoofed node/channel failure data can reach payment history and graph updates. Consider rejecting plaintext TLC error packets when decoding a non-empty hop route, or limiting plaintext decoding to local/origin errors.

PR #1415, opened 2026-06-04
fix: bound peer-added funding tx complexity before chain lookups

doitian:
## Summary

During funding collaboration, a remote peer's `TxUpdate` could include an arbitrary number of non-local inputs. Each peer-added input was verified with a separate `get_live_cell` CKB RPC inside `VerifyFundingTx`, so a single near-frame-size P2P message could amplify into many serialized chain lookups while the channel actor blocked, enabling a remote availability attack across channels or auto-accepted openings.

This change bounds peer-added funding transaction complexity using cheap structural checks performed **before** any chain lookup.

- Add `validate_peer_funding_tx_complexity` in [`ckb/funding/limits.rs`](crates/fiber-lib/src/ckb/funding/limits.rs) enforcing budgets relative to the local funding tx: serialized size (P2P frame cap, 133120 bytes), peer-added inputs (64), outputs (3), and cell deps (8). Witnesses are not capped separately; their cost is bounded by the serialized size limit.
- Call the budget check first in `FundingTx::update_for_peer` and move the per-input `get_live_cell` loop to the end, after all cheap checks.
- Re-check complexity in `handle_tx_collaboration_msg` before the blocking `VerifyFundingTx` call (defense in depth).
- New `FundingError::PeerFundingTxExceedsLimit(String)` (non-temporary) names the violated budget; the channel actor relays it to the peer via `TxAbort`.
- Unit tests in [`ckb/tests/funding_limits_tests.rs`](crates/fiber-lib/src/ckb/tests/funding_limits_tests.rs); limits documented in [`docs/specs/p2p-message.md`](docs/specs/p2p-message.md).

## Breaking

The external funding script may fail when the built tx exceed the complexity upper bound.

## Test plan

- [x] `cargo nextest run -p fnn -p fiber-bin funding_limits_tests ckb::tests::error`
- [x] `cargo clippy --all-targets --all-features -p fnn -p fiber-bin -- -D warnings`
- [x] `cargo fmt --all -- --check`

### DIFF
--- a/crates/fiber-lib/src/ckb/error.rs
+++ b/crates/fiber-lib/src/ckb/error.rs
@@ -24,6 +24,9 @@ pub enum FundingError {
     #[error("Peer sent us an invalid funding tx")]
     InvalidPeerFundingTx,
 
+    #[error("Funding tx rejected: peer-added complexity exceeds allowed limit ({0})")]
+    PeerFundingTxExceedsLimit(String),
+
     #[error(
         "Funding tx rejected: peer-added input #{input_index} has the local funding source lock args"
     )]
@@ -105,6 +108,7 @@ impl FundingError {
             CkbTxBuilderError(e) => error_chain_has_transient(e),
             CkbTxUnlockError(e) => error_chain_has_transient(e),
             PeerInputUsesOurFundingLock { .. } => false,
+            PeerFundingTxExceedsLimit(_) => false,
             _ => false,
         }
     }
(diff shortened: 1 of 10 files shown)


chainTe:
## Security Review

Reviewed range: `0dba35a971d0e18540355333cee7011cd7195758..b12171f2a962a5bf29fe6596b2d5778c279c7d7e`

Validation:
- Targeted Cargo tests were attempted, but local builds failed before test execution: the default feature build omits a store backend, and `--features rocksdb` failed in `ckb-librocksdb-sys` because clang could not find `stdbool.h`.
- Reviewed the peer `TxUpdate` funding path, the new peer-added complexity limits, the CKB verification path, error classification, protocol docs, and regression tests.

Notes:
- The new checks bound peer-added funding transaction complexity before the expensive CKB chain lookups, with a repeat guard in the CKB funding verifier.

PR #1425, opened 2026-06-08
fix: persist commitment ack intent before send

jjyr:
## Summary

- Persist the updated channel actor state together with the pending `CommitDiff` before sending `CommitmentSigned`.
- Move the in-memory ack/signing-state update ahead of the outbound frame so restart state matches the durable replay metadata.
- Replay the stored commitment during reestablish when the node crashed after persisting the send intent but before the peer observed the frame.

## Root Cause

`handle_commitment_signed_command` stored `CommitDiff` before sending `CommitmentSigned`, but persisted `waiting_ack` only later through the actor-message tail write. A crash in between could leave durable state saying no ack was expected while the peer may legitimately send `RevokeAndAck`.

## Validation

- `cargo nextest run --features sqlite -p fnn channel_commit_diff`
- `cargo nextest run --features sqlite -p fnn reestablish`
- `cargo clippy --features sqlite -p fnn --all-targets -- -D warnings`

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -2034,8 +2034,21 @@ where
             replay_order_hint: Some(ReplayOrderHint::RevokeThenCommit),
             created_at_ms: now_timestamp_as_millis_u64(),
         };
+        state.last_was_revoke = false;
+        let signing_flags = match flags {
+            CommitmentSignedFlags::SigningCommitment(flags) => {
+                let flags = flags | SigningCommitmentFlags::OUR_COMMITMENT_SIGNED_SENT;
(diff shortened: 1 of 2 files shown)


chainTe:
## Security Review

Reviewed range: `2b279115e01be5735526e0505b1c6b431cf3c9fd..f4c2b27ca0d2bb8220887ff18b74df114b485868`

Validation:
- Completed diff-scoped source review of `crates/fiber-lib/src/fiber/channel.rs` and `crates/fiber-lib/src/store/store_impl/mod.rs`.
- Ran `git diff --check` for the reviewed range; no whitespace errors were reported.

Notes:
- Review was scoped to the PR diff and directly supporting channel-state persistence/reestablishment paths.

PR #1426, opened 2026-06-08
[codex] Fix synthetic rejected tx tracing

jjyr:
## Summary

- Treat synthetic `ReportRejected` results as immediate tracer deliveries by passing `None` into `report_tracing_result` instead of applying the normal confirmation delay.
- Stop retrying `send_tx` after a permanent send failure has already been reported.
- Add a regression covering `ReportRejected` followed by an `Unknown` poll while the tracer uses four confirmations.

## Root Cause

`InFlightCkbTxActor` reports permanent CKB send failures through `ReportRejected`, but the tx tracing actor previously stored that synthetic `Rejected` status at the current tip and required the normal confirmation delay. A later poll could then replace the status with `Unknown` before the callback fired, leaving funding/closing recovery stuck.

## Validation

- `cargo nextest run synthetic_rejected_callback_fires_before_next_unknown_poll -p fnn --features rocksdb`
- `cargo nextest run -p fnn --features rocksdb ckb::tx_tracing_actor`
- `cargo nextest run test_abort_funding_on_committing_funding_tx_on_chain -p fnn --features rocksdb`
- `cargo clippy --all-targets --features rocksdb -p fnn -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/fiber-lib/src/ckb/tx_tracing_actor.rs
+++ b/crates/fiber-lib/src/ckb/tx_tracing_actor.rs
@@ -147,12 +147,18 @@ impl Actor for CkbTxTracingActor {
         match message {
             CreateTracer(arguments) => state.create_tracer(myself, arguments).await,
             RemoveTracers(arguments) => state.remove_tracers(arguments),
-            ReportRejected(tx_hash) => state.report_rejected(myself, tx_hash).await,
+            ReportRejected(tx_hash) => state.report_tracing_result(
+                CkbTxTracingResult {
+                    tx_hash,
+                    tx_status: TxStatus::Rejected(
+                        "Transaction rejected when submitting".to_string(),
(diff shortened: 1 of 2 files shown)


chainTe:
## Security Review

Reviewed range: `2b279115e01be5735526e0505b1c6b431cf3c9fd..afb3ae19c5c804a75259e724da5cc133a1f83b54`

Validation:
- `git diff --check origin/develop...HEAD` passed in the review checkout.
- Attempted `cargo test -p fnn synthetic_rejected_callback_fires_before_next_unknown_poll`; local compilation failed before running because the command did not enable a `fiber-store` backend feature.

Notes:
- Reviewed the synthetic rejected transaction tracing path, permanent send failure retry suppression, and funding/closing failure consumers.

PR #1441, opened 2026-06-11
[codex] Fix active gossip sync accounting

jjyr:
## Summary

Fix active gossip sync state accounting so disconnected peers no longer consume the active sync budget, and make active sync cursor advancement depend on store validation instead of trusting the last remote response message.

## Root Cause

Active gossip sync used a monotonic global finished-peer counter. A transient peer could return an empty GetBroadcastMessagesResult, be counted as finished, disconnect, and leave that stale count consuming future active sync slots. Active sync also advanced its per-peer cursor from the last response message before store validation, allowing attacker supplied timestamps to poison sync progress.

## Changes

- Track active sync completion through current peer state instead of a stale global counter.
- Add session_id and sync_id guards to ignore stale completion messages from old sync actors or disconnected/reconnected peers.
- Move active sync cursor advancement behind store validation via SaveActiveSyncMessages.
- Reject future timestamp active sync messages without saving or advancing cursor.
- Add regressions for disconnected finished peers releasing budget and future timestamp active sync rejection.

## Validation

- cargo check -p fnn --tests --features rocksdb
- cargo clippy --all-targets --features rocksdb -p fnn -- -D warnings
- cargo nextest run -p fnn --features rocksdb test_active_sync_rejects_future_timestamp_without_saving
- cargo nextest run -p fnn --features rocksdb test_disconnected_finished_active_sync_peer_releases_budget
- cargo nextest run -p fnn --features rocksdb test_gossip_sync_starts_immediately_on_peer_connect
- cargo nextest run -p fnn --features rocksdb test_sync_node_announcement_on_startup
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -430,7 +430,12 @@ pub(crate) enum GossipActorMessage {
     PruneStaleGossipMessages(u64),
 
     // The active syncing process is finished for a peer.
-    ActiveSyncingFinished(Pubkey, Cursor),
+    ActiveSyncingFinished {
+        peer: Pubkey,
+        session_id: SessionId,
+        sync_id: u64,
+        cursor: Cursor,
+    },
 
     // A malicious peer is found during gossip validation or policy checks.
     MaliciousPeerFound {
@@ -617,7 +622,7 @@ where
 
     async fn update_peer_filter(
         &self,
-        state: &mut GossipActorState<S, C>,
+        state: &mut GossipActorState<S>,
         pubkey: &Pubkey,
         after_cursor: &Cursor,
         myself: ActorRef<GossipActorMessage>,
@@ -756,11 +761,11 @@ impl DelayedGossipMessageQueue {
     }
 }
 
-pub struct GossipSyncingActorState<S, C> {
+pub struct GossipSyncingActorState<S> {
     peer_pubkey: Pubkey,
+    session_id: SessionId,
+    sync_id: u64,
     gossip_actor: ActorRef<GossipActorMessage>,
-    chain_actor: ActorRef<CkbChainMessage>,
-    chain_client: C,
     store: ExtendedGossipMessageStore<S>,
     // The problem of using the cursor from the store is that a malicious peer may only
     // send large cursor to us, which may cause us to miss some messages.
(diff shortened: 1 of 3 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `3bbf5ea0ed7debd83a707b5f28264bee2fd7371f..63dd3fe8ebe7bc0a6fd77ceda122115b4a2d60cf`

Validation:
- Reviewed the active gossip sync accounting changes in `crates/fiber-lib/src/fiber/gossip.rs` and the new regression tests.

Notes:
- Medium: a connected peer selected for active sync can return a future-timestamp gossip message. The new `SaveActiveSyncMessages` path rejects it and the child sync actor stops, but the parent peer state appears to remain `ActiveGet`. Because `PolicyRejectedMessage` only scores 25 against the default ban threshold of 100, one such response can keep consuming an active-sync slot without immediate disconnect. Please release or reset the matching active-sync state on rejected/error child exits, guarded by the existing `session_id` and `sync_id`, and add a regression test where the rejected peer stays connected while another honest peer can still be selected for active sync.

jjyr:
Fixed by introduce a new failed state, the failed state does not count as an active sync, so other peers can be selected immediately. It also has a cooldown: after retry_after_ms, the peer can be selected for active sync again. If the peer keeps returning rejected gossip messages, the existing gossip policy score keeps increasing and eventually reaches the ban threshold, causing the peer to be banned and disconnected.

@chainTe please review again

chainTe:
## Security Review Follow-up

Result: the previously reported active gossip sync budget issue is fixed in the updated PR head.
Reviewed head: `287ce99bc73ec13137af5581b8814d2c9ee2256e`

Validation:
- Confirmed the rejected active-sync path now sends `ActiveSyncingAborted` with `peer`, `session_id`, and `sync_id`, and the parent only applies it to the matching live session.
- Confirmed rejected peers move to `FailedActiveSyncing(retry_after_ms)`, which does not count as active sync capacity, while repeated rejected gossip still contributes to policy scoring toward ban/disconnect.
- Passed focused local tests with SQLite backend:
  - `cargo test -p fnn --features sqlite test_active_sync_rejects_future_timestamp_without_saving -- --nocapture`
  - `cargo test -p fnn --features sqlite test_rejected_active_sync_peer_releases_budget_while_connected -- --nocapture`
  - `cargo test -p fnn --features sqlite test_repeated_rejected_active_sync_peer_is_eventually_banned -- --nocapture`

Note: a RocksDB-backed focused test build was attempted locally but this environment could not compile `ckb-librocksdb-sys` because binding generation could not find `stdbool.h`.

PR #1440, opened 2026-06-11
fix: prevent funding abort when tx is committed on chain after restart

doitian:
Fix the funding transaction abort after restart issue described in #1433.

When a node restarts while a funding transaction is being confirmed (the
tx was accepted into the pool before restart, then committed to a block
during the restart), re-broadcasting the already committed transaction
after restart returns `-301 TransactionFailedToResolve: Unknown(OutPoint)`
because the inputs are already spent on chain. Previously,
`InFlightCkbTxActor` treated this as a permanent broadcast error and
immediately reported the transaction as rejected via `ReportRejected`,
which caused `abort_funding` to be called even though the transaction
was already confirmed on chain.

Note: when the tx is still in the pool (not yet committed), the
CkbChainActor handles the duplicate gracefully (error codes -1107/-1111
are swallowed and Ok(()) is returned). The -301 error only surfaces
when the tx has been committed to a block.

## Changes

- Removed the `is_permanent_error` / `ReportRejected` path that classified
  `-301` as a fatal rejection.
- Removed the `permanent_send_failure_reported` guard flag.
- `send_tx` now queries `get_transaction` before broadcasting:
  - If **Committed** or **Rejected**: skip the broadcast entirely and let the
    existing tx tracer handle the terminal status.
  - If **Pending**, **Proposed**, or **Unknown**: proceed with broadcast.
  - If the query fails: log a warning and attempt the broadcast anyway.

_Note: This PR description was generated by AI (opencode:deepseek-v4-flash)._

PR #1444, opened 2026-06-12
fix: report permanent SendTx failures via tracer to restore abort_funding

doitian:
## Summary
- Replace `ReportRejected` with `ReportSendTxError` so permanent SendTx failures are stored in `CkbTxTracingActor` and synthesized as `Rejected` only when chain status is still `Unknown`.
- Restore `InFlightCkbTxActor` reporting of permanent `-301` / `TransactionFailedToResolve` errors so `FundingTransactionFailed` and `abort_funding` fire again on unbroadcastable funding txs.
- Preserve PR #1440 behavior: preflight skips `SendTx` when status is already `Committed` or `Rejected`, preventing false abort after restart.

## Test plan
- [x] `synthetic_rejected_callback_fires_before_next_unknown_poll`
- [x] `unknown_poll_with_permanent_send_tx_error_reports_rejected`
- [x] `non_unknown_poll_clears_send_tx_error`
- [x] `permanent_send_tx_error_on_unknown_tx_emits_funding_transaction_failed` (GHSA regression)
- [x] `committed_preflight_skips_send_tx_and_error_reporting` (restart regression)

### DIFF
--- a/crates/fiber-lib/src/ckb/actor.rs
+++ b/crates/fiber-lib/src/ckb/actor.rs
@@ -77,7 +77,7 @@ pub enum CkbChainMessage {
     SendTx(TransactionView, RpcReplyPort<Result<(), RpcError>>),
     CreateTxTracer(CkbTxTracer),
     RemoveTxTracers(Hash256),
-    ReportRejected(Hash256),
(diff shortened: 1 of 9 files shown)


gpBlockchain:
lgtm

PR #1451, opened 2026-06-15
fix: enforce commitment delay epoch check on non-first watchtower settlements

quake:
## Summary
- Non-first watchtower settlements (settlement_witness: Some path) did not check that the full commitment delay had elapsed before allowing remaining balance settlement after TLC processing
- This allowed a malicious peer to settle stale commitment balances after only a reduced TLC delay (1/3 or 2/3 of full commitment delay)
- Fix: add `cell_header_epoch + delay_epoch <= current_epoch` checks before allowing remaining balance settlement in non-first settlement paths


quake:
it's by design

PR #1454, opened 2026-06-15
fix: use pruning-aware apply_remove_tlc in clean_up_failed_tlcs to prevent unbounded commitment-point growth

quake:
## Summary
- `clean_up_failed_tlcs` called `self.tlc_state.apply_remove_tlc()` directly, bypassing the `remote_commitment_points` pruning in `ChannelActorState::apply_remove_tlc()`
- This caused unbounded growth of `remote_commitment_points` on failed TLC cleanup, leading to unbounded per-channel disk usage
- Fix: use `self.apply_remove_tlc()` which includes commitment-point pruning

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -6515,7 +6515,7 @@ impl ChannelActorState {
             .collect::<Vec<TLCId>>();
 
         for tlc_id in to_be_removed_ids {
-            self.tlc_state.apply_remove_tlc(tlc_id);
+            self.apply_remove_tlc(tlc_id);
         }
     }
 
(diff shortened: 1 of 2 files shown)


chenyukang:
The failed test `test_trampoline_routing_max_trampoline_hops_success` is not related to this change.

This PR seems ok for me, it's better to add a test for this change.

quake:
Added unit test `test_clean_up_failed_tlcs_prunes_commitment_points` that:

1. Sets up a channel with 3 remote_commitment_points at numbers 1, 3, 5
2. Adds one TLC (refs commit 3) that is confirmed and flagged REMOVE — will be cleaned up
3. Adds another TLC (refs commit 5) that is NOT confirmed — should remain
4. Calls `clean_up_failed_tlcs()`
5. Asserts that commit point 3 is pruned while 1 and 5 remain

**Without the fix**: `self.tlc_state.apply_remove_tlc` bypasses pruning → all 3 points remain.  
**With the fix**: `self.apply_remove_tlc` includes pruning → only 1 and 5 remain.

The `test_trampoline_routing_max_trampoline_hops_success` failure is unrelated (CI sqlite timeout)

PR #1455, opened 2026-06-15
fix: verify channel update signatures in TryBroadcastMessages to prevent gossip bypass via onion errors

quake:
## Summary
- `ChannelUpdate` messages from onion errors (e.g., TLC failure updates) bypassed gossip verification when broadcast via `TryBroadcastMessages`
- This allowed remote peers to inject unverified channel updates through error-into-routing attacks
- Fix: validate `ChannelUpdate` signatures against stored channel announcements before broadcasting; reject invalid/unverifiable messages

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -3831,13 +3831,49 @@ where
             }
             GossipActorMessage::TryBroadcastMessages(messages) => {
                 trace!("Trying to broadcast message: {:?}", &messages);
-                state
-                    .store
-                    .actor
-                    .send_message(ExtendedGossipMessageStoreMessage::SaveAndBroadcastMessages(
-                        messages,
-                    ))
-                    .expect("store actor alive");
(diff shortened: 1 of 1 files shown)


jjyr (approved):

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `9ac566c50c23398fff401f4a92e4212b8963c8a6..7c0422f364781ae992fa7cff40396159b8f2891c`

Validation:
- Reviewed the changed `TryBroadcastMessages` path plus supporting payment error, network forwarding, store, and graph update paths.

Finding:
- Medium: the missing-announcement branch still allows an update-bearing onion error to reach `SaveAndBroadcastMessages` without signature verification. When `get_latest_channel_announcement(&channel_update.channel_outpoint)` is `None`, the new guard returns `true`, but this entrypoint then uses the direct save/broadcast sink rather than the dependency-aware gossip path. That can persist an unverifiable orphan `ChannelUpdate` by outpoint/direction. Please either drop missing-announcement updates here or route them through the normal dependency-aware verification path, and add regression coverage for an onion-error `ChannelUpdate` whose announcement is absent locally.

PR #1456, opened 2026-06-15
fix: add reestablish timeout to prevent peer from stalling TLC safety handling indefinitely

quake:
## Summary
- When a channel enters reestablishing state after peer disconnect, `reestablishing=true` blocks all non-ReestablishChannel messages and disables the peer response timeout
- A malicious peer that reconnects but never sends `ReestablishChannel` can stall the channel indefinitely, suppressing TLC expiry checks and preimage fulfillment
- Fix: add a 5-minute `REESTABLISH_TIMEOUT` that triggers a force shutdown if reestablish doesn't complete within the timeout


crates/fiber-lib/src/fiber/gossip.rs
```diff
@@ -3823,13 +3823,34 @@ where
...
             GossipActorMessage::TryBroadcastMessages(messages) => {
                 trace!("Trying to broadcast message: {:?}", &messages);
-                state
-                    .store
-                    .actor
-                    .send_message(ExtendedGossipMessageStoreMessage::SaveAndBroadcastMessages(
-                        messages,
-                    ))
-                    .expect("store actor alive");
+                let validated_messages: Vec<BroadcastMessageWithTimestamp> = messages
```
jjyr: This change do not described in the PR content, please add description about the change or revert it

PR #1461, opened 2026-06-16
fix: prevent funding abort when channel already signed

quake:
## Summary

Fixes queued stop race in funding timeout handling (#6).

`CheckFundingTimeout` checks `can_abort_funding_on_timeout()` then enqueues `Stop(AbortFunding)`. If peer `TxSignatures` arrives between the timeout check and Stop execution, the channel transitions to signed/broadcast state — but the Stop handler still unconditionally aborts, deleting state for a funding tx that may confirm on chain.

## Fix

Re-verify `can_abort_funding_on_timeout()` in the `Stop(AbortFunding)` handler before proceeding. If the channel state has advanced past the abortable window, skip the abort and just stop the actor.

## Verification

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -3402,7 +3402,6 @@ where
                             state.get_id(),
                             state.state
                         );
-                        myself.stop(Some(format!("ChannelStopped: {:?}", reason)));
                         return Ok(());
                     }
                     if let Some(detail) = reason.funding_abort_detail() {
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -3393,13 +3393,9 @@ where
                 if reason == StopReason::Abandon {
                     state.update_state(ChannelState::Closed(CloseFlags::ABANDONED));
                 } else if reason.is_abort_funding() {
-                    // Re-verify that abort is still valid in case channel state
-                    // changed after the timeout event was scheduled (e.g. a queued
-                    // TxSignatures already signed and broadcast the funding tx).
-                    // Only skip abort when the funding is actually confirmed on chain;
-                    // if funding failed (FundingTransactionFailed) or hasn't yet
-                    // confirmed, abort remains correct regardless of signature flags.
-                    if state.funding_tx_confirmed_at.is_some() {
+                    // For timeout-based aborts: re-verify that abort is still
+                    // valid. FundingFailed always proceeds unconditionally.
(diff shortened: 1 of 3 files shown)


doitian (approved):

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `7fca7b085ca99856b88d76bfd29bc4d57b281a9d..76f2229beef101019e1ad6b3198955e13d72957b`

Validation:
- Source trace reviewed the changed `Stop(AbortFunding)` branch plus network actor cleanup and funding-confirmation routing.

Finding:
- High: the new non-abortable `Stop(AbortFunding)` branch skips closing/deleting the channel, but still calls `myself.stop(...)`. Because the state is not `Closed(FUNDING_ABORTED)`, `post_stop` reports `StopReason::PeerDisConnected`; `NetworkActor::on_channel_actor_stopped` then removes the channel's `pending_channels`/`outpoint_channel_map` entries. A later funding confirmation depends on `pending_channels.remove(outpoint)` to route `FundingTransactionConfirmed`, so a timeout stop that loses the race to `TxSignatures` can still strand an already-signed funding transaction outside `ChannelReady`.

Suggested fix: when `can_abort_funding_on_timeout()` is false, treat the stale abort stop as obsolete and keep the actor/outpoint routing alive, or use a lifecycle path that preserves confirmation routing and does not mark the opening attempt failed. A regression test should queue `Stop(AbortFunding)`, advance the channel past the local-signature point before the stop runs, then assert a later funding confirmation still reaches the channel.

chainTe:
## Security Review

Result: 1 confirmed finding remains after re-review.
Reviewed range: `260c442a35f37eaf55abfbb1b5f9de05a571f132..4cd12fdf535e441fc424edc41f8c244e80ff1802`

Validation:
- Re-reviewed the updated `Stop(AbortFunding)` guard, timeout enqueue path, funding-failure path, actor cleanup, and funding confirmation routing.

Notes:
- The latest guard skips abort only after the channel actor has already processed funding confirmation (`funding_tx_confirmed_at.is_some()`). A stale timeout `Stop(AbortFunding)` can still be queued before signatures, then run after `OUR_TX_SIGNATURES_SENT` but before confirmation is processed. In that signed-before-confirmation window the guard does not fire, so the channel still closes as funding-aborted and cleanup can remove the confirmation routing.
- The added regression test does not exercise the actual `ChannelEvent::Stop(StopReason::AbortFunding)` handler in that window; it only mutates/checks state fields directly. Please distinguish timeout-originated stale aborts from real `FundingTransactionFailed` aborts, or carry an explicit abort source so funding failure can still abort while obsolete timeout stops return without stopping the actor.

PR #1466, opened 2026-06-16
fix: debounce payment retry scans on reestablish to prevent resource exhaustion

quake:
## Summary

ChannelReady events from reestablish trigger scanning of all pending payment attempts via `get_pending_attempts_by_channel_outpoint`. A malicious peer that rapidly reconnects/reestablishes can exhaust resources by triggering repeated store scans and spawning excessive `RetrySendPayment` events (#13).

## Fix

Add a 60-second `CHANNEL_READY_RETRY_DEBOUNCE_MS` cooldown tracked per-channel outpoint in a `HashMap<OutPoint, u64>`. Scans are only performed once per debounce interval, limiting the impact of reestablish spam.

## Verification

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -174,6 +174,11 @@ const FUNDING_RETRY_MAX_TOTAL_ATTEMPTS: u32 = 5;
 const FUNDING_RETRY_BASE_MILLIS: u64 = 2000;
 const FUNDING_RETRY_MAX_MILLIS: u64 = 60_000;
 
+/// Debounce interval for payment retry scans triggered by ChannelReady
+/// (e.g. after reestablish). Prevents resource exhaustion when a peer
+/// rapidly reconnects/reestablishes.
+const CHANNEL_READY_RETRY_DEBOUNCE_MS: u64 = 60_000;
+
 fn funding_retry_delay(retry_count: u32) -> Duration {
     let shift = retry_count.min(63);
     let factor = 1u64 << shift;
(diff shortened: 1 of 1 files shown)


jjyr:
Please fix following findings

[P1] Debounce drops trailing ChannelReady, can strand payments. At [network.rs:1469](https://github.com/nervosnetwork/fiber/blob/16f79b81c792b95c6299948ab66ab3617a906adf/crates/fiber-lib/src/fiber/network.rs#L1469), skipped event is just discarded. But payment actor can stop non-final when channel down ([payment.rs:1078](https://github.com/nervosnetwork/fiber/blob/16f79b81c792b95c6299948ab66ab3617a906adf/crates/fiber-lib/src/fiber/payment.rs#L1078)), and network removes it ([network.rs:5702](https://github.com/nervosnetwork/fiber/blob/16f79b81c792b95c6299948ab66ab3617a906adf/crates/fiber-lib/src/fiber/network.rs#L5702)). If earlier ChannelReady already set debounce timestamp, later real reconnect inside 60s is skipped, no actor restarts, and payment may stay dormant forever if channel remains ready. Fix: use trailing-edge debounce: when suppressed, schedule one scan at cooldown expiry.

[P3] New debounce map not cleared on channel removal. outpoint_channel_map is retained/cleaned at [network.rs:5525](https://github.com/nervosnetwork/fiber/blob/16f79b81c792b95c6299948ab66ab3617a906adf/crates/fiber-lib/src/fiber/network.rs#L5525), but last_channel_ready_scan at [network.rs:3970](https://github.com/nervosnetwork/fiber/blob/16f79b81c792b95c6299948ab66ab3617a906adf/crates/fiber-lib/src/fiber/network.rs#L3970) keeps old outpoints forever. Long-lived node + channel churn = stale growth. Clear same place as outpoint_channel_map.

gpBlockchain:
Security review found one issue that still appears present on current develop after this PR was merged.

The new trailing-edge debounce suppresses immediate pending-payment scans, but every suppressed ChannelReady schedules its own delayed ChannelReady with send_after. Because the state only records last_channel_ready_scan and does not record that a trailing scan is already pending, a burst of reconnect/reestablish-driven ChannelReady events can enqueue many delayed actor messages. At cooldown expiry, one delayed event refreshes the timestamp and the rest can reschedule again, turning a burst into long-lived network-actor/timer pressure plus repeated ChannelReady notifications and periodic retry scans.

Suggested fix: coalesce trailing scans per channel outpoint, for example with a pending trailing-scan set or stored/cancellable timer handle, and consider using a dedicated internal retry-scan event so the delayed callback does not re-emit public ChannelReady side effects. A focused regression should assert that multiple ChannelReady events inside one debounce interval create at most one delayed retry scan per channel.

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `260c442a35f37eaf55abfbb1b5f9de05a571f132..9d8c34224a6958dade2f968976b866e288a30cfc`

Validation:
- Focused static trace of the changed `NetworkActorEvent::ChannelReady` retry-scan debounce path and supporting channel reestablish emission path.

Notes:
- The debounce still schedules one deferred `ChannelReady` timer for every suppressed reestablish completion. A burst of repeated reconnect/reestablish events can leave many delayed actor messages; when they fire together, one updates `last_channel_ready_scan` and the rest are suppressed again and re-schedule into the next debounce window. Please coalesce deferred retry scans per channel/outpoint, for example with a pending-deferred marker/handle that is cleared when the deferred scan runs or the channel is removed.

PR #1467, opened 2026-06-16
fix: remove panic paths from UDT regex and CKB RPC byte responses

quake:
## Summary

Removes unconditional panics that could crash the node:

1. **UDT regex**: `Regex::new(&udt.script.args).expect("invalid expression")` → uses `match` to handle invalid regex, logs warning and skips entry
2. **CKB RPC bytes format**: `panic!("bytes response format not used")` in two `From` impls → uses `and_then` to return `None` for the transaction field, logs warning

## Verification

### DIFF
--- a/crates/fiber-lib/src/ckb/client.rs
+++ b/crates/fiber-lib/src/ckb/client.rs
@@ -33,10 +33,14 @@ impl From<Option<ckb_jsonrpc_types::TransactionWithStatusResponse>> for GetTxRes
     fn from(value: Option<ckb_jsonrpc_types::TransactionWithStatusResponse>) -> Self {
         match value {
             Some(response) => Self {
-                transaction: response.transaction.map(|tx| match tx.inner {
-                    ckb_jsonrpc_types::Either::Left(json) => transaction_view_from_json(json),
-                    ckb_jsonrpc_types::Either::Right(_) => {
-                        panic!("bytes response format not used");
+                transaction: response.transaction.and_then(|tx| match tx.inner {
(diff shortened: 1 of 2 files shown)


jjyr (approved):

doitian (approved):

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `develop..ca3b5595bb091c044c70c8a2e1df3eeab4b64c4b`

Validation:
- Focused review traced the changed disconnect handling against the inbound channel-open path, base-branch behavior, persistent store, and pending-channel RPC listing.

Notes:
- A remote peer can repeatedly create inbound pending channel-open records and disconnect before local acceptance. This PR now marks those records failed and persists them while freeing the bounded in-memory pending slot, so failed records can accumulate without the reviewed code showing a cap, TTL, or pruning path. Please restore deletion for remote-aborted inbound requests that never created a channel actor, or add bounded retention before freeing the pending slot.

chainTe:
## Security Review Correction

The earlier security review comment for this PR was incorrect: the reported channel-open persistence issue is unrelated to PR #1467 and is withdrawn from this PR review.

Corrected result: No reportable findings in the actual PR #1467 diff.

Reviewed range: `develop..ca3b5595bb091c044c70c8a2e1df3eeab4b64c4b`

Validation:
- Rechecked the PR changed-file set: `crates/fiber-lib/src/ckb/client.rs` and `crates/fiber-lib/src/ckb/config.rs`.

PR #1468, opened 2026-06-16
fix: prevent duplicate InFlightCkbTxActor spawns for the same tx

quake:
## Summary

`trace_tx` and `send_tx` unconditionally spawned a new `InFlightCkbTxActor` per call with no deduplication. Repeated calls from channel events, reconnection, or RPC could spawn duplicate actors for the same transaction, wasting CKB RPC resources and enabling DoS (#6).

## Fix

Add `inflight_tracers: HashSet<Hash256>` to `NetworkActorState`. Both `trace_tx` and `send_tx` now check for existing tracers before spawning. Cleanup happens when tracer callbacks fire (FundingTransactionConfirmed/Failed, ClosingTransactionConfirmed/Failed).

## Verification

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -63,8 +63,8 @@ use super::types::{
     BroadcastMessageWithTimestamp, FiberMessage, ForwardTlcResult, GossipMessage, Init, OpenChannel,
 };
 use super::{
-    FiberConfig, InFlightCkbTxActor, InFlightCkbTxActorArguments, InFlightCkbTxKind,
-    ASSUME_NETWORK_ACTOR_ALIVE,
+    FiberConfig, InFlightCkbTxActor, InFlightCkbTxActorArguments, InFlightCkbTxActorMessage,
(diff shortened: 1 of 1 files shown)


doitian (approved):

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `origin/develop..5eff006d47caaf574987895f552e17560b76a1fb`

Validation:
- Focused static trace through `crates/fiber-lib/src/fiber/network.rs` and `InFlightCkbTxActor`.

Notes:
- Medium: the new tx-hash dedup set can suppress the local broadcast path for a fully signed funding transaction. In the funding-signature flow, the network actor queues `FundingTransactionPending` and then calls `trace_tx` for the same hash. Since `trace_tx` now inserts the hash first with no transaction payload, the later queued `FundingTransactionPending -> send_tx` path returns early and never starts the actor with `transaction: Some(tx)` for local broadcast/retry. Consider storing the actor ref per tx hash and upgrading/sending `InFlightCkbTxActorMessage::SendTx(tx)` when `send_tx` arrives for an existing trace-only actor; also remove the hash if actor spawn fails after insertion.

PR #1469, opened 2026-06-17
feat(watchtower): support unlocking multiple TLCs in settlement transactions

doitian:
Previously the settlement transaction builder stopped at the first TLC with a matching preimage, leaving other eligible TLCs unclaimed. This change collects all TLCs with known preimages and includes them in the settlement, unlocking them in a single transaction.

It a refactoring which reduces on-chain foot prints to save transaction fees.

### DIFF
--- a/crates/fiber-lib/src/watchtower/actor.rs
+++ b/crates/fiber-lib/src/watchtower/actor.rs
@@ -24,7 +24,7 @@ use ckb_types::{
 };
 use molecule::prelude::Entity;
 use ractor::{Actor, ActorProcessingErr, ActorRef};
-use secp256k1::{Message, SecretKey, SECP256K1};
+use secp256k1::{Message, SECP256K1};
 use strum::AsRefStr;
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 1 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `381346cf7cf39ed155801c8aa8c1b651f8d9fe5a..10c85fd53b370844d6d821c1d4a5a08d8133c522`

Validation:
- Static diff review of `crates/fiber-lib/src/watchtower/actor.rs` with supporting witness-format checks in `crates/fiber-lib/src/fiber/channel.rs`.

Notes:
- The multi-TLC settlement path appends multiple unlock records but still emits the commitment witness header byte as `0x01`; that byte is the witness `unlock_count`, so a transaction with more than one preimage unlock can be rejected or validated against a different settlement state than the Rust code hashed into the next commitment output. Please encode the actual number of unlocks and add a focused two-preimage settlement test.

doitian:
Closing in favor of issue #1470 which captures the problem statement. The diff here serves as a reference implementation. Will revisit when bandwidth permits.

PR #1489, opened 2026-06-24
Ignore spent funding cell(s) in channel gossip message 

### LINKED ISSUE #530: Ignore channel gossip message when funding cell is consumed
https://github.com/contrun/fiber/blob/f2e45420d0ec4282bc44fab2980638b6576e90b8/src/fiber/gossip.rs#L1793-L1801

We are currently using `get_transaction` rpc to fetch channel status. This may return a consumed output. It is better to use `get_live_cell` and ignore channel messages when their corresponding funding cells are consumed.

bolajahmad:
Closes #530 

## Description

Currently, during `ChannelAnnouncement`, the funding TX is fetched `[using get_transaction](https://github.com/nervosnetwork/fiber/blob/develop/crates/fiber-lib/src/ckb/client.rs#L220C4-L227C6)`. 

This approach only checks that the `Funding Outpoint` was Committed. But a Committed Outpoint does not guarantee availability of the Cell. 

This PR adds that layer of security.

- Ensure the Outpoint is Committed
- Ensure that the underlying Cell is still Live 
- No point in broadcasting a dead cell.

##  Tests

``` bash
cargo test -p fnn --all-features \
ckb::tests::test_utils::test_mock_ckb_chain_client_reports_dead_live_cell -- --exact \
 && cargo test -p fnn --all-features  \
fiber::tests::gossip::test_reject_channel_announcement_when_funding_outpoint_is_dead \ 
-- --exact
```
`cargo fmt --all -- --check`
`cargo clippy -p fnn --features rocksdb --all-targets -- -D warnings`
`git diff --cached --check`

### DIFF
--- a/crates/fiber-lib/src/ckb/client.rs
+++ b/crates/fiber-lib/src/ckb/client.rs
@@ -1,6 +1,6 @@
 use crate::ckb::config::new_ckb_rpc_async_client;
 use crate::ckb::CkbConfig;
-use ckb_jsonrpc_types::JsonBytes;
+use ckb_jsonrpc_types::{CellWithStatus, JsonBytes, OutPoint};
 use ckb_sdk::rpc::ckb_indexer::{Cell, CellType, Order, Pagination, ScriptType, SearchKey, Tx};
 use ckb_types::H256;
 
@@ -112,6 +112,11 @@ fn first_input_tx_hash(txs: &[Tx]) -> Option<H256> {
 #[async_trait::async_trait]
 pub trait CkbChainClient: Send + Sync {
     async fn get_transaction(&self, hash: H256) -> Result<GetTxResponse, anyhow::Error>;
+    async fn get_live_cell(
+        &self,
+        out_point: OutPoint,
+        with_data: bool,
+    ) -> Result<CellWithStatus, anyhow::Error>;
     async fn get_cells(
         &self,
         search_key: SearchKey,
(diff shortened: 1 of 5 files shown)


chainTe:
## Security Review

Result: No reportable security findings in the intended change.
Reviewed range: `3c25bcf16200e5d641dcd9b79f086f391e976172..8562d79ca726099fc2cb0295f10e197241a37196`

Validation:
- Targeted build validation: `cargo check -p fnn --features sqlite` fails because `crates/fiber-lib/src/fiber/gossip.rs` calls `is_funding_outpoint_live(...)`, but that function is not defined in this PR.
- `cargo check -p fnn --features rocksdb` could not reach the PR code in this local environment because RocksDB binding generation failed on a missing local Clang header (`stdbool.h`).

Notes:
- The added live-cell gate is security-positive in intent, but the helper implementation needs to be added (or the call updated) before this can be buildable and fully validated.

quake:
@bolajahmad we added `block_hash` to the rpc get_live_cell in ckb 0.208 ( https://github.com/nervosnetwork/ckb/pull/5269 ) 

I think we can use this new field to improve the existing live cell checking code.

bolajahmad:
> @bolajahmad we added `block_hash` to the rpc get_live_cell in ckb 0.208 ( [nervosnetwork/ckb#5269](https://github.com/nervosnetwork/ckb/pull/5269) )
> 
> I think we can use this new field to improve the existing live cell checking code.

If I understand correctly and based on the fixes I made, the new `block_hash` field in the response has the same effect as checking if the `cell` is Some. If anything,I wonder if it's also okay to push this change and follow up with another commit to upgrade to new `v5.1.0`. If I'm missing something, please let me know!

quake:
> > @bolajahmad we added `block_hash` to the rpc get_live_cell in ckb 0.208 ( [nervosnetwork/ckb#5269](https://github.com/nervosnetwork/ckb/pull/5269) )
> > I think we can use this new field to improve the existing live cell checking code.
> 
> If I understand correctly and based on the fixes I made, the new `block_hash` field in the response has the same effect as checking if the `cell` is Some. If anything,I wonder if it's also okay to push this change and follow up with another commit to upgrade to new `v5.1.0`. If I'm missing something, please let me know!

lgtm 👍

PR #1495, opened 2026-06-29
fix: resolve graph sync deadlock caused by message timestamp ordering

quake:
## Problem

When a ChannelUpdate has a timestamp earlier than its own ChannelAnnouncement's on-chain block timestamp, `get_broadcast_messages` returns the CU before the CA. The active sync batch processor `save_active_sync_messages` hits `has_dependencies_available() == false` on the first CU and `break`s immediately, leaving all messages stuck in the in-memory cache. The cursor never advances, causing a permanent deadlock.

## Root Cause

`ChannelAnnouncement` gets its stored timestamp from the CKB block timestamp (on-chain), while `ChannelUpdate` gets its timestamp from the node's local clock. If the CU was generated before the funding tx was confirmed, the CU sorts before its own CA in the DB ascending order.

## Fix

1. Change 4 `break` sites in `save_active_sync_messages` to `continue`:
   - `Duplicate` (cached, not in store)
   - `no_deps` (missing ChannelAnnouncement)
   - `deferred` (verification deferred)
   - `ckb_deferred` (CKB RPC deferred)

2. When any messages were skipped due to missing deps, return `Pending` instead of `Validated` so the syncing actor retries the same cursor. Once the CAs are saved (by the same batch or by Tick), the skipped CUs become stored duplicates and the cursor advances naturally.

## Tests

Added 3 unit tests covering:
- CU skipped, CA later in same batch is processed
- Cached duplicate skipped, remaining batch processed
- CU stored duplicate when CA already in store

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -2141,6 +2141,10 @@ impl<S: GossipMessageStore, C: CkbChainClient> ExtendedGossipMessageStoreState<S
         let mut latest_validated_cursor = None;
         let now_ms = now_timestamp_as_millis_u64();
 
+        let mut skipped_no_deps = 0usize;
+        let mut skipped_deferred = 0usize;
+        let mut skipped_ckb_deferred = 0usize;
+
         for message in messages {
             if !self.allow_inbound_remote_broadcast_message(peer, &message, now_ms) {
                 result = Some(ActiveSyncSaveMessagesResult::Pending);
@@ -2160,8 +2164,7 @@ impl<S: GossipMessageStore, C: CkbChainClient> ExtendedGossipMessageStoreState<S
                             continue;
                         }
                         None => {
-                            result = Some(ActiveSyncSaveMessagesResult::Pending);
-                            break;
+                            continue;
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


crates/fiber-lib/src/fiber/gossip.rs
```diff
@@ -2160,8 +2164,7 @@ impl<S: GossipMessageStore, C: CkbChainClient> ExtendedGossipMessageStoreState<S
                             continue;
                         }
                         None => {
-                            result = Some(ActiveSyncSaveMessagesResult::Pending);
-                            break;
+                            continue;
```
jjyr, line 2169: Should we count this branch as

```
skipped_deferred += 1;
```

quake (commented):
Added skipped_duplicate_cached counter and included in Pending check. Updated in 4ddb119f.

jjyr (approved):

crates/fiber-lib/src/fiber/gossip.rs
```diff
@@ -2141,6 +2141,11 @@ impl<S: GossipMessageStore, C: CkbChainClient> ExtendedGossipMessageStoreState<S
         let mut latest_validated_cursor = None;
         let now_ms = now_timestamp_as_millis_u64();
 
+        let mut skipped_no_deps = 0usize;
+        let mut skipped_deferred = 0usize;
+        let mut skipped_ckb_deferred = 0usize;
+        let mut skipped_duplicate_cached = 0usize;
```
chenyukang, line 2147: seems we can use one boolean flag to replace these counts, since there is no other usage for them, only used to check a final status.

PR #1496, opened 2026-06-29
fix: adjust CA timestamp to sort before its own CUs

quake:
## Problem

`ChannelAnnouncement` gets its stored timestamp from the CKB block confirmation time, while `ChannelUpdate` uses the node's local clock. When a CU is generated before the funding tx is confirmed on-chain, the CU timestamp can be earlier than the CA block timestamp.

In `get_broadcast_messages`, messages are sorted by timestamp ascending. This means a CU can appear before its own CA in active sync responses, triggering the dependency-ordering deadlock addressed in #1495.

## Fix

Subtract 1ms from the block timestamp when storing a `ChannelAnnouncement` via `verify_and_save_broadcast_message`. This ensures the CA always has a strictly earlier timestamp than its CUs in the DB cursor ordering, regardless of local clock skew.

The adjustment uses `saturating_sub(1)` so it is safe even when the block timestamp is 0.

## Test

Added `test_ca_adjusted_timestamp_sorts_before_cu` which:
1. Sets a known block timestamp
2. Saves a CA (gets adjusted to block_ts - 1) and its CU (timestamp = block_ts)
3. Asserts via `get_broadcast_messages` that the CA sorts before its CU

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -3273,11 +3273,13 @@ async fn verify_and_save_broadcast_message<S: GossipMessageStore>(
                         .await?;
                 let _ = verify_channel_announcement(channel_announcement, &on_chain_info, store)
                     .await?;
-                store.save_channel_announcement(
-                    on_chain_info.timestamp,
-                    channel_announcement.clone(),
-                );
-                (on_chain_info.timestamp, true)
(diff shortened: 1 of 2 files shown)


quake:
Closing: the issue is not about CA/CU timestamp ordering. As confirmed in code review, CU timestamp is always now_timestamp_as_millis_u64() which is >= the historical block timestamp. The real problem was cursor jumping, fixed in #1495.

PR #1498, opened 2026-06-29
Fix CCH incoming invoice cleanup on expiry

### LINKED ISSUE #1216: [cch]Cleanup incoming_invoice when CCH order expires to ensure resource consistency
Currently, when a CCH order expires due to expiry_delta_seconds in Fiber, only the order itself is marked as failed (CchOrder.status = Failed), but the associated incoming_invoice is not cleaned up. In business reality, if an expired order is failed but the incoming_invoice is not cancelled/failed, it can cause:
- Pending TLC or hold invoice resources remain locked
- Cooperative channel shutdown or Watchtower processes can be blocked
- Users may mistakenly attempt payments to invalid orders

Suggested improvements:
1. Add a CancelIncomingInvoice action, symmetric with the existing SettleIncomingInvoice
2. When an order enters the Failed status, proactively trigger cancel/cleanup behavior for the incoming_invoice, covering both Fiber and Lightning types
3. The scheduler expire_order logic should ideally route through the actor/state machine to ensure dispatcher actions can always be triggered (rather than a direct store update)
4. Supplement integration tests to verify that after order expiry, incoming_invoice is appropriately cancelled/released

Relevant code modules:
- crates/fiber-lib/src/cch/actions/mod.rs
- crates/fiber-lib/src/cch/actions/cancel_incoming_invoice.rs (recommended new module)
- scheduler / actor / state machine logic

Improving this cleanup mechanism will ensure consistency on failed paths and more robust resource handling.

jjyr:
## Summary
- add a CCH CancelIncomingInvoice action for safely cleaning up failed orders
- route scheduler expiry through the CCH actor so failure actions are dispatched consistently
- guard cancel paths against preimage/settlement races across CCH, Fiber actor, and RPC flows
- make settle retry idempotent when the incoming Fiber invoice is already paid

Fixes #1216

## Safety notes
- order expiry only fails Pending orders, not outgoing in-flight or outgoing-success orders
- incoming cancel/expiry events are ignored after the outgoing leg starts, so stale incoming events cannot erase an outgoing success path
- Fiber cancel_invoice rejects cancellation once a preimage exists, including the public RPC path
- final Failed orders restore cleanup before scheduling prune on startup

## Tests
- cargo nextest run -p fnn --features rocksdb test_settle_incoming_invoice_already_paid_marks_order_success
- cargo nextest run -p fnn --features rocksdb 'cch::tests'
- cargo fmt --all -- --check
- cargo clippy -p fnn --features rocksdb --all-targets -- -D warnings
- git diff --cached --check


doitian (approved):

gpBlockchain:
lgtm

PR #1503, opened 2026-06-30
fix: add CKB RPC timeout for WASM

quake:
## Problem

WASM nodes report syncing only 10+ NodeAnnouncements in 5 minutes, then stalling. The root cause is that WASM's `reqwest` `ClientBuilder` has no `timeout()` method, so CKB RPC calls (`get_transaction`, `get_block_timestamp`) can hang indefinitely during gossip CA verification.

When Tick calls `prune_messages_to_be_saved()` → `verify_and_save_broadcast_message()` → `get_channel_tx()` → CKB RPC hangs, the entire store maintenance loop is blocked. NAs (which don't need CKB RPC) get processed, but CAs stall forever.

## Fix

Add `with_ckb_rpc_timeout()` helper that wraps RPC calls with `tokio::time::timeout` on WASM only. Native continues to use `builder.timeout()` for OS-level socket timeout.

```rust
async fn with_ckb_rpc_timeout<F, T, E>(fut: F) -> Result<T, anyhow::Error> {
    #[cfg(target_arch = "wasm32")]
    { tokio::time::timeout(CKB_RPC_TIMEOUT, fut).await?? }
    #[cfg(not(target_arch = "wasm32"))]
    { fut.await.map_err(Into::into) }
}
```

Applied to both `get_transaction` and `get_block_timestamp` in `CkbRpcClient`.

### DIFF
--- a/crates/fiber-lib/src/ckb/client.rs
+++ b/crates/fiber-lib/src/ckb/client.rs
@@ -1,4 +1,6 @@
 use crate::ckb::config::new_ckb_rpc_async_client;
+#[cfg(target_arch = "wasm32")]
+use crate::ckb::config::CKB_RPC_TIMEOUT;
 use crate::ckb::CkbConfig;
 use ckb_jsonrpc_types::JsonBytes;
 use ckb_sdk::rpc::ckb_indexer::{Cell, CellType, Order, Pagination, ScriptType, SearchKey, Tx};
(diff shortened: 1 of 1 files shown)


quake:
Closing: tokio::time::timeout panics on WASM because std::time::Instant::now is not implemented. WASM's reqwest/fetch has no timeout support, and tokio timers don't work there either. WASM nodes verified all CKB RPCs succeed, so the stall is caused by something else.

PR #1504, opened 2026-06-30
fix: add CKB RPC timeout for WASM

quake:
## Problem

WASM builds panic with `std::time not implemented` because `tokio::time::timeout` calls `std::time::Instant::now()`, which is unavailable on WASM.

## Fix

Replace `tokio::time::timeout` with `tokio::select!` + `tokio::time::sleep`, which works on both WASM and native (sleep does not require `Instant::now`).

```rust
async fn with_ckb_rpc_timeout<F, T, E>(fut: F) -> Result<T, anyhow::Error> {
    #[cfg(target_arch = "wasm32")]
    {
        tokio::select! {
            result = fut => result.map_err(Into::into),
            _ = tokio::time::sleep(CKB_RPC_TIMEOUT) => {
                Err(anyhow::anyhow!("CKB RPC timed out after {:?}", CKB_RPC_TIMEOUT))
            }
        }
    }
    #[cfg(not(target_arch = "wasm32"))]
    {
        fut.await.map_err(Into::into)  // native uses builder.timeout()
    }
}
```

Applied to both `get_transaction` and `get_block_timestamp` in `CkbRpcClient`.

### DIFF
--- a/crates/fiber-lib/src/ckb/client.rs
+++ b/crates/fiber-lib/src/ckb/client.rs
@@ -1,4 +1,6 @@
 use crate::ckb::config::new_ckb_rpc_async_client;
+#[cfg(target_arch = "wasm32")]
+use crate::ckb::config::CKB_RPC_TIMEOUT;
 use crate::ckb::CkbConfig;
 use ckb_jsonrpc_types::JsonBytes;
 use ckb_sdk::rpc::ckb_indexer::{Cell, CellType, Order, Pagination, ScriptType, SearchKey, Tx};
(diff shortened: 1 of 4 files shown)


Officeyutong:
Please try https://docs.rs/tokio_with_wasm/latest/tokio_with_wasm/time/fn.timeout.html

crates/fiber-lib/src/ckb/client.rs
```diff
@@ -213,15 +215,85 @@ async fn find_first_input_tx_hash_async(
...
+    use std::pin::Pin;
+    use std::task::{Context, Poll};
+    use std::time::Duration;
+
+    use wasm_bindgen::prelude::{Closure, JsCast};
+    use wasm_bindgen_futures::JsFuture;
+
+    pub fn timeout<F>(dur: Duration, fut: F) -> WasmTimeout<F> {
+        let promise = js_sys::Promise::new(&mut |resolve, _reject| {
+            let window = web_sys::window().expect("no window");
```
Officeyutong: This may not work on `WebWorker`, there is no `window` object, only `globalThis`

PR #1506, opened 2026-06-30
Fix payment TLC expiry limit random delta

### LINKED ISSUE #1218: [cch] tlc_expiry_limit is not enforced on actual PaymentHopData expiry (random expiry delta may exceed the limit)
## Bug description

`tlc_expiry_limit` currently constrains the route-building/path-finding phase, but after PR #875 the actual outgoing TLC expiry can still exceed that limit.

PR #875 introduced a random expiry delta for privacy:
- `NetworkGraph::rand_tlc_expiry_delta()`
- applied in `NetworkGraph::build_router_from_path()` and ultimately every built PaymentHopData

However, `tlc_expiry_limit` is checked earlier only against the route's computed `incoming_tlc_expiry`, without accounting for this additional random delta.

As a result, a payment can pass route validation with `tlc_expiry_limit`, but later produce actual `PaymentHopData.expiry` values larger than the caller intended.

## Relevant code

- `src/fiber/payment.rs` - `SendPaymentDataBuilder::validate()`: only checks static limits, not accounting for random delta.
- `src/fiber/graph.rs`
  - In path-finding, `incoming_tlc_expiry > tlc_expiry_limit` returns false (does not include random delta).
  - In `build_router_from_path()`, final expiry for each hop is calculated as `now + incoming_tlc_expiry + rand_tlc_expiry_delta` (for per-hop) or `now + last_expiry_delta + rand_tlc_expiry_delta` (for final hop), where rand_tlc_expiry_delta comes from `rand_tlc_expiry_delta()`.

## Why this matters

This breaks the expected semantics of `tlc_expiry_limit`.

Callers expect `tlc_expiry_limit` to cap the full outgoing payment expiry budget, but after PR #875 it only caps the pre-randomized route computation. This risk is material for flows like CCH, which rely on `tlc_expiry_limit` for timelock safety. See also PR #1148.

## Expected behavior

`tlc_expiry_limit` should remain a hard upper bound for the actual outgoing payment expiry, even after applying privacy-related randomization (rand_tlc_expiry_delta).

## Suggested fix

One of the following (preferred: both):

1. Include worst-case random expiry delta in path-finding budget (subtract a max random slack from the limit), or
2. Add a final hard check before emitting the outgoing TLC / PaymentHopData, or
3. Both

## Regression test suggestion

Add a test where:
- `tlc_expiry_limit` is exactly equal to the route-computed expiry budget
- random expiry delta (as in PR #875) would make the final emitted expiry exceed that limit

Expected result:
- the payment should fail instead of emitting a route with effective expiry beyond `tlc_expiry_limit`

## Related PRs
- PR #875: Add a random expiry delta for payment router
- PR #1148: fix(cch): check HTLC expiry delta before sending outgoing payment

---

> Issue root cause: PR #875's logic in `NetworkGraph::build_router_from_path()` applies random expiry after all tlc_expiry_limit logic, so the final TLC expiry can be out of budget despite passing path-finding constraint. Random expiry is not accounted for or capped against the user intent!

jjyr:
## Summary

- Reserve the random TLC expiry delta in route validation so `tlc_expiry_limit` remains a hard budget for actual outgoing `PaymentHopData.expiry`.
- Add a final hard check when building payment hops to reject any actual expiry beyond `now + tlc_expiry_limit`.
- Add regression coverage for normal routing, MPP routing, trampoline routing, and explicit routers.
- Update existing tight-limit tests to account for the random expiry reserve.

## Root Cause

`tlc_expiry_limit` was checked against the route-computed `incoming_tlc_expiry`, but `build_router_from_path()` added `rand_tlc_expiry_delta` afterward. A route could pass validation and still emit outgoing hop expiries beyond the caller's intended budget.

Fixes #1218.

## Validation

- `cargo nextest run -p fnn --no-default-features --features sqlite,watchtower random_expiry`
- `cargo nextest run -p fnn --no-default-features --features sqlite,watchtower fiber::tests::graph`
- `cargo nextest run -p fnn --no-default-features --features sqlite,watchtower tlc_expiry`
- `cargo check -p fnn --no-default-features --features sqlite`

`cargo clippy -p fnn --no-default-features --features sqlite -- -D warnings` still fails on existing warnings outside this change:

- `crates/fiber-lib/src/store/store_impl/mod.rs:47` unused import `tracing::warn`
- `crates/fiber-lib/src/ckb/client.rs:155` unused function `find_first_input_tx_hash`

### DIFF
--- a/crates/fiber-lib/src/fiber/graph.rs
+++ b/crates/fiber-lib/src/fiber/graph.rs
@@ -1995,6 +1995,11 @@ where
         let mut hops_data = Vec::with_capacity(route.len() + 1);
         let last_expiry_delta =
             final_hop_expiry_delta_override.unwrap_or(payment_data.final_tlc_expiry_delta);
+
+        // `RouterHop::incoming_tlc_expiry` and `last_expiry_delta` are relative expiry
+        // budgets. `PaymentHopData::expiry` below is an absolute timestamp after adding `now`
+        // and the privacy random delta. The largest base budget decides how much room is left
+        // for the actual random delta while still honoring `tlc_expiry_limit` as a hard cap.
         let max_base_expiry_delta = route
             .iter()
             .map(|r| r.incoming_tlc_expiry)
@@ -2125,6 +2130,11 @@ where
         DEFAULT_TLC_EXPIRY_DELTA
     }
 
+    // Path finding runs before the actual random expiry delta is selected, so this helper is
+    // intentionally a deterministic reserve, not the real random value. Reserving the minimum
+    // possible random delta prevents path finding from choosing a route that already consumes the
+    // whole `tlc_expiry_limit`. `build_router_from_path` still caps the actual random delta against
+    // the selected route's remaining budget and performs the final absolute-expiry hard check.
     fn route_tlc_expiry_limit(&self, tlc_expiry_limit: u64) -> Result<u64, PathFindError> {
         let random_expiry_reserve = self.min_rand_tlc_expiry_delta();
         tlc_expiry_limit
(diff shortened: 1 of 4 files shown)


gpBlockchain:
lgtm

PR #1509, opened 2026-07-01
Fix: cache gossip channel onchain info

chenyukang:
It's only a performence optmizaton, not security issue.

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -33,7 +33,7 @@ use tentacle::{
     traits::ServiceProtocol,
     SessionId,
 };
-use tokio::sync::oneshot;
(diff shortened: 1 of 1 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `1df34625fe9c703a92c93ecce5e6b517156aa900..bad15b7bb06a93194f03dbd63267c07c6dbd6813`

Validation:
- Performed focused source tracing of the changed gossip cache path plus inbound gossip, active sync, and channel announcement validation paths.

Notes:
- Medium resource-exhaustion issue: `ChannelOnchainInfo` is inserted into the new cache before the channel announcement is fully validated, so invalid remote announcements can retain CKB output data in an entry-count-only cache. Please cache only after successful announcement validation or remove failed entries, and add a byte/TTL/LRU bound in addition to the 10,000-entry cap.

PR #1511, opened 2026-07-01
fix: retry only failed payment attempts on channel ready

chenyukang:
When a payment attempt is in `Created` status, there maybe two concrete state:
1. the ChannelActor haven't processed SendPaymentCommand
2. the ChannelActor already processed AddTlc, or there is already a retrying tlc task in queue.

So we can not just Resend the attempt with `Created` status.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -9643,16 +9643,17 @@ impl ChannelActorState {
     pub(crate) fn owns_payment_attempt(&self, payment_hash: Hash256, attempt_id: u64) -> bool {
         let attempt_id = Some(attempt_id);
 
-        self.tlc_state
-            .all_tlcs()
-            .any(|tlc| tlc.payment_hash == payment_hash && tlc.attempt_id == attempt_id)
-            || self.retryable_tlc_operations.iter().any(|operation| {
-                matches!(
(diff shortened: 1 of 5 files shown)


crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -9640,6 +9640,21 @@ impl ChannelActorState {
         !self.retryable_tlc_operations.is_empty()
     }
 
+    pub(crate) fn owns_payment_attempt(&self, payment_hash: Hash256, attempt_id: u64) -> bool {
+        let attempt_id = Some(attempt_id);
+
+        self.tlc_state
+            .all_tlcs()
```
jjyr: [P2 — terminal TLCs can strand Created payments](https://github.com/nervosnetwork/fiber/blob/b657cb1139a9b08e03d11ee3acaa9bab8e4bbd11/crates/fiber-lib/src/fiber/channel.rs#L9646-L9648): all_tlcs() includes removed TLCs whose removal has already been applied. If the process crashes after persisting that terminal state but before the payment actor processes its result event, the attempt remains Created. On reconnect, this predicate suppresses retry even though the TLC cannot produce another event, leaving the payment stuck indefinitely. Ownership should exclude TLCs with AppliedFlags::REMOVE, or the outcome must be durably replayed.

PR #1517, opened 2026-07-04
Fix WASM DB worker IPC synchronization over SharedArrayBuffer

Officeyutong:
## Summary

Fixes a WASM runtime panic in `fiber-wasm-db-worker` caused by fragile IPC synchronization between `fiber-wasm` and `fiber-wasm-db-worker`.

The IPC protocol uses `SharedArrayBuffer` with `Atomics.wait/notify`, but the command words were previously read and written with normal `Int32Array` accessors. This could let the DB worker observe an unexpected command state, such as `Waiting` or stale `ResponseTakeWhile`, and hit `unreachable!()` in the main loop.

## Changes

- Add atomic IPC helpers in `fiber-wasm-db-common` for command word load/store.
- Use atomic stores when publishing commands and clearing command states.
- Use atomic loads when reading command states.
- Make DB worker wait helpers loop until the command state actually changes.
- Make browser-side DB response waits tolerate wakeups where the state is still `Waiting`.
- Avoid panicking on stale top-level `ResponseTakeWhile`; log and continue instead.

## Verification

- `cargo fmt --all`
- `cargo check -p fiber-wasm-db-common -p fiber-wasm-db-worker -p fiber-store --target wasm32-unknown-unknown`
- `cargo clippy -p fiber-wasm-db-common -p fiber-wasm-db-worker -p fiber-store --target wasm32-unknown-unknown -- -D warnings`

### DIFF
--- a/crates/fiber-store/src/browser.rs
+++ b/crates/fiber-store/src/browser.rs
@@ -1,7 +1,9 @@
 use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
+use fiber_wasm_db_common::load_command;
 use fiber_wasm_db_common::read_command_payload;
+use fiber_wasm_db_common::store_command;
 use fiber_wasm_db_common::write_command_with_payload;
 use fiber_wasm_db_common::DbCommandRequest;
 use fiber_wasm_db_common::DbCommandResponse;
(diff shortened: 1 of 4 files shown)


gpBlockchain:
lgtm

PR #1520, opened 2026-07-06
Avoid supervisor panic on actor failure

quake:
## Summary
- log linked child actor failures instead of re-panicking from supervisors
- apply the behavior consistently to root, network, and gossip supervisors
- add regression coverage that a linked child actor failure does not stop the root actor

## Tests
- cargo nextest run -p fnn --features sqlite root_actor_does_not_panic_when_linked_child_fails
- cargo fmt --all -- --check

### DIFF
--- a/crates/fiber-lib/src/actors.rs
+++ b/crates/fiber-lib/src/actors.rs
@@ -1,6 +1,10 @@
 use ractor::{Actor, ActorProcessingErr, ActorRef, SupervisionEvent};
 use tokio_util::{sync::CancellationToken, task::TaskTracker};
-use tracing::debug;
+use tracing::{debug, error};
+
+pub(crate) fn log_actor_failed(who: impl std::fmt::Debug, err: impl std::fmt::Debug) {
+    error!("Actor unexpectedly panicked (id: {:?}): {:?}", who, err);
+}
 
 /// A root actor that listens for cancellation token and stops all sub actors (those who started by spawn_linked).
 pub struct RootActor;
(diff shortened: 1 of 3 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `bc361aaaa40d1394b83e6a1808869b0b06c48c13..e341db7b1b607681d9089187449f07af5d1cf9ff`

Validation:
- Reviewed the changed supervisor handlers in `actors.rs`, `fiber/gossip.rs`, and `fiber/network.rs`, plus directly supporting actor lifecycle, channel cleanup, payment cleanup, and recovery routing paths.
- Verified `ractor` 0.15.13 skips `post_stop` when an actor self-panics.

Finding:
- Medium: `NetworkActor::handle_supervisor_evt` now logs `ActorFailed` without cleanup or respawn. If a linked channel or payment actor self-panics, ractor skips that child actor's `post_stop`, so Fiber does not emit the existing `ChannelActorStopped` / `PaymentActorStopped` cleanup events. The stale actor ref can remain registered, causing reconnect/reestablish or payment-resume messages to be sent to a dead actor and dropped until manual restart.

Recommended fix: when `NetworkActor` receives `ActorFailed`, map the failed child to its channel id / payment hash and run the same cleanup as the stopped-event paths, or respawn from persisted state where safe. At minimum, remove stale channel/payment actor entries so recovery can proceed.

Notes:
- The PR direction is good: avoiding supervisor panic reduces blast radius. The missing control is actor-class-specific cleanup/escalation after failure.
- Critical root-supervised singleton actors should also have a fail-fast, respawn, or health signal strategy; logging-only failure handling is a residual lifecycle risk.

PR #1522, opened 2026-07-06
Enforce delay for follow-up watchtower settlements

quake:
## Summary
- require the full commitment delay before follow-up final-balance settlements
- apply the readiness gate to both local and remote follow-up settlement branches
- add regression coverage that a non-first final-balance settlement returns `Ok(None)` before the full delay

## Tests
- cargo nextest run -p fnn --features sqlite non_first_final_balance_waits_full_commitment_delay watchtower::actor::tests
- cargo fmt --all -- --check
- git diff --check


quake:
Not a vulnerability. Revocation only applies to the funding output; once the first settlement consumes the funding cell, follow-up settlements cannot be revoked. The missing delay check on follow-up settlements is contract-design-consistent and poses no additional risk.

PR #1523, opened 2026-07-06
Disable standalone watchtower RPC key export

quake:
## Summary
- reject `standalone_watchtower_rpc_url` at startup because the current watchtower RPC payload exports settlement private keys
- apply the guard to native and WASM startup paths before network actors are started
- add config-level regression coverage for rejecting standalone watchtower RPC while allowing the built-in watchtower config

## Tests
- cargo nextest run -p fnn --features sqlite standalone_watchtower_rpc_url_is_rejected built_in_watchtower_config_is_allowed
- cargo check -p fiber-bin --features sqlite
- cargo check -p fiber-wasm --target wasm32-unknown-unknown
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/fiber-bin/src/main.rs
+++ b/crates/fiber-bin/src/main.rs
@@ -216,6 +216,10 @@ async fn run_node(
 
             info!("Starting fiber");
 
+            fiber_config
+                .validate_standalone_watchtower_rpc()
+                .map_err(ExitMessage)?;
+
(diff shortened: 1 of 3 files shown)


quake:
Watchtower is designed to be a trusted component; the settlement private key export via standalone watchtower RPC is by-design behavior. No fix needed.

PR #1530, opened 2026-07-06
Do not force-close on invalid CommitmentSigned

quake:
## Summary
- remove force-close on malformed CommitmentSigned verification failure
- a peer-crafted invalid CommitmentSigned should not trigger an on-chain force-close

## Tests
- cargo fmt --all -- --check
- git diff --check


quake:
Not a bug - force-close on invalid CommitmentSigned is intentional behavior to protect channel state consistency.

PR #1531, opened 2026-07-06
Cap default shutdown fee to funding fee rate

quake:
## Summary
- cap the default cooperative shutdown fee to `funding_fee_rate` when no explicit fee is provided
- a peer-controlled excessive `commitment_fee_rate` should not dictate the shutdown fee

## Tests
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -2297,9 +2297,10 @@ where
                 }
             };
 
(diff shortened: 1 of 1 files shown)


quake:
Not a vulnerability. commitment_fee_rate is a mutually negotiated parameter agreed upon during channel opening. Using it as the default shutdown fee is by-design behavior.

PR #1532, opened 2026-07-06
Ban peers after repeated invalid channel messages

quake:
## Summary
- track per-peer count of messages targeting non-existent channels
- disconnect and ban peer after 20 repeated invalid messages
- cleanup counter on normal peer disconnect

## Tests
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -5380,7 +5380,7 @@ where
                 }
                 return;
             }
-            // Ban expired, clean up
+            // Ban expired — clean up all ban state and resume normal reconnect.
             self.banned_peers.remove(&remote_pubkey);
             self.requested_disconnect_peers.remove(&remote_pubkey);
         }
@@ -5477,7 +5477,6 @@ where
         }
 
         self.peer_session_map.remove(&pubkey);
-        self.invalid_channel_msg_count.remove(&pubkey);
         if let Some(channel_ids) = self.peer_channel_index.get_channels(&pubkey) {
             for channel_id in channel_ids {
                 if let Some(channel) = self.channels.get(&channel_id) {
@@ -5517,8 +5516,8 @@ where
     async fn disconnect_and_ban_peer(&mut self, pubkey: Pubkey) {
         // Don't remove from peer_session_map — let the normal disconnect flow
         // handle cleanup (channel actor notifications, pending records, etc.).
-        // Record the ban with expiry so the peer cannot reconnect for a cooldown.
-        self.invalid_channel_msg_count.remove(&pubkey);
+        // Don't clear invalid_channel_msg_count — preserve violation state
+        // across voluntary reconnects.
         self.requested_disconnect_peers.insert(pubkey);
         let ban_until = now_timestamp_as_millis_u64().saturating_add(10 * 60 * 1000);
         self.banned_peers.insert(pubkey, ban_until);
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -5382,6 +5382,7 @@ where
             }
             // Ban expired — clean up all ban state and resume normal reconnect.
             self.banned_peers.remove(&remote_pubkey);
+            self.invalid_channel_msg_count.remove(&remote_pubkey);
             self.requested_disconnect_peers.remove(&remote_pubkey);
         }
 
@@ -5477,6 +5478,11 @@ where
         }
 
         self.peer_session_map.remove(&pubkey);
+        // Keep the counter only if the peer was recently banned, to prevent
+        // reconnection bypass. Otherwise clean up to avoid unbounded growth.
+        if !self.banned_peers.contains_key(&pubkey) {
+            self.invalid_channel_msg_count.remove(&pubkey);
+        }
         if let Some(channel_ids) = self.peer_channel_index.get_channels(&pubkey) {
             for channel_id in channel_ids {
                 if let Some(channel) = self.channels.get(&channel_id) {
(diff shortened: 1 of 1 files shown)


chenyukang (commented):
I found a few issues in this patch:

1. `send_message_to_channel_actor` does not cover the normal remote-invalid-channel path. In `handle_peer_message`, the `!found` branch for `FiberMessage::ChannelNormalOperation` returns `Error::ChannelNotFound` before calling `send_message_to_channel_actor`, so a peer can keep sending channel messages for arbitrary non-existing channels without incrementing the new counter or being disconnected. The counter needs to live in that `!found` branch, or that branch needs to route through the same policy.

2. `disconnect_and_ban_peer` resets `invalid_channel_msg_count` on disconnect and does not actually ban the peer. Adding the pubkey to `requested_disconnect_peers` only suppresses our own auto-reconnect; it does not reject inbound reconnects, and the count starts over from zero. A malicious peer can reconnect and repeat 20 invalid messages indefinitely. This needs a real temporary deny/ban entry, or a counter with expiry that survives disconnects.

3. `disconnect_and_ban_peer` removes the peer from `peer_session_map` before calling `control.disconnect`. When the real `PeerDisconnected` event arrives, `on_peer_disconnected` sees an unknown session and returns early, so existing channel actors are not notified with `ChannelEvent::PeerDisconnected`, pending inbound channel-open records are not cleaned, and normal reconnect bookkeeping is skipped. This should follow the existing `DisconnectPeer` command pattern: request the disconnect but let the disconnect event remove the session and run cleanup.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -5990,6 +6002,21 @@ where
                     }
                 }
                 (message, _) => {
+                    if let Some(pubkey) = peer_pubkey {
```
chenyukang, line 6051: This counter is only reached after `send_message_to_channel_actor` is called, but the normal remote-invalid-channel path returns earlier in `handle_peer_message`: when `FiberMessage::ChannelNormalOperation` has `!found`, it returns `Error::ChannelNotFound` before reaching this helper. A peer can therefore keep sending messages for arbitrary non-existing channels without incrementing this counter. The counting/ban policy needs to run in that `!found` branch, or that branch needs to route through this same policy.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -5477,6 +5479,16 @@ where
         self.seed_peer_reconnect_backoff_if_needed(&peer_id, PeerReconnectTrigger::Disconnected);
     }
 
+    async fn disconnect_and_ban_peer(&mut self, pubkey: Pubkey) {
+        if let Some(peer) = self.peer_session_map.remove(&pubkey) {
+            self.invalid_channel_msg_count.remove(&pubkey);
```
chenyukang: Resetting `invalid_channel_msg_count` here means this is not really a ban. `requested_disconnect_peers` only suppresses our own auto-reconnect; it does not reject inbound reconnects, and after reconnect the peer gets a fresh counter. A malicious peer can reconnect and repeat the 20 invalid messages indefinitely. This needs a real temporary deny/ban entry, or a counter with expiry that survives disconnects.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -5477,6 +5479,16 @@ where
         self.seed_peer_reconnect_backoff_if_needed(&peer_id, PeerReconnectTrigger::Disconnected);
     }
 
+    async fn disconnect_and_ban_peer(&mut self, pubkey: Pubkey) {
+        if let Some(peer) = self.peer_session_map.remove(&pubkey) {
```
chenyukang: Removing the peer from `peer_session_map` before calling `control.disconnect` bypasses the normal disconnect cleanup. When the actual `PeerDisconnected` event arrives, `on_peer_disconnected` sees an unknown session and returns early, so channel actors are not notified with `ChannelEvent::PeerDisconnected`, pending inbound channel-open records are not cleaned, and reconnect bookkeeping is skipped. This should follow the existing `DisconnectPeer` command pattern: request the disconnect, then let the disconnect event remove the session and run cleanup.

chenyukang (commented):
Inline notes for the issues I mentioned in the earlier review summary.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -5441,6 +5477,7 @@ where
         }
 
         self.peer_session_map.remove(&pubkey);
+        self.invalid_channel_msg_count.remove(&pubkey);
```
chenyukang: [P1] Preserve invalid-message strikes across voluntary reconnects

This removes the per-peer counter on every normal disconnect, but the ban is only created when the counter becomes greater than 20. A peer can therefore send exactly 20 invalid-channel messages, disconnect voluntarily, reconnect with a fresh counter, and repeat indefinitely without ever entering `banned_peers`. Please retain the violation state across disconnects until an explicit expiry window, and add a regression test covering 20 invalid messages, disconnect, reconnect, then one more invalid message.

crates/fiber-lib/src/fiber/network.rs
```diff
@@ -5349,6 +5366,25 @@ where
...
+                    remote_pubkey,
+                    ban_until.saturating_sub(now_timestamp_as_millis_u64())
+                );
+                if let Err(err) = self.control.disconnect(session.id).await {
+                    error!("Failed to disconnect banned peer: {:?}", err);
+                }
+                return;
+            }
+            // Ban expired, clean up
+            self.banned_peers.remove(&remote_pubkey);
```
chenyukang, line 5384: [P2] Expire temporary bans independently of an inbound reconnect

Expired entries are removed only when the same peer reaches `on_peer_connected`. Banning also inserts the peer into `requested_disconnect_peers`, so all local reconnect paths skip it. If the remote peer does not initiate another connection after the ten-minute deadline, no code removes either entry or resumes local reconnects: the temporary ban effectively becomes permanent for outbound connectivity, and peers that never return leave stale entries in both maps indefinitely. Please schedule expiry or purge expired bans during maintenance, remove the requested-disconnect marker, and resume reconnect backoff where appropriate.

PR #1534, opened 2026-07-06
Cleanup gossip peer outbound state on disconnect

quake:
## Summary
- remove per-peer outbound token bucket on peer disconnect
- replace state clone in estimate_outbound_message_delay with non-mutating peek
- prevents unbounded peer_outbound HashMap growth

## Tests
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/fiber-lib/src/fiber/gossip.rs
+++ b/crates/fiber-lib/src/fiber/gossip.rs
@@ -3969,6 +3969,7 @@ where
             }
             GossipActorMessage::PeerDisconnected(pubkey, _session) => {
                 state.peer_states.remove(&pubkey);
+                state.policy.remove_outbound_peer(&pubkey);
                 state.drop_delayed_outbound_messages_for_peer(
                     &pubkey,
                     now_timestamp_as_millis_u64(),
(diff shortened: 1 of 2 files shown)


crates/fiber-lib/src/fiber/gossip.rs
```diff
@@ -3969,6 +3969,7 @@ where
             }
             GossipActorMessage::PeerDisconnected(pubkey, _session) => {
                 state.peer_states.remove(&pubkey);
+                state.policy.remove_outbound_peer(&pubkey);
```
chenyukang: here removes the peer outbound bucket before dropping delayed outbound messages, but    `drop_delayed_outbound_messages_for_peer` refunds each queued message, and refund_outbound_message recreates the same peer bucket via `peer_outbound.entry(...).or_insert_with(...)`. So any disconnected peer with delayed gossip payloads still leaves an outbound bucket behind, which means this PR does not actually clean up the peer state in the case most relevant to queue/backpressure cleanup. Move remove_outbound_peer after the delayed-message refund/drop, or add a refund path that does not recreate per-peer state during disconnect cleanup.

jjyr (approved):

PR #1535, opened 2026-07-07
Reconstruct commitment tx on demand instead of persisting

quake:
## Summary
- remove persisted storage of `latest_commitment_transaction` from `ChannelActorData`
- reconstruct commitment transaction from state fields on demand during force-close via `build_commitment_tx_and_settlement_data`
- field kept with `#[serde(default)]` for backward compat with old persisted data

## Why
The commitment transaction is fully determined by state fields already persisted (funding outpoint, commitment numbers, TLCs, settlement amounts, delay epoch, pubkeys). Storing it redundantly costs `~500B per channel` and incurs deserialization overhead that slows down unrelated state reads (e.g., preimage cleanup scanning TLCs across channels).

## Tests
- cargo nextest run -p fnn --features sqlite test_tx_signatures_after_channel_ready_rejected
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -7815,7 +7815,6 @@ impl ChannelActorState {
             }
         }
         self.commit_remote_nonce(commitment_signed.next_commitment_nonce);
-        self.latest_commitment_transaction = Some(commitment_tx.data());
         Ok(true)
     }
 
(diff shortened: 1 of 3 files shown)


quake:
Cannot remove this field: the commitment transaction contains MuSig2 aggregated signature witnesses that cannot be reconstructed from state. The peer's partial signature is only available during the CommitmentSigned exchange and is not persisted. Closing.

PR #1536, opened 2026-07-07
Store commitment tx witnesses instead of full transaction

quake:
## Summary
- add `latest_commitment_tx_witnesses` field to store only the witness data (~81 bytes) instead of the full commitment transaction (~500 bytes)
- commitment tx body (inputs, outputs) is reconstructed from state fields on demand
- full transaction field kept for backward compat; new code prefers the lightweight witnesses field
- no migration needed

## Tests
- cargo nextest run -p fnn --features sqlite test_tx_signatures_after_channel_ready_rejected test_abort_funding_on_sign_funding_tx_failure test_abort_funding_on_committing_funding_tx_on_chain
- cargo fmt --all -- --check
- git diff --check


quake:
Cannot rebuild commitment tx from state: clean_up_failed_tlcs() runs after tx construction, changing tlc_state. The stored witness signs a tx whose output references pre-cleanup TLCs, but rebuilding from current state produces a different output with different hash. Closing.

Issue #1497, opened 2026-06-29
CCH restart does not recover LND outgoing success/preimage after downtime


gpBlockchain:
## Summary

A CCH `send_btc` order can remain stuck in `OutgoingInFlight` if the LND outgoing payment succeeds while CCH is down.

In this case LND has already completed the outgoing payment and learned the preimage, but after CCH restarts the order does not recover the terminal LND payment state, does not transition to `OutgoingSuccess`, and never settles the Fiber incoming invoice.

## Impact

The cross-chain swap can get stuck after the BTC/Lightning outgoing leg has succeeded. The Fiber incoming invoice remains unsettled even though the preimage is available in LND. The payer-side Fiber payment stays inflight and can eventually fail/stall while CCH has already paid the outgoing BTC invoice.

## Reproducer

The following regression test waits until the order reaches `OutgoingInFlight`, stops CCH/Fiber node1, settles the destination LND hold invoice while CCH is down, verifies the LND invoice is `SETTLED`, then restarts CCH and waits for the order to become `Success`.

```python
def test_cch_r004_send_btc_recovers_lnd_success_during_cch_downtime(self):
    """CCH-R004 / CCH-T006.

    If LND outgoing payment succeeds while CCH is down, restart must recover
    the terminal LND payment/preimage and settle the Fiber incoming invoice.
    """
    self.open_wrapped_btc_channel_to_cch()

    preimage = self.generate_random_preimage()
    payment_hash = sha256_hex(preimage)
    lnd_invoice = self.LNDs[1].addholdinvoice(
        payment_hash.replace("0x", ""),
        1000,
        "CCH-R004 send_btc restart recovery",
    )
    order = self.fiber1.get_client().send_btc(
        {
            "btc_pay_req": lnd_invoice["payment_request"],
            "currency": "Fibd",
        }
    )
    assert order["payment_hash"] == payment_hash

    payment = self.fiber2.get_client().send_payment(
        {"invoice": order["incoming_invoice"]["Fiber"]}
    )
    self.wait_cch_order_state(
        self.fiber1, payment_hash, "OutgoingInFlight", timeout=120
    )

    self.fiber1.stop()
    self.LNDs[1].ln_cli_with_cmd(f"settleinvoice {preimage.replace('0x', '')}")
    wait_lnd_invoice_state(self.LNDs[1], payment_hash, "SETTLED")
    self.restart_cch()

[...]

PR #1546, opened 2026-07-10
fix: recover CCH LND payments after restart

### LINKED ISSUE #1501: CCH send_btc can stay OutgoingInFlight if LND completes outgoing payment during tracker downtime
## Summary

A CCH `send_btc` order can stay stuck in `OutgoingInFlight` if the CCH-side LND node restarts while the outgoing LND payment is in flight. LND later resumes and completes the payment, but CCH does not recover the terminal payment/preimage, so the Fiber incoming side is not reliably finalized from the order state machine.

This reproduces with commit `15399ac` (`fnn Fiber v0.9.0-rc4`, built 2026-06-26). A patched binary that starts a per-payment LND tracker for the in-flight outgoing payment passes the same test.

## Impact

The outgoing BTC/LND payment may already be successful, but the CCH order remains non-terminal. Clients polling `get_cch_order` keep seeing `OutgoingInFlight` instead of `Success`.

## Reproduction Test

```python
from cch_restart_helpers import CchRestartBase, sha256_hex, wait_lnd_invoice_state


class TestCchLndRestart(CchRestartBase):
    def test_cch_r201_lnd_sender_restarts_during_outgoing_inflight(self):
        """CCH-R201.

        CCH's LND node restarts while a send_btc outgoing payment is in-flight.
        After LND comes back and the hold invoice is settled, CCH should recover
        the stream and finish the Fiber side.
        """
        self.open_wrapped_btc_channel_to_cch()

        preimage = self.generate_random_preimage()
        payment_hash = sha256_hex(preimage)
        lnd_invoice = self.LNDs[1].addholdinvoice(
            payment_hash.replace("0x", ""),
            1000,
            "CCH-R201 LND restart during outgoing inflight",
        )
        order = self.fiber1.get_client().send_btc(
            {
                "btc_pay_req": lnd_invoice["payment_request"],
                "currency": "Fibd",
            }
        )
        payment = self.fiber2.get_client().send_payment(
            {"invoice": order["incoming_invoice"]["Fiber"]}
        )
        self.wait_cch_order_state(
            self.fiber1, payment_hash, "OutgoingInFlight", timeout=120
        )

        self.LNDs[0].stop()
        self.LNDs[0].start()
        self.LNDs[1].ln_cli_with_cmd(f"settleinvoice {preimage.replace('0x', '')}")

        wait_lnd_invoice_state(self.LNDs[1], payment_hash, "SETTLED")
        self.wait_cch_order_state(self.fiber1, payment_hash, "Success", timeout=180)
        self.wait_payment_state(self.fiber2, payment["payment_hash"], "Success")
```

## Observed Failure

With commit `15399ac` (`fnn Fiber v0.9.0-rc4`, built 2026-06-26), the order remains stuck in `OutgoingInFlight`:

```text
TimeoutError: payment:0x5c21325ef5344d42abceebfefd26f597ab2099303e8fb298cd7e3b0d16326513 status did not reach state: OutgoingInFlight, expected:Success , within timeout period.
```

The LND side can complete the payment after restart, for example:

```text
CRTR: Resuming payment <payment_hash>
SRVR: Adding preimage=<preimage> to witness cache for <payment_hash>
CRTR: Resumed payment <payment_hash> completed
```

But CCH does not advance the order from `OutgoingInFlight` to `Success`.

## Expected Behavior

After LND restarts and completes the outgoing payment, CCH should recover the terminal LND payment state/preimage and settle the Fiber incoming invoice. `get_cch_order` should eventually return `Success`.

## Likely Root Cause

CCH currently relies on the global LND payment tracking stream. When LND restarts, that stream is interrupted with an error such as:

```text
Error tracking LND payments, retry 15 seconds later: status: Unknown, message: "routerrpc server shutting down"
```

After reconnect, the global stream does not reliably replay the already terminal outgoing payment/preimage for the in-flight CCH order. CCH therefore misses the terminal `Success` event and keeps the order in `OutgoingInFlight`.

A patched build that starts a per-payment `track_payment_v2(payment_hash)` tracker when the outgoing LND payment enters `InFlight` recovers the terminal `Succeeded` update after LND restart and passes this test.

jjyr:
## Summary

- track each active outgoing Lightning payment by payment hash with `TrackPaymentV2`
- retry the per-payment stream after LND or connection failures
- deduplicate active trackers and clean them up on terminal payment or order states
- remove the redundant global `TrackPayments` stream

Fixes #1501 #1497 .

## Problem

CCH previously relied on LND's global `TrackPayments` stream after the first
`SendPaymentV2` update. If LND restarted while a payment was in flight and the
payment reached a terminal state before CCH reconnected, the new global stream
did not reliably replay that terminal update. The CCH order could therefore
remain in `OutgoingInFlight`, without the preimage needed to settle the incoming
Fiber payment.

## Solution

Outgoing Lightning orders now start an idempotent per-payment tracker using
`TrackPaymentV2(payment_hash)`. This RPC returns the requested payment's current
state after reconnecting, including a terminal state reached while LND or CCH
was unavailable.

The tracker:

- emits every payment update through the existing CCH tracking event path
- retries transient connection and stream failures
- exits after `Succeeded` or `Failed`
- uses per-tracker IDs so a stale completion cannot remove a newer tracker
- is cancelled when the CCH order reaches a final state

The former global `TrackPayments` stream is removed because every active CCH
outgoing Lightning payment now has an explicit per-hash tracker. Outgoing Fiber
payments continue to use the existing store-change tracking path. Persisted
active orders resume per-payment tracking through the existing startup action
scheduling.

## Testing

- TDD regression test added before the implementation
- focused dispatcher suite — 39 passed
- focused LND tracker suite — 9 passed
- `cargo nextest run -p fnn -p fiber-bin cch::` — 131 passed
- `git diff --check`

A live LND restart E2E was not run as part of this change; reconnect request
formation, deduplication, terminal handling, cancellation, and lifecycle cleanup
are covered by focused tests.

### DIFF
--- a/crates/fiber-lib/src/cch/trackers/lnd_trackers.rs
+++ b/crates/fiber-lib/src/cch/trackers/lnd_trackers.rs
@@ -12,20 +12,23 @@
 //!
 //! ## Architecture
 //!
-//! The actor uses a message-passing model with two main message types:
+//! Tracking requests are submitted through these message types:
 //! - `TrackInvoice(Hash256)`: Adds invoice to tracking queue
 //! - `TrackPayment(Hash256)`: Starts an idempotent per-payment tracker
-//! - `InvoiceTrackerCompleted{...}`: Sent by spawned tracker tasks when they finish
 //!
-//! When a tracker task completes (successfully or with error), it ALWAYS sends
-//! `InvoiceTrackerCompleted` back to the actor. The actor maintains two data structures:
+//! Spawned tasks send completion messages back to the actor for lifecycle cleanup.
+//!
+//! When an invoice tracker task completes (successfully or with error), it ALWAYS sends
+//! `InvoiceTrackerCompleted` back to the actor. Invoice scheduling uses two data structures:
 //! - `invoice_queue`: VecDeque of pending invoice hashes
 //! - `active_invoice_trackers`: Number of active invoice trackers
 //!
-//! When completion message arrives:
+//! When an invoice completion message arrives:
 //! 1. Decrement `active_invoice_trackers` counter
 //! 2. Re-queue if failed
 //! 3. Dequeue invoices from the queue and start tracking
+//!
+//! Active payment trackers are keyed by payment hash and removed when their task completes.
 
 use std::{
     collections::{HashMap, VecDeque},
@@ -171,7 +174,8 @@ struct ActivePaymentTracker {
 /// It provides the following features:
 ///
 /// ## Payment Tracking
-/// - Automatically tracks all LND payments in the background
+/// - Tracks explicitly requested outgoing payments by payment hash
+/// - Deduplicates active trackers and reconnects until LND reports a terminal status
 /// - Sends `CchTrackingEvent::PaymentChanged` events to the output port
 ///
 /// ## Invoice Tracking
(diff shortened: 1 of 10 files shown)


gpBlockchain:
lgtm

crates/fiber-lib/src/cch/trackers/lnd_trackers.rs
```diff
@@ -297,6 +335,76 @@ impl Actor for LndTrackerActor {
...
+                "Outgoing payment tracker already active for payment_hash={}",
+                payment_hash
+            );
+            return;
+        }
+
+        let tracker_id = self.next_payment_tracker_id;
+        self.next_payment_tracker_id = self.next_payment_tracker_id.wrapping_add(1);
+        let token = self.token.child_token();
+        self.active_payment_trackers.insert(
```
chenyukang: [P1] Bound the number of per-payment LND trackers. Every distinct payment hash inserted here can spawn a long-lived task and `TrackPaymentV2` gRPC stream; stream errors retry forever every 15 seconds, while accepted and in-flight orders are not expired by the scheduler. Deduplication limits duplicate hashes only, so many different hashes can grow tasks, streams, memory, and retry work without bound. Please add global admission or capacity control, or a bounded queue, while preserving restart recovery.

PR #1547, opened 2026-07-10
fix: report CCH receive amount including fee

### LINKED ISSUE #1499: CCH receive_btc amount_sats excludes fee while Lightning hold invoice charges amount + fee
## Summary

`receive_btc` returns and stores `amount_sats` as the principal amount, but the generated Lightning hold invoice charges `amount_sats + fee_sats`. This contradicts the API schema, which documents `CchOrderResponse.amount_sats` as "including fee", and makes `get_cch_order` report an amount lower than the actual Lightning invoice clients must pay.

## Impact

Clients that call `receive_btc` or poll `get_cch_order` see `amount_sats = 1000`, while the Lightning invoice they are expected to pay is `1500` sats when the CCH base fee is `500` sats. This can cause UI/API accounting mismatches and under-report the amount required for the incoming Lightning leg.

## Code pointers

- `crates/fiber-json-types/src/cch.rs`: `CchOrderResponse.amount_sats` is documented as `Amount required to pay in Satoshis, including fee`.
- `crates/fiber-json-types/src/convert.rs`: conversion passes the stored `order.amount_sats` directly into the JSON response.
- `crates/fiber-lib/src/cch/actor.rs`: `send_btc` stores the generated invoice amount as `amount + fee`.
- `crates/fiber-lib/src/cch/actor.rs`: `receive_btc` creates the LND hold invoice from `(amount_sats + fee_sats) * 1000`, but stores only `amount_sats` in `CchOrder`.

## Reproducer

This regression test was added in `fiber-py-integration-test` under `test_cases/fiber/cch/test_cch_amount_contract_from_analysis.py`:

```python
import pytest

from framework.basic_fiber_with_cch import FiberCchTest


class CchAnalysisBase(FiberCchTest):
    def restart_cch(self, extra_config=None):
        config = {
            "cch": True,
            "cch_lnd_cert_path": f"{self.LNDs[0].tmp_path}/tls.cert",
            "cch_lnd_rpc_url": f"https://localhost:{self.LNDs[0].rpc_port}",
        }
        if extra_config:
            config.update(extra_config)
        self.fiber1.stop()
        self.fiber1.prepare(update_config=config)
        self.fiber1.start()


class TestCchAmountContractFromAnalysis(CchAnalysisBase):
    def _create_receive_btc_order_with_base_fee(self, amount_sats=1000, fee_sats=500):
        self.restart_cch(
            {
                "cch_base_fee_sats": fee_sats,
                "cch_fee_rate_per_million_sats": 0,
            }
        )
        invoice = self.fiber2.get_client().new_invoice(
            {
                "amount": hex(amount_sats),
                "currency": "Fibd",
                "description": "CCH-T004 receive_btc fee observable",
                "udt_type_script": self.get_account_udt_script(
                    self.fiber1.account_private
                ),
                "payment_preimage": self.generate_random_preimage(),
                "hash_algorithm": "sha256",
            }
        )
        order = self.fiber1.get_client().receive_btc(
            {"fiber_pay_req": invoice["invoice_address"]}
        )
        decoded = self.LNDs[0].ln_cli_with_cmd(
            f"decodepayreq {order['incoming_invoice']['Lightning']}"
        )
        return order, decoded

    def test_cch_t004_receive_btc_response_and_get_order_amount_sats_include_fee(self):
        amount_sats = 1000
        fee_sats = 500
        expected_total_sats = amount_sats + fee_sats
        order, decoded = self._create_receive_btc_order_with_base_fee(
            amount_sats=amount_sats,
            fee_sats=fee_sats,
        )
        stored_order = self.fiber1.get_client().get_cch_order(
            {"payment_hash": order["payment_hash"]}
        )

        assert int(decoded["num_satoshis"]) == expected_total_sats
        assert {
            "receive_btc_response": int(order["amount_sats"], 16),
            "get_cch_order": int(stored_order["amount_sats"], 16),
        } == {
            "receive_btc_response": expected_total_sats,
            "get_cch_order": expected_total_sats,
        }
```

## Observed failure

```text
E       AssertionError: assert {'receive_btc_response': 1000, 'get_cch_order': 1000} == {'receive_btc_response': 1500, 'get_cch_order': 1500}
E         Differing items:
E         {'get_cch_order': 1000} != {'get_cch_order': 1500}
E         {'receive_btc_response': 1000} != {'receive_btc_response': 1500}
```

The decoded Lightning invoice amount is `1500`, so the incoming Lightning leg is charging `amount + fee`, but both the direct `receive_btc` response and `get_cch_order` expose only `1000`.

## Expected behavior

For `receive_btc`, `amount_sats` in both the direct response and persisted `get_cch_order` response should follow the schema and report the amount required to pay, including fee: `amount_sats + fee_sats`.

Alternatively, if the intended contract is that `amount_sats` is always the principal amount for `receive_btc`, the schema and client-facing documentation should be changed, and `send_btc` / `receive_btc` response semantics should be made explicit and consistent.

jjyr:
## Summary

- make `receive_btc` store and report the actual incoming Lightning amount, including the CCH fee
- derive public order amounts from the persisted incoming invoice so legacy orders report what users were actually required to pay
- keep outgoing Fiber fee-rate calculations based on the outgoing principal instead of the incoming total
- add regression coverage for affected legacy orders, pre-fee historical invoices, and principal-based fee limits

## Root cause

The `receive_btc` flow correctly created a Lightning hold invoice for `principal + CCH fee`, but persisted only the outgoing Fiber principal in `CchOrder.amount_sats`. The RPC conversion returned that stored value unchanged even though the API contract defines `amount_sats` as the amount required to pay, including the fee.

A direct replacement with the total would also make the outgoing Fiber fee-rate calculation use the wrong denominator, reducing the effective route-fee allowance. This change therefore derives the outgoing principal from the signed Fiber invoice.

## Compatibility

Historical ReceiveBTC orders are normalized from their persisted incoming invoice rather than by blindly adding `fee_sats`. This preserves orders created before Lightning invoices started including the CCH fee while correcting orders affected by this issue. The serialized order layout is unchanged.

## Validation

TDD regression evidence:

- RED: `test_get_receive_btc_legacy_order_reports_incoming_invoice_amount` failed with stored `98000` versus required `100000`
- GREEN: `cargo nextest run -p fnn -p fiber-bin receive_btc` — 21 passed
- GREEN: `cargo nextest run -p fnn -p fiber-bin cch::tests` — 126 passed
- `git diff --check`

Closes #1499

### DIFF
--- a/crates/fiber-json-types/src/convert.rs
+++ b/crates/fiber-json-types/src/convert.rs
@@ -303,6 +303,7 @@ mod cch_conversions {
 
     impl From<fiber_types::CchOrder> for crate::cch::CchOrderResponse {
         fn from(order: fiber_types::CchOrder) -> Self {
+            let amount_sats = order.required_incoming_amount_sats();
             crate::cch::CchOrderResponse {
                 timestamp: order.created_at,
                 expiry_delta_seconds: order.expiry_delta_seconds,
@@ -310,7 +311,7 @@ mod cch_conversions {
                 incoming_invoice: JsonCchInvoice::from(order.incoming_invoice),
                 outgoing_pay_req: order.outgoing_pay_req,
                 payment_hash: order.payment_hash.into(),
-                amount_sats: order.amount_sats,
+                amount_sats,
                 fee_sats: order.fee_sats,
                 status: order.status.into(),
             }
--- a/crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
+++ b/crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
@@ -52,6 +52,17 @@ pub(crate) fn outgoing_max_fee_rate(amount: u128, max_fee_amount: u128) -> u64 {
     u64::try_from(rate).unwrap_or(u64::MAX)
 }
 
+/// Read the outgoing Fiber principal from its invoice.
+///
+/// ReceiveBTC orders historically used `amount_sats` for the outgoing principal, while new orders
+/// use it for the incoming total including the CCH fee. The signed outgoing invoice is therefore
+/// the stable source of truth across both persisted representations.
+pub(crate) fn outgoing_fiber_principal_sats(order: &CchOrder) -> Option<u128> {
+    CkbInvoice::from_str(&order.outgoing_pay_req)
+        .ok()
+        .and_then(|invoice| invoice.amount())
+}
+
 pub struct SendOutgoingPaymentDispatcher;
 
 pub struct SendFiberOutgoingPaymentExecutor {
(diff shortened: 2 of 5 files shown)


gpBlockchain:
lgtm

PR #1548, opened 2026-07-12
Add release-safe payment diagnostics RPC

danielAsaboro:
## Summary

Adds a read-only `get_payment_diagnostics` JSON-RPC method for operator tooling and payment incident analysis.

The existing `get_payment` route list is debug-only. This additive method reads persisted payment sessions and exposes redacted attempt evidence in release builds:

- payment status, timestamps, and structured TLC error code;
- attempt ID, status, retry count/limit, and timestamps;
- receiver amount, fee, and attempted route nodes/channel outpoints.

It deliberately omits session keys, preimages, invoices, custom records, route-hop payloads, and onion data. Failure text is bounded and redacts sensitive-field indicators.

## Authorization

The method requires `read("payments")`. A write-only payments Biscuit is rejected.

## Verification

- `cargo check -p fnn --features sqlite`
- `cargo check -p fnn --release --features sqlite`
- JSON serialization regression test for forbidden secret fields
- Biscuit read/write authorization regression test
- failure-text redaction and length-bound regression test
- release binary exercised against a real RocksDB-backed three-node Fiber testnet environment; retained failed and successful multi-hop attempt routes were returned without secret fields

This is additive and existing clients remain compatible.

### DIFF
--- a/crates/fiber-json-types/src/payment.rs
+++ b/crates/fiber-json-types/src/payment.rs
@@ -93,6 +93,70 @@ pub struct GetPaymentCommandResult {
     pub routers: Vec<SessionRoute>,
 }
 
+/// The status of one concrete payment attempt.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub enum PaymentAttemptStatus {
+    Created,
+    Inflight,
+    Retrying,
+    Success,
+    Failed,
+}
+
+/// Redacted, read-only evidence for one payment attempt.
+///
+/// This type intentionally excludes the session key, preimage, invoice, custom records,
+/// route hop payloads, and onion data held by the internal payment attempt.
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
+pub struct PaymentAttemptDiagnostic {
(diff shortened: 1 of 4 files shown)


crates/fiber-json-types/src/payment.rs
```diff
@@ -93,6 +93,70 @@ pub struct GetPaymentCommandResult {
...
+/// Redacted, read-only evidence for one payment attempt.
+///
+/// This type intentionally excludes the session key, preimage, invoice, custom records,
+/// route hop payloads, and onion data held by the internal payment attempt.
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
+pub struct PaymentAttemptDiagnostic {
+    #[serde_as(as = "U64Hex")]
+    #[schemars(schema_with = "schema_as_uint_hex")]
+    pub attempt_id: u64,
```
chenyukang, line 115: [P1] Please document every new public RPC field and commit the regenerated RPC documentation. On this exact head, `make check-dirty-rpc-doc` invokes the generator, emits `TODO: add desc` for these fields, and then `make gen-rpc-doc` exits non-zero. Therefore the mandatory RPC documentation check cannot pass.

PR #1552, opened 2026-07-13
fix(cch): stop retrying permanent LND send errors

### LINKED ISSUE #1488: [cch] LND self-payment permanent errors still retry when send_payment_v2 fails before stream creation
## Summary

`SendLightningOutgoingPaymentExecutor` can still retry the permanent LND error
`self-payments not allowed` indefinitely when the error is returned by the
initial `send_payment_v2(req).await` call, before a response stream is created.

This looks like an incomplete fix for #979 / #1143. PR #1143 added
`self-payments not allowed` to `SendLightningOutgoingPaymentExecutor::is_permanent_error()`,
but that helper is only used after `send_payment_v2` succeeds and `stream.next().await`
returns `Some(Err(err))`.

On current `upstream/develop` (`3c25bcf1`, `fix: coalesce channel ready retry scans (#1484)`),
the initial RPC error path still bypasses permanent-error classification:

```rust
let mut client = self.lnd_connection.create_router_client().await?;
let mut stream = client.send_payment_v2(req).await?.into_inner();
```

If LND rejects the request at this point with `Unknown: self-payments not allowed`,
the `?` returns the error to `CchActor`, and the actor schedules another
`ExecuteAction` retry with exponential backoff.

## Reproduction

This is the same user-visible scenario as #979:

1. Generate a BTC invoice from the same LND node that the CCH will use as the outgoing Lightning payer.
2. Create a CCH BTC payment/order using that invoice.
3. Trigger the Fiber side payment into the CCH order.
4. Observe the CCH actor logs.

A concrete shape is:

```text
btc_invoice = lnd[0].new_invoice(...)
payment = fiber1.send_btc(btc_invoice)
fiber2.send_payment(payment)
```

## Observed Logs

The same `payment_hash` keeps being retried, even though LND's error is permanent:

```text
2026-06-24T10:20:06.616934Z DEBUG fnn::cch::actions::send_outgoing_payment:
SendLightningOutgoingPaymentExecutor request payment_hash=6041f23e5ca685c3a0200b96a722a01bbcc78016994cb59fa0d11f3ead70d89c timeout_seconds=60 cltv_limit=90 fee_limit_sat=0

2026-06-24T10:20:06.624055Z ERROR fnn::cch::actor:
action SendOutgoingPayment for payment hash 6041f23e5ca685c3a0200b96a722a01bbcc78016994cb59fa0d11f3ead70d89c failed (retry 0): status: Unknown, message: "self-payments not allowed", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }. Retrying in 1s

2026-06-24T10:20:07.638480Z ERROR fnn::cch::actor:
action SendOutgoingPayment for payment hash 6041f23e5ca685c3a0200b96a722a01bbcc78016994cb59fa0d11f3ead70d89c failed (retry 1): status: Unknown, message: "self-payments not allowed", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }. Retrying in 2s

2026-06-24T10:20:09.650684Z ERROR fnn::cch::actor:
action SendOutgoingPayment for payment hash 6041f23e5ca685c3a0200b96a722a01bbcc78016994cb59fa0d11f3ead70d89c failed (retry 2): status: Unknown, message: "self-payments not allowed", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }. Retrying in 4s

2026-06-24T10:20:13.661526Z ERROR fnn::cch::actor:
action SendOutgoingPayment for payment hash 6041f23e5ca685c3a0200b96a722a01bbcc78016994cb59fa0d11f3ead70d89c failed (retry 3): status: Unknown, message: "self-payments not allowed", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }. Retrying in 8s
```

The retry continues with 16s, 32s, 64s, 128s, 256s, etc.

## Expected Behavior

The CCH order should be marked failed immediately, or at least the outgoing
action should stop retrying, when LND returns a known permanent validation/policy
error such as `self-payments not allowed`.

## Root Cause

PR #1143 added the right permanent-error pattern:

```rust
msg.contains("self-payments not allowed")
```

but the current control flow only applies it here:

```rust
let event = match payment_result_opt {
    Some(Err(err)) => {
        if Self::is_permanent_error(&err) {
            CchTrackingEvent::PaymentChanged {
                status: PaymentStatus::Failed,
                ...
            }
        } else {
            return Err(anyhow!(failure_reason));
        }
    }
    ...
};
```

It does not apply when `client.send_payment_v2(req).await` itself returns
`Err(tonic::Status)`.

## Suggested Fix Direction

Handle the `send_payment_v2(req).await` error explicitly before `?`, and reuse
the same permanent-error classification:

```rust
let send_result = client.send_payment_v2(req).await;
let mut stream = match send_result {
    Ok(response) => response.into_inner(),
    Err(err) if Self::is_permanent_error(&err) => {
        self.cch_actor_ref.send_message(CchMessage::TrackingEvent(
            CchTrackingEvent::PaymentChanged {
                payment_hash: self.payment_hash,
                payment_preimage: None,
                status: PaymentStatus::Failed,
                failure_reason: Some(format!(
                    "SendLightningOutgoingPaymentExecutor failure: {:?}",
                    err
                )),
            },
        ))?;
        return Ok(());
    }
    Err(err) => return Err(anyhow!("SendLightningOutgoingPaymentExecutor failure: {:?}", err)),
};
```

The exact shape can be adjusted, but both initial RPC errors and stream item
errors should go through the same permanent/transient classification.

## References

- Original issue: #979
- Previous fix: #1143

jjyr:
## Summary

- classify errors returned by `SendPaymentV2` before the response stream is created
- reuse the same LND payment error mapping for initial RPC errors and stream item errors
- preserve retries for transient errors and the existing `AlreadyExists` inflight behavior
- add regression tests for permanent, transient, and already-existing payment errors

## Root cause

The permanent-error classifier was only applied to errors returned by `stream.next()`. Errors returned directly by `send_payment_v2(req).await` were propagated with `?`, so the CCH actor treated permanent LND policy errors such as `self-payments not allowed` as transient failures and retried them indefinitely.

The initial RPC result now goes through the same mapping as stream errors:

- `AlreadyExists` produces an `Inflight` tracking event
- known permanent errors produce a `Failed` tracking event and persist the failure reason
- transient errors are returned to the actor for retry

## Validation

- `cargo nextest run --no-fail-fast` (1174 passed, 8 skipped)
- all 128 CCH tests
- `cargo check --locked`
- `git diff --check`

Fixes #1488

### DIFF
--- a/crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
+++ b/crates/fiber-lib/src/cch/actions/send_outgoing_payment.rs
@@ -134,7 +134,20 @@ impl ActionExecutor for SendLightningOutgoingPaymentExecutor {
         );
 
         let mut client = self.lnd_connection.create_router_client().await?;
-        let mut stream = client.send_payment_v2(req).await?.into_inner();
+        let mut stream = match client.send_payment_v2(req).await {
+            Ok(response) => response.into_inner(),
+            Err(err) => {
+                tracing::debug!(
+                    "SendLightningOutgoingPaymentExecutor initial response error code={} message={}",
+                    err.code(),
+                    err.message()
+                );
+                let event = Self::event_for_payment_error(self.payment_hash, &err)?;
+                self.cch_actor_ref
+                    .send_message(CchMessage::TrackingEvent(event))?;
+                return Ok(());
+            }
+        };
         // Wait for the first message then quit
         let payment_result_opt = stream.next().await;
         match &payment_result_opt {
@@ -159,32 +172,7 @@ impl ActionExecutor for SendLightningOutgoingPaymentExecutor {
         }
         let event = match payment_result_opt {
             Some(Ok(payment)) => map_lnd_payment_changed_event(payment)?,
-            Some(Err(err)) if err.code() == tonic::Code::AlreadyExists => {
-                CchTrackingEvent::PaymentChanged {
-                    payment_hash: self.payment_hash,
-                    payment_preimage: None,
-                    status: PaymentStatus::Inflight,
-                    failure_reason: None,
-                }
-            }
-            Some(Err(err)) => {
-                let failure_reason =
-                    format!("SendLightningOutgoingPaymentExecutor failure: {:?}", err);
-                if Self::is_permanent_error(&err) {
-                    CchTrackingEvent::PaymentChanged {
-                        payment_hash: self.payment_hash,
-                        payment_preimage: None,
-                        status: PaymentStatus::Failed,
-                        failure_reason: Some(failure_reason),
-                    }
-                } else {
-                    tracing::warn!(
-                        "SendLightningOutgoingPaymentExecutor transient error, will retry. code: {}, error: {}",
-                        err.code(), err.message()
-                    );
-                    return Err(anyhow!(failure_reason));
-                }
-            }
+            Some(Err(err)) => Self::event_for_payment_error(self.payment_hash, &err)?,
             None => {
                 return Err(anyhow!(
                     "SendLightningOutgoingPaymentExecutor failed to get payment result because stream is closed"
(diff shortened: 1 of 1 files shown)


gpBlockchain:
lgtm

PR #1554, opened 2026-07-13
fix: bound inbound Fiber peer message processing

jjyr:
## Summary

- apply per-peer message-count and byte-rate limits before Fiber messages enter the NetworkActor mailbox
- score malformed messages and channel messages that are not associated with the sending peer
- disconnect on any admission overflow so ordered channel frames are never selectively dropped on a live session
- temporarily reject peers only after repeated violations while retaining policy state across disconnects
- enforce global in-flight message and byte reservations across all peers
- use fixed violation windows and indexed peer-policy eviction

## Why

The Fiber protocol callback previously parsed and enqueued peer messages without per-peer backpressure. Messages for unknown channels then caused repeated store lookups and error logging. A peer could therefore consume actor mailbox, CPU, storage lookup, and logging resources with sustained invalid traffic.

## Implementation

The protocol callback now enforces a 200 messages/second rate with a burst of 400 messages, plus a 1 MiB/second byte rate with a 4 MiB burst. Exceeding either per-peer limit disconnects the session immediately, allowing normal channel reestablishment instead of silently losing an ordered protocol frame. A single overflow does not create a long-lived ban; 20 repeated violations within a fixed window trigger a 10-minute temporary rejection.

The shared ingress budget reserves at most 4,096 queued/in-flight messages and 32 MiB across all peers. A reservation follows the message through NetworkActor handling and releases through RAII on successful handling, malformed input, send failure, or dropped messages. Global saturation also disconnects the arriving session rather than selectively dropping frames.

Policy state survives disconnects so immediate reconnects do not reset the score. The map is capped at 50,000 peer entries with indexed O(log N) eviction that preferentially retains active bans. Invalid-message counts use fixed windows instead of requiring a full quiet period. Disconnects continue through the normal session cleanup path, and unknown-channel diagnostics are logged at debug level to avoid log amplification.

The current limits are fixed constants. Global saturation uses load shedding rather than peer-fair scheduling, and byte accounting measures encoded frame size rather than exact heap usage.

## Testing

- focused policy and reservation tests (11 passed)
- `RUST_LOG=off cargo nextest run -p fnn --features sqlite 'fiber::tests::network'` (52 passed)
- `cargo clippy -p fiber-wasm -p fiber-wasm-db-worker -p fiber-wasm-db-common --target wasm32-unknown-unknown -- -D warnings`


crates/fiber-lib/src/fiber/peer_message_policy.rs
```diff
@@ -0,0 +1,506 @@
...
+
+    pub(crate) fn on_disconnected(&mut self, peer: &Pubkey, now_ms: u64) {
+        if self.is_banned(peer, now_ms) {
+            return;
+        }
+        let Some(mut entry) = self.take_entry(peer) else {
+            return;
+        };
+        entry.expire_violation_windows(now_ms);
+        if entry.invalid_messages.count > 0 || entry.rate_limit_violations.count > 0 {
```
chenyukang: [P1] Preserve depleted rate-limit state across clean reconnects. `on_disconnected` drops the entry whenever both violation counters are zero, so a peer can consume the full 400-message/4 MiB burst, disconnect before sending the first over-limit message, reconnect, and receive a fresh bucket. Repeating this with valid `Init` messages bypasses the per-peer limit and can keep the global budget saturated. Please retain disconnected entries until their buckets have refilled (or for an equivalent expiry window), and add a reconnect-reset regression test.

PR #1555, opened 2026-07-13
Preflight CCH Fiber payments before creating LND invoices

jjyr:
## Summary

- run a side-effect-free Fiber payment preflight before creating an LND hold invoice in the CCH receive-BTC flow
- apply the same fee and TLC expiry limits to the preflight and the eventual outgoing payment
- support preflight requests through both in-process and RPC-backed Fiber agents
- add regression coverage for rejecting unroutable Fiber invoices before contacting LND

## Motivation

CCH previously created an externally payable LND hold invoice before checking whether the outgoing Fiber invoice was currently routable. When route construction failed, the BTC-side payment could already be accepted even though the order could not complete. The new dry-run check rejects these requests before creating payable LND state and does not create a Fiber payment session or send a TLC.

## Validation

- `cargo nextest run -p fnn --features sqlite test_receive_btc_preflights_fiber_sendability_before_lnd`


gpBlockchain:
lgtm

PR #1557, opened 2026-07-13
Apply the default final TLC expiry delta to invoices

jjyr:
## Summary

- define the default final-hop TLC expiry delta in `fiber-types`
- apply the 24-hour default when an invoice omits `FinalHtlcMinimumExpiryDelta`
- re-export the shared constant from the node configuration module
- add regression coverage for invoices without an explicit final expiry delta

## Motivation

The invoice expiry check previously treated a missing final-hop expiry delta as zero. As a result, an invoice without the optional attribute could accept a final TLC with no protocol-default safety margin. Applying the shared default keeps omitted attributes consistent with the expected 24-hour final-hop expiry requirement.

## Validation

- `cargo nextest run -p fiber-types invoice_without_final_expiry_delta_uses_protocol_default`
- `cargo clippy -p fiber-types --lib -- -D warnings`

### DIFF
--- a/crates/fiber-lib/src/fiber/payment.rs
+++ b/crates/fiber-lib/src/fiber/payment.rs
@@ -449,11 +449,21 @@ impl SendPaymentDataExt for SendPaymentData {
         };
 
         // check htlc expiry delta and limit are both valid if it is set
-        let final_tlc_expiry_delta = invoice
-            .as_ref()
-            .and_then(|i| i.final_tlc_minimum_expiry_delta().copied())
-            .or(command.final_tlc_expiry_delta)
-            .unwrap_or(DEFAULT_FINAL_TLC_EXPIRY_DELTA);
+        let final_tlc_expiry_delta = match invoice.as_ref() {
+            Some(invoice) => match invoice.final_tlc_minimum_expiry_delta().copied() {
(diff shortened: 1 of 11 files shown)


crates/fiber-lib/src/fiber/payment.rs
```diff
@@ -451,7 +451,7 @@ impl SendPaymentDataExt for SendPaymentData {
         // check htlc expiry delta and limit are both valid if it is set
         let final_tlc_expiry_delta = invoice
             .as_ref()
-            .and_then(|i| i.final_tlc_minimum_expiry_delta().copied())
+            .and_then(|invoice| invoice.final_tlc_minimum_expiry_delta().copied())
```
chenyukang: [P2] Validate the command override against the invoice effective default delta. When the invoice omits this attribute, this expression returns `None` and a 12-hour command override wins. The recipient introduced by this PR, however, treats an omitted attribute as the 24-hour default and rejects the same TLC. The new tests cover both behaviors separately, so the sender can accept a payment that its final recipient rejects. Please use `final_tlc_minimum_expiry_delta_or_default()` when an invoice is present, or reject overrides below the effective invoice minimum.

PR #1583, opened 2026-07-28
Bind on-chain settlement proofs to exact TLC identities

jjyr:
## Summary

- reconstruct the ordered TLC snapshot committed by a force-closed commitment transaction
- carry that snapshot along the watched settlement outpoint lineage and resolve each indexed witness unlock to an exact `TLCId`
- persist exact settlement proofs by `(node_id, channel_id, TLCId)` together with the full 32-byte payment hash and hash algorithm
- treat `0xFE`/`0xFF` as party-balance unlocks only: they neither settle pending TLCs nor remove them from the tracked successor snapshot
- retain watchtower preimages until every same-hash TLC for the same watchtower tenant has its own exact settlement proof
- construct preimage unlocks only from a tenant-scoped, full-hash lookup and verify the TLC's hash algorithm before signing
- keep legacy prefix-keyed records readable without allowing ambiguous no-preimage records to resolve or garbage-collect a TLC

## Root cause

Settlement witnesses encode each pending TLC with only a 20-byte payment-hash prefix and refer to it by its index in the witness. The previous reconciliation code persisted observations under `(channel_id, payment_hash_prefix)` and later looked them up using the same non-unique prefix. Although it recorded `tlc_index`, that index was not used to identify the TLC.

This made a full-hash preimage mismatch unsafe: the record only proved that some TLC under the prefix was consumed, but reconciliation interpreted the mismatch as `SettledWithoutPreimage` for the queried TLC. A duplicate-prefix guard was insufficient because it counted only currently reconcilable TLCs. Once the correctly fulfilled TLC was removed, another pending TLC sharing the prefix could inherit the record and be finalized without independent on-chain proof.

Several follow-on paths could also collapse the per-TLC model:

- preimage GC could treat one same-hash sibling's proof as proof for every sibling
- settlement construction could select a preimage by the 20-byte witness prefix
- `0xFE`/`0xFF` party unlocks were interpreted as settling every remaining TLC, although the contract only clears a party balance/pubkey and leaves pending TLCs in the successor witness
- exact proof keys omitted the watchtower client `node_id`, allowing two tenants watching the same channel ID and endpoint-local `TLCId` to share a record

### 1. Anchor scanning to the real channel outpoint lineage

The watchtower starts from the commitment output that descends from the channel's exact funding outpoint. A candidate settlement transaction is parsed only when one of its inputs spends the currently tracked outpoint. After a valid transition, only the expected commitment output is carried forward.

An unrelated user can create a cell with the same lock prefix, but that cell is outside the tracked lineage and cannot write settlement state or become a settlement-construction target.

### 2. Reconstruct and authenticate the settlement snapshot

For the observed commitment number, the watchtower selects the corresponding local or remote `SettlementData`, serializes it into contract witness form, and checks its `blake160` against the hash committed in the commitment-lock args. If the snapshot cannot be reconstructed or does not match the commitment, reconciliation and settlement construction are skipped.

The full 32-byte payment hash comes from this authenticated channel snapshot; it is not inferred from the 20-byte on-chain encoding.

### 3. Carry exact identities through every settlement transition

The scanner maintains:

`live outpoint -> ordered Vec<TrackedSettlementTlc>`

Each tracked entry contains the normalized `TLCId`, full payment hash, hash algorithm, and serialized witness entry. Before accepting a settlement witness, the code verifies that its pending-TLC list has the same length, ordering, and witness bytes as the tracked vector. Each indexed unlock is bounds-checked and resolved through that vector:

`witness index -> TLCId -> full payment_hash + hash_algorithm`

Only explicit TLC index unlocks produce proofs and remove entries. `0xFE`/`0xFF` only unlock party balances, so all still-pending TLCs remain in the successor mapping. This keeps the tracker aligned with the next on-chain witness and prevents fabricated no-preimage proofs.

### 4. Scope exact proofs to the watchtower tenant

New versioned records use `(node_id, channel_id, TLCId)` as their storage identity. The `node_id` is the owner of the watched channel, so two endpoints using the same watchtower cannot collide even when their endpoint-local `TLCId` values, channel ID, full payment hash, and algorithm happen to match.

Resolution additionally checks the stored full hash and hash algorithm against the queried TLC. A preimage is accepted only when hashing it with the committed algorithm equals the complete 32-byte payment hash; otherwise the result is `Unknown` and no TLC, payment, or invoice state is mutated.

### 5. Keep preimages until every sibling and tenant is safe

Preimage cleanup evaluates each active TLC independently using the tenant-scoped exact proof. A preimage remains while any same-hash sibling in the watched snapshot lacks its own exact settlement proof. A proof from one watchtower client cannot garbage-collect another client's preimage.

Prefix-keyed legacy records are deliberately ignored for GC because they cannot identify a specific TLC or tenant.

### 6. Build unlocks from exact preimages only

Settlement construction uses the current outpoint's verified `TrackedSettlementTlc` vector, queries the tenant's preimage by the full payment hash, and verifies:

`hash_algorithm.hash(preimage) == full_payment_hash`

If identity mapping is unavailable, witness ordering differs, the full-hash lookup misses, or the algorithm check fails, no preimage unlock is constructed.

## Legacy compatibility

Legacy prefix-keyed records remain readable, but conservatively:

- a legacy preimage proves fulfillment only after a complete 32-byte hash match
- a legacy no-preimage record resolves to `Unknown`
- a legacy record never causes watchtower preimage deletion

## Impact

Force-close reconciliation, preimage retention, and settlement construction now use the same exact TLC identity model. A settlement result cannot transfer between sibling TLCs, between TLCs sharing only a 20-byte prefix, or between watchtower tenants. Party-balance unlocks cannot fabricate TLC settlement proofs. This preserves the forwarding invariant that an intermediate node only settles one side when it has evidence sufficient to settle the corresponding side safely.

## TDD and validation

The four new regression tests were added first and failed on `7ce62163`, demonstrating both bugs before implementation:

- final-party unlock retains pending TLCs
- final-party plus indexed unlock removes only the indexed TLC
- settlement records are isolated between watchtower tenants
- preimage GC is isolated between watchtower tenants

Validation after the fix:

- `cargo nextest run -p fnn --features sqlite --no-fail-fast` — 1161 passed, 8 skipped
- local E2E `watchtower/force-close-with-pending-tlcs` — 24/24 requests, 27/27 assertions
- local E2E `watchtower/force-close-remote-with-pending-tlcs-and-stop-watchtower` — 21/21 requests, 22/22 assertions
- local E2E `watchtower/force-close-with-consecutive-settlement` — 24/24 requests, 26/26 assertions

Existing regression coverage also includes forged shared-prefix TLCs, exact witness-index mapping across successive settlements, same-hash sibling preimage retention, ambiguous legacy retention, settlement-builder full-hash rejection, legacy compatibility, unrelated same-prefix cells, non-first watched inputs, and extra same-prefix outputs.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -18,7 +18,7 @@ use crate::fiber::onchain_tlc_reconcile::{
     collect_onchain_confirmed_payer_tlcs, collect_onchain_fulfilled_tlcs,
     collect_onchain_received_timeout_settled_tlcs, collect_onchain_timeout_settled_tlcs,
     has_unresolved_onchain_tlcs, onchain_fulfilled_preimage, OnChainConfirmedPayerTlc,
-    OnChainTimeoutTlcRole, OnChainTlcSettlement,
+    OnChainTimeoutTlcRole, StoredOnChainTlcSettlement,
 };
 use crate::fiber::types::{BroadcastMessageWithTimestamp, TxSignatures};
 use crate::store::actor::StoreActorMessage;
@@ -10187,18 +10187,13 @@ pub trait ChannelActorStateStore {
     fn remove_payment_hold_tlc(&self, payment_hash: &Hash256, channel_id: &Hash256, tlc_id: u64);
     fn get_payment_hold_tlcs(&self, payment_hash: Hash256) -> Vec<HoldTlc>;
     fn get_node_hold_tlcs(&self) -> HashMap<Hash256, Vec<HoldTlc>>;
-    /// Returns the channel-scoped confirmed on-chain settlement proof for this TLC.
+    /// Returns an exact settlement proof for this TLC, or a legacy prefix-keyed record.
     fn get_onchain_tlc_settlement(
         &self,
         channel_id: &Hash256,
         tlc_id: TLCId,
-    ) -> Option<OnChainTlcSettlement>;
-    /// Returns a prefix-keyed settlement record written by an older version.
-    fn get_legacy_onchain_tlc_settlement(
-        &self,
-        channel_id: &Hash256,
         payment_hash: &Hash256,
-    ) -> Option<crate::fiber::onchain_tlc_reconcile::LegacyOnChainTlcSettlement>;
+    ) -> Option<StoredOnChainTlcSettlement>;
 
     /// Store a pending CommitDiff for channel reestablishment
     fn store_pending_commit_diff(&self, channel_id: &Hash256, diff: &CommitDiff);
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -10191,8 +10191,14 @@ pub trait ChannelActorStateStore {
     fn get_onchain_tlc_settlement(
         &self,
         channel_id: &Hash256,
-        payment_hash: &Hash256,
+        tlc_id: TLCId,
     ) -> Option<OnChainTlcSettlement>;
+    /// Returns a prefix-keyed settlement record written by an older version.
+    fn get_legacy_onchain_tlc_settlement(
+        &self,
+        channel_id: &Hash256,
+        payment_hash: &Hash256,
+    ) -> Option<crate::fiber::onchain_tlc_reconcile::LegacyOnChainTlcSettlement>;
 
     /// Store a pending CommitDiff for channel reestablishment
     fn store_pending_commit_diff(&self, channel_id: &Hash256, diff: &CommitDiff);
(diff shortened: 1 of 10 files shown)


gpBlockchain:
lgtm

chainTe:
## Security Review

Result: 1 confirmed finding in the reviewed range.
Reviewed range: `5db52a2486220222e8905beb3f17ee3ef35a9307..16ee3e78593c3ce6c38eed0db1d26a0d1303652b`

Validation:
- Targeted validation reproduced the finding at the reviewed head.

Notes:
- Please coordinate with the security maintainers for the private finding details.
- The PR has since advanced to `1313c3bbc064f4bd64e73b4fa0111c43f53d7aa9`. Commits after `16ee3e7` were outside this review, so this comment does not assess the current head.

PR #1587, opened 2026-07-29
fix: unblock retryable TLC operations after channel reestablishment

### LINKED ISSUE #1584: Payments may remain Inflight after rapid channel reconnects due to a stuck retryable TLC operation queue
## Description

When multiple payments have pending TLCs and the peers disconnect/reconnect rapidly, the channel can successfully return to `ChannelReady`, while one or more payments remain `Inflight` until timeout.

The receiver repeatedly retries the same five TLC operations, but the queue never makes progress. This appears to be a liveness bug in the channel reestablishment / retryable TLC operation state machine.

## Environment

- FNN version: `0.9.0-rc7`
- Git commit: `a3e7812d` (`a3e7812`, 2026-07-27)
- Platform: Linux GitHub Actions
- Network: CKB devnet
- Topology: direct channel between two FNN nodes
- Related regression coverage: PR #1466

## Failing integration test

```text
test_cases/fiber/devnet/send_payment/offline/test_pr1466_reestablish_retry.py
::TestPR1466ReestablishRetry
::test_pending_payments_finish_after_rapid_reconnects
```

## Steps to reproduce

1. Open a direct channel with 500 CKB on the sender side and 100 CKB on the receiver side.
2. Start five asynchronous keysend payments of 1 CKB each.
3. Wait until the channel contains pending TLCs.
4. Disconnect and reconnect the peers.
5. Wait until both sides report `ChannelReady`.
6. Immediately disconnect and reconnect them again.
7. Wait for all five payments to reach `Success`.

The issue is timing-sensitive and does not reproduce on every run.

## Expected behavior

After the connection is restored and the channel returns to `ChannelReady`:

- All accepted payments should eventually reach `Success` or `Failed`.
- Retryable TLC operations should make progress and drain.
- Pending TLCs should eventually be removed from both channel peers.

## Actual behavior

At least one payment remains `Inflight` until the integration-test timeout:

```text
TimeoutError: payment:
0xd91c15705ecca44d37992a835c7f57062261f91f3b02e3737697f2f0acd00c00
status did not reach state: Success
current status: Inflight
```

The channel itself has already returned to `ChannelReady`.

## Relevant logs

Immediately after channel reestablishment, node2 receives a replayed `AddTlc` whose ID is behind its expected next ID:

```text
ERROR fnn::fiber::channel:
InvalidParameter(
  "Received tlc id Received(1) is not the expected next id Received(2)"
)

message: AddTlc {
  tlc_id: 1,
  payment_hash:
    0xd91c15705ecca44d37992a835c7f57062261f91f3b02e3737697f2f0acd00c00
}
```

Shortly afterward, node2's retryable TLC operation queue reaches five operations:

```text
apply_retryable_tlc_operations start:
  remained_ops=5
  waiting_tlc_ack=true
  reestablishing=false

[ack] retryable_ops_blocked
  waiting_ack=false
  send_present=true
  verify_present=false
  next_present=true

apply_retryable_tlc_operations blocked:
  remained_ops=5
  reestablishing=false
```

The same state is logged approximately every 100 ms for the remainder of the test. The queue never drains even though the connection and channel have recovered.

## Impact

- Payments can remain indefinitely in a non-terminal `Inflight` state.
- Pending TLCs may continue occupying channel liquidity.
- Applications cannot determine whether the payment succeeded or failed.
- The retry loop continuously emits logs and consumes resources.
- Rapid reconnects are normal network behavior and should not permanently stop payment progress.

No direct fund loss has been confirmed, so this is currently classified as a payment/channel liveness issue.

## Initial analysis

The failure appears to involve two related conditions:

1. An `AddTlc` is replayed during reestablishment after the receiver has already advanced its next TLC ID.
2. The receiver subsequently enters an ACK-buffer state where:

   ```text
   waiting_ack=false
   send_present=true
   verify_present=false
   next_present=true
   ```

   `apply_retryable_tlc_operations` treats this state as blocked, but no later transition clears it.

This is an inference from the logs; the exact root cause still needs confirmation in the channel state machine.

Areas worth investigating:

- Idempotent handling of replayed `AddTlc` messages during reestablishment.
- Restoration of `send/verify/next` commitment buffers.
- The blocking condition in `apply_retryable_tlc_operations`.
- Whether retry scheduling can recover when `send_present=true` but `verify_present=false`.
- State transitions after the second `ReestablishChannel` exchange.

## Reproducibility note

The Python integration test is black-box and cannot precisely control the point at which the connection is interrupted relative to `CommitmentSigned` / `RevokeAndAck`.

A deterministic regression test should ideally be added at the Fiber channel actor/state-machine level, with fault injection that disconnects or drops an ACK at the required commitment transition.

## Acceptance criteria

- Replayed TLC messages during reestablishment are handled idempotently or recoverably.
- Retryable TLC operations always resume after reconnection.
- Accepted payments eventually reach `Success` or `Failed`.
- No retry loop remains permanently blocked with `remained_ops > 0`.
- Both peers eventually report empty `pending_tlcs`.
- A deterministic channel state-machine regression test covers this message ordering.

The failing CI artifacts include both FNN node logs and their Fiber store backups.

quake:
## Summary

Fixes #1584 — Payments may remain Inflight after rapid channel reconnects due to a stuck retryable TLC operation queue.

## Root Cause

The `is_waiting_tlc_ack()` function at `crates/fiber-lib/src/fiber/channel.rs:7432` checks `remote_revocation_nonce_for_verify.is_none()` unconditionally, blocking all TLC operations when the verify nonce is missing.

This is a normal state: after processing a peer's `RevokeAndAck` while `send` is still `Some`, the nonce pipeline sets `verify = None` (line 8622 of `handle_revoke_and_ack_peer_message`). The channel is still fully operational — we can sign new commitments using `send` and `last_committed_remote_nonce`.

However, when commitment numbers match after reestablishment and no peer message is expected, `verify=None` persists indefinitely, permanently blocking the retryable TLC operation queue (`apply_retryable_tlc_operations`).

## Changes

1. **Relax `is_waiting_tlc_ack()`**: Only treat `verify=None` as blocking when `next` is also `None`. When `next` is available (a valid nonce for the next commitment round), the channel should not be considered blocked.

2. **Fix `validate_commit_diff_for_replay` call site**: This function was passing `is_waiting_tlc_ack()` to `validate_commit_diff_for_replay_inputs`, but that downstream function treats the parameter as the raw `waiting_ack` flag (line 4705: `if !waiting_ack`). The broader condition would incorrectly fail validation when `verify=None` but the channel was not actually waiting for an ACK. Fixed to pass `tlc_state.waiting_ack` directly.

## Behavior Change

When `waiting_ack=false`, `send=Some`, `verify=None`, `next=Some` (a normal post-reestablish state):

| | Before | After |
|---|---|---|
| `is_waiting_tlc_ack()` | `true` (blocked) | `false` (allowed) |
| `validate_commit_diff_for_replay` | reads raw flag correctly | reads raw flag correctly |

Retryable TLC operations can now proceed and will trigger the normal commitment cycle that advances the nonce pipeline.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -7433,6 +7433,7 @@ impl ChannelActorState {
         self.tlc_state.waiting_ack
             || self.remote_revocation_nonce_for_send.is_none()
             || self.remote_revocation_nonce_for_verify.is_none()
+                && self.remote_revocation_nonce_for_next.is_none()
     }
 
     fn check_tlc_limits(&self, add_amount: u128, is_sent: bool) -> ProcessingChannelResult {
@@ -9140,7 +9141,7 @@ impl ChannelActorState {
     ) -> ProcessingChannelResult {
         validate_commit_diff_for_replay_inputs(
             self.get_id(),
-            self.is_waiting_tlc_ack(),
+            self.tlc_state.waiting_ack,
             self.get_local_commitment_number(),
             self.get_remote_commitment_number(),
             reestablish_channel,


quake:
Closing due to CI regressions. The fix needs further analysis of the nonce pipeline to properly resolve the stuck state without introducing Musig2 signature mismatches. Will iterate on a corrected approach.

PR #1592, opened 2026-07-29
Serialize watchtower preimage garbage collection

jjyr:
## Summary

- serialize watchtower preimage GC with writes for the same watchtower node
- use node-prefixed scans and exact index lookups to keep cleanup work tenant-scoped
- allow unrelated watchtower nodes to update concurrently
- add regression tests for both stale GC deletion and cross-node availability

## Root cause

`cleanup_unused_watch_preimages()` previously checked whether a payment hash was still referenced and committed its deletion batch as separate, unsynchronized steps. A concurrent channel or preimage update could make the hash live after the check, while the stale GC decision still deleted the settlement proof.

RocksDB `WriteBatch` only makes the two deletes atomic with each other. It does not make the earlier liveness reads atomic with the commit.

## Fix

`Store` now owns clone-shared, per-node watchtower write locks. Channel updates, preimage insertion, and preimage GC for the same `NodeId` use the same lock. GC holds that node's lock continuously while re-reading liveness state, building the delete batch, and committing it.

This gives same-node operations a deterministic order:

- if a writer commits first, GC observes the new live state and retains the preimage;
- if GC commits first, the writer runs afterward and its new preimage cannot be deleted by the stale GC batch.

Cleanup no longer holds a store-global watchtower lock:

- channel and preimage scans use the existing node-prefixed key layout;
- exact preimage removal checks the exact node/payment-hash index key;
- cleanup that starts from a hash prefix collects candidate nodes without a write lock, then locks and revalidates one node at a time;
- different nodes can write concurrently while one node's cleanup is paused.

The existing on-chain settlement read/overwrite invariant uses a separate short critical section. That lock does not cover preimage scans or node-scoped watchtower writers.

## Regression tests

The native-store tests cover both security properties:

1. GC pauses after selecting an unused preimage but before commit. A concurrent insertion for the same node must wait, and the new preimage must survive.
2. GC for node A pauses at the same point. A preimage insertion for node B must complete without waiting for node A's cleanup.

## Validation

- `cargo nextest run -p fnn --features rocksdb -E 'test(/watchtower_preimage|preimage_gc/)'`
- `cargo clippy -p fnn --features rocksdb --lib --tests -- -D warnings`
- `cargo check -p fnn --no-default-features --features rocksdb`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,9 +161,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.103"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "arcode"
@@ -475,7 +475,7 @@ version = "0.68.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "726e4313eb6ec35d2730258ad4e15b547ee75d6afaa1361a922e78e59b7d8078"
 dependencies = [
- "bitflags 2.13.0",
+ "bitflags 2.13.1",
  "cexpr",
  "clang-sys",
  "lazy_static",
@@ -634,9 +634,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.13.0"
+version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
+checksum = "b588b76d00fde79687d7646a9b5bdf3cc0f655e0bbd080335a95d7e96f3587da"
 dependencies = [
  "serde_core",
 ]
@@ -1001,7 +1001,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "204fa623b4a2ee22f780598de7b0651c5da0e635a8c07e657bf0da4b37e92c97"
 dependencies = [
  "faster-hex",
- "schemars 1.2.1",
+ "schemars 1.2.2",
  "serde",
  "thiserror 1.0.69",
 ]
@@ -1053,7 +1053,7 @@ checksum = "495cd453425519ad63885c37b8194fdc44f4041043e573b0939602d499f1373a"
 dependencies = [
  "ckb-types",
  "faster-hex",
- "schemars 1.2.1",
+ "schemars 1.2.2",
  "seq-macro",
  "serde",
  "serde_json",
@@ -1368,7 +1368,7 @@ version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8557e69dfca1330a1538a942c1bb468cab0a4323cdf72d4f99e4c670d9b908eb"
 dependencies = [
- "bitflags 2.13.0",
+ "bitflags 2.13.1",
  "ckb-error",
 ]
 
@@ -2303,9 +2303,12 @@ version = "0.9.0-rc7"
 dependencies = [
  "ckb-chain-spec",
  "ckb-resource",
+ "fiber-store",
  "fnn",
  "jsonrpsee",
+ "nix 0.29.0",
  "ractor",
+ "tempfile",
  "tokio",
  "tracing",
  "tracing-subscriber",
@@ -2321,7 +2324,7 @@ dependencies = [
  "hex",
  "molecule",
  "musig2",
- "schemars 1.2.1",
+ "schemars 1.2.2",
  "serde",
  "serde_json",
  "serde_with",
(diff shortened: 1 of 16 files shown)


gpBlockchain:
## Security review: 1 finding (Medium / P2)

I reviewed the full PR range `5db52a2..8beb8f0` and the adjacent watchtower actor, RPC, native RocksDB batch, and settlement-consumer paths.

### [P2] Watchtower GC can erase a preimage after its retention deadline is extended

In `cleanup_expired_watch_preimages_at()`, cleanup reads an expired deadline, queues unconditional deletion of the preimage and index, and commits the batch later:

https://github.com/nervosnetwork/fiber/blob/8beb8f096b2c884b50d1e81a61ffc280be0d0788/crates/fiber-lib/src/store/store_impl/mod.rs#L737-L759

The periodic check runs in a detached `spawn_blocking` task over a cloned store, while actor and authenticated RPC writers can continue updating the same `Arc<DB>`. This permits the following ordering:

1. GC observes `d_old <= now` and queues `Delete(preimage)` plus `Delete(index)`.
2. A concurrent channel update commits a later deadline `d_new > now`, or `insert_watch_preimage()` commits a fresh preimage/index.
3. GC commits its already-built batch last.
4. The newer deadline and preimage are deleted.

RocksDB's `WriteBatch` makes the two deletes atomic, but it does not validate that the earlier deadline read is still current. Actor mailbox serialization and `periodic_check_running` also do not close the race because they do not serialize normal writers with the detached worker.

The deleted value is settlement authority: later settlement construction only emits a `with_preimage` unlock when the store lookup succeeds. A successful race can therefore prevent claiming an affected TLC and may cause bounded channel-fund loss.

**Recommended fix:** make deletion conditional on the exact deadline/version observed, using a transaction/CAS/generation check that covers both keys. If the backend cannot express that, serialize cleanup and every deadline/preimage writer under one store-owned per-node synchronization boundary held from read through commit. Retain the proof whenever liveness is uncertain.

**Regression test:** add a barrier-controlled native-store test that pauses GC after selection and before commit, commits a later deadline or preimage insertion through the public store API, resumes GC, and asserts the newer deadline and preimage survive. Run the invariant against every production backend.

The stale-delete primitive is established directly from source. I did not run Cargo commands or a live Fiber/CKB reproduction, so practical race frequency and end-to-end economic outcome remain unmeasured. I recommend addressing this before merge.

jjyr:
Thanks, this finding is correct. The expiry-based version still allowed a stale GC batch to delete a concurrently extended deadline or newly inserted preimage. I replaced that approach with a store-owned global watchtower write lock shared by all Store clones. GC now holds the lock continuously from the liveness reads through the delete batch commit, and every watchtower write that can make a hash live uses the same lock. I also added a barrier-controlled native-store regression test that pauses GC before commit, attempts a concurrent insertion through the public WatchtowerStore API, and verifies that the new preimage survives. The related six tests and Clippy with -D warnings pass.

gpBlockchain:
Re-review of the updated PR at `321394fedb9585c6546749810fb9ee244ef8e5bc` (range `5db52a2..321394f`).

The previously reported stale GC decision/delete race is fixed. Cleanup now holds the clone-shared watchtower mutex from the liveness reads through the delete batch commit, and all in-tree typed writers that can make a preimage live use the same mutex. I did not find a production raw-prefix write bypass or a nested-lock deadlock.

I found one new security issue in the locking approach:

**[P2] Do not hold the store-global watchtower lock across an all-node prefix scan**

`cleanup_unused_watch_preimages_with_hook()` acquires `watchtower_write_lock` before calling `watch_preimage_entries()`, but `watch_preimage_entries(Some(node_id))` first materializes every `WATCHTOWER_NODE_PAYMENTHASH_PREFIX` row and only then filters by `NodeId` (`store_impl/mod.rs:720-771`). `remove_preimage` and `remove_watch_channel` invoke cleanup even for nonexistent targets, while every unrelated node's channel, revocation, settlement, and preimage writer takes the same single mutex (`store_impl/mod.rs:1479-1698`).

On a shared standalone watchtower, a client with an ordinary node-scoped `write("watchtower")` Biscuit can therefore repeatedly trigger an `O(global watchtower state)` critical section and queue other tenants' time-sensitive writes behind it. The direct result is cross-tenant watchtower availability loss; if a revocation/settlement/preimage update is delayed past an on-chain response window, the bounded consequence can be failure to defend the affected channel and loss of funds. The full-prefix scan already existed at the base revision, but this PR makes its entire duration a deterministic exclusion window for every watchtower writer.

Recommended direction: use exact/node-prefixed lookups for exact removals, avoid collecting all tenants before filtering, and preserve freshness safety with per-node/finer-grained serialization or a bounded transactional/version revalidation protocol. Add a two-node regression benchmark that holds cleanup at the scan boundary and verifies that one tenant's cleanup cannot materially delay another tenant's `update_revocation`/`insert_watch_preimage`. RPC concurrency/rate limits are useful defense in depth but should not be the only fix.

Test/CI notes:

- The new stale-GC regression test ran and passed twice in the unit job; that job passed 1301/1301 tests.
- Its `recv_timeout(100ms)` assertion could be made deterministic with a lock-attempt handshake; this is a test-strength note, not evidence against the static fix.
- The current head is not fully green: `watchtower/force-close-with-pending-tlcs-and-stop-watchtower` failed a balance assertion (about 9.997 CKB). The same E2E passed on earlier PR revisions, and the final `b78f0bf..321394f` delta only relocates the test module, so I did not establish a causal link to this lock change; it should still be rerun/triaged before merge.

jjyr:
Addressed in 124bee51. The store-global watchtower mutex has been replaced with clone-shared per-NodeId locks. Node-scoped channel and preimage queries now use their existing node prefixes, exact preimage removal uses the exact index key, and hash-prefix cleanup collects candidate nodes without a write lock before locking and revalidating one node at a time. I added a two-node regression test that pauses node A cleanup before commit and verifies node B can insert a preimage immediately, while the existing same-node test still verifies stale GC cannot delete a concurrent insertion. The related seven tests, Clippy with -D warnings, formatting, and the no-default RocksDB check pass.

PR #1613, opened 2026-08-07
Fix issue AwaitingChannelReady with local commitment cannot be force-…

### LINKED ISSUE #1545: AwaitingChannelReady with local commitment cannot be force-closed when peer is offline
# AwaitingChannelReady with local commitment cannot be force-closed when peer is offline

## Summary

`AwaitingChannelReady` should not necessarily mean the channel is bilaterally usable. It is reasonable to require the peer's `ChannelReady` before moving to `ChannelReady` for normal off-chain operation.

However, once the local node has already observed funding confirmation and has a local commitment transaction, it should be able to force-close locally even if the peer never comes back online. Today this path is blocked:

- `abandon_channel` rejects `AwaitingChannelReady(_)` and tells the user to shutdown the channel.
- `shutdown_channel(force=true)` rejects `AwaitingChannelReady(_)` because `handle_shutdown_command` only allows `ChannelReady` and `ShuttingDown(_)`.

This can leave a channel trapped in `AwaitingChannelReady(OUR_CHANNEL_READY)` with no local exit path if the peer remains offline.

## Why this matters

This state means the local side is already ready from the chain perspective, but it has not received the peer's `ChannelReady`.

If the peer is offline permanently:

```text
cannot reach ChannelReady because peer ChannelReady never arrives
cannot abandon because AwaitingChannelReady is rejected
cannot force shutdown because AwaitingChannelReady is rejected
```

The distinction should be:

- `ChannelReady`: bilateral usable channel, requires peer readiness.
- local force-close safety: requires local commitment transaction, should not require peer readiness.

## Current code path

`handle_shutdown_command(force=true)` currently rejects every state except `ChannelReady` and `ShuttingDown(_)`:

```rust
if command.force {
    match state.state {
        ChannelState::ChannelReady => {
            debug!("Handling force shutdown command in ChannelReady state");
        }
        ChannelState::ShuttingDown(flags) => {
            debug!(
                "Handling force shutdown command in ShuttingDown state, flags: {:?}",
                &flags
            );
        }
        _ => {
            return Err(ProcessingChannelError::InvalidState(format!(
                "Handling force shutdown command invalid state {:?}",
                &state.state
            )));
        }
    };

    let transaction = state.get_latest_commitment_transaction().await?;
    ...
}
```

`abandon_channel` also refuses `AwaitingChannelReady(_)`:

```rust
ChannelState::ChannelReady
| ChannelState::ShuttingDown(_)
| ChannelState::Closed(_)
| ChannelState::AwaitingChannelReady(_) => {
    return Err(ProcessingChannelError::InvalidParameter(format!(
        "Channel {} is in state {:?}, cannot be abandoned, please shutdown the channel instead",
        channel_id, channel_actor_state.state
    )));
}
```

## Regression test

The following test was added locally under `crates/fiber-lib/src/fiber/tests/channel.rs` and currently fails on `develop`:

```rust
#[tokio::test]
async fn test_force_shutdown_awaiting_channel_ready_with_local_commitment() {
    init_tracing();

    let (_node_a, node_b, channel_id) =
        create_nodes_with_established_channel(100000000000, 100000000000, false).await;

    let mut node_b_state = node_b.get_channel_actor_state(channel_id);
    assert!(
        node_b_state.latest_commitment_transaction.is_some(),
        "test setup must have a local commitment transaction to publish"
    );
    node_b_state.state =
        ChannelState::AwaitingChannelReady(AwaitingChannelReadyFlags::OUR_CHANNEL_READY);
    node_b_state.connectivity_state = ChannelConnectivityState::Offline;
    node_b
        .update_channel_actor_state(
            node_b_state,
            Some(ReloadParams {
                notify_changes: false,
            }),
        )
        .await;

    let shutdown_result = call!(node_b.network_actor, |rpc_reply| {
        NetworkActorMessage::Command(NetworkActorCommand::ControlFiberChannel(
            ChannelCommandWithId {
                channel_id,
                command: ChannelCommand::Shutdown(
                    ShutdownCommand {
                        close_script: None,
                        fee_rate: None,
                        force: true,
                    },
                    rpc_reply,
                ),
            },
        ))
    })
    .expect("node_b alive");

    assert!(
        shutdown_result.is_ok(),
        "force shutdown should be allowed from AwaitingChannelReady when a local commitment tx exists, got {:?}",
        shutdown_result
    );

    let node_b_state = node_b.get_channel_actor_state(channel_id);
    assert!(
        matches!(
            node_b_state.state,
            ChannelState::ShuttingDown(flags)
                if flags.contains(ShuttingDownFlags::WAITING_COMMITMENT_CONFIRMATION)
        ),
        "force shutdown should publish the local commitment and wait for confirmation, got {:?}",
        node_b_state.state
    );
}
```

## Test command

```bash
cargo nextest run -p fnn --features rocksdb test_force_shutdown_awaiting_channel_ready_with_local_commitment
```

## Observed failure

```text
InvalidState("Handling force shutdown command invalid state AwaitingChannelReady(AwaitingChannelReadyFlags(OUR_CHANNEL_READY))")
```

Full assertion failure:

```text
force shutdown should be allowed from AwaitingChannelReady when a local commitment tx exists, got Err("Invalid state: Handling force shutdown command invalid state AwaitingChannelReady(AwaitingChannelReadyFlags(OUR_CHANNEL_READY))")
```

## Expected behavior

`shutdown_channel(force=true)` should be allowed from `AwaitingChannelReady(OUR_CHANNEL_READY)` when the local node has a publishable latest commitment transaction.

The expected state transition is:

```text
AwaitingChannelReady(OUR_CHANNEL_READY)
  -- force shutdown with local commitment tx -->
ShuttingDown(WAITING_COMMITMENT_CONFIRMATION)
```

This preserves the meaning of `ChannelReady` as a bilateral usable state, while still allowing the local node to exit safely when the peer never sends `ChannelReady` or never comes back online.

chenyukang:
Fixes https://github.com/nervosnetwork/fiber/issues/1545

Allow channel with status of `AwaitingChannelReady` and local commitment to shutdown force.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -2216,6 +2216,15 @@ where
                         &flags
                     );
                 }
+                ChannelState::AwaitingChannelReady(flags)
+                    if flags.contains(AwaitingChannelReadyFlags::OUR_CHANNEL_READY)
+                        && state.latest_commitment_transaction.is_some() =>
+                {
+                    debug!(
+                        "Handling force shutdown command in AwaitingChannelReady state, flags: {:?}",
+                        &flags
+                    );
+                }
                 _ => {
                     return Err(ProcessingChannelError::InvalidState(format!(
                         "Handling force shutdown command invalid state {:?}",
@@ -9971,7 +9980,9 @@ impl ChannelActorState {
         match self.state {
             ChannelState::ShuttingDown(flags)
                 if flags.contains(ShuttingDownFlags::WAITING_COMMITMENT_CONFIRMATION) => {}
-            ChannelState::ChannelReady | ChannelState::ShuttingDown(..)
+            ChannelState::ChannelReady
+            | ChannelState::ShuttingDown(..)
+            | ChannelState::AwaitingChannelReady(_)
                 if force && !close_by_us => {}
             _ => {
                 return Err(ProcessingChannelError::InvalidState(format!(
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -2659,7 +2659,9 @@ where
                 for (_pubkey, channel_id, channel_state) in self.store.get_channel_states(None) {
                     if matches!(
                         channel_state,
-                        ChannelState::ChannelReady | ChannelState::ShuttingDown(..)
+                        ChannelState::ChannelReady
+                            | ChannelState::ShuttingDown(..)
+                            | ChannelState::AwaitingChannelReady(_)
                     ) {
                         if let Some(actor_state) = self.store.get_channel_actor_state(&channel_id) {
                             let funding_lock_script = state
@@ -4014,7 +4016,9 @@ where
 
         if !matches!(
             state.state,
-            ChannelState::ChannelReady | ChannelState::ShuttingDown(..)
+            ChannelState::ChannelReady
+                | ChannelState::ShuttingDown(..)
+                | ChannelState::AwaitingChannelReady(_)
         ) {
             return;
         }
--- a/crates/fiber-lib/src/fiber/tests/channel.rs
+++ b/crates/fiber-lib/src/fiber/tests/channel.rs
@@ -1,7 +1,7 @@
 use crate::ckb::tests::test_utils::{
     complete_commitment_tx, MockChainActorMiddleware, MockChainActorState,
 };
-use crate::ckb::{CkbChainMessage, FundingContext, FundingTx};
+use crate::ckb::{CkbChainMessage, FundingContext, FundingTx, GetShutdownTxResponse};
 use crate::fiber::channel::{
     funding_timeout_check_delay, merge_external_funding_witnesses, AddTlcResponse, ChannelActor,
     ChannelActorMessage, ChannelActorState, ChannelActorStateStore, ChannelOpenRecordStore,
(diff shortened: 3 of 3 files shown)


gpBlockchain:
The initiator-side fix looks effective, but the peer-side state transition is not covered and currently fails when the peer is also in `AwaitingChannelReady`.

I reproduced the following case on `3b8d1109`:

1. Node B is in `AwaitingChannelReady(OUR_CHANNEL_READY)` and force-closes with its local commitment.
2. Node A is in `AwaitingChannelReady(THEIR_CHANNEL_READY)`.
3. Node A receives `ClosingTransactionConfirmed(force=true, close_by_us=false)`.

Expected:

```text
Closed(UNCOOPERATIVE_REMOTE | WAITING_ONCHAIN_SETTLEMENT)
```

Actual:

```text
InvalidState(
  "Expecting commitment transaction confirmed event in unexpected state
   AwaitingChannelReady(AwaitingChannelReadyFlags(THEIR_CHANNEL_READY))
   force true close_by_us false"
)
```

The guard in `update_close_transaction_confirmed` currently accepts a remote force-close only from `ChannelReady` or `ShuttingDown`:

```rust
ChannelState::ChannelReady | ChannelState::ShuttingDown(..)
    if force && !close_by_us => {}
```

It likely also needs to accept `ChannelState::AwaitingChannelReady(_)` for a confirmed remote force-close. A confirmed on-chain commitment already proves the funding output has been closed, regardless of whether the `ChannelReady` handshake completed.

Could we add a two-sided integration regression that asserts:

- initiator: `Closed(UNCOOPERATIVE_LOCAL | WAITING_ONCHAIN_SETTLEMENT)`
- peer: `Closed(UNCOOPERATIVE_REMOTE | WAITING_ONCHAIN_SETTLEMENT)`
- both sides record the same `shutdown_transaction_hash`
- the peer-side transition also works after restart / through the persisted-state path?

gpBlockchain:
lgtm

Issue #1612, opened 2026-08-07
Uncommitted RemoteRemoved TLC is skipped during on-chain timeout reconciliation


gpBlockchain:
## Scenario

Topology: `A-B-C-D`, payment direction `A -> D`. B force-closes the `B-C` channel, which is still in `ShuttingDown(WAITING_COMMITMENT_CONFIRMATION)` when D's TLC expires.

C propagates `RemoveTlc` to B. B receives the peer message and immediately calls `set_offered_tlc_removed()`, changing the downstream TLC to `RemoteRemoved` and setting `removed_reason`.

The subsequent `CommitmentSigned` is rejected because `B-C` is already in the shutdown state with `WAITING_COMMITMENT_CONFIRMATION`. The remove handshake therefore never reaches `RemoveAckConfirmed` or `apply_remove_tlc_operation`, and B never propagates the result to the upstream `A-B` TLC.

## Failure

After the TLC is settled on-chain, timeout reconciliation skips it because `removed_reason` is already present:

```rust
.filter(|tlc| tlc.removed_reason.is_none())
```

`has_unresolved_onchain_tlcs()` also treats every TLC with `removed_reason.is_some()` as already resolved, even when the off-chain remove commitment handshake was not confirmed.

As a result, the B-C actor considers on-chain reconciliation complete and stops, while the upstream A-B offered TLC remains permanently present. The integration test `test_mid_node_shutdown_when_d_expiry` eventually times out waiting for A's offered TLC balance to become zero.

The issue is intermittent because it depends on whether C's `RemoveTlc` reaches B while B-C is still shutting down. If the B-C closing transaction is confirmed first, the peer message is dropped for the closed channel and B's local `removed_reason` remains empty, so the on-chain timeout path can reconcile the TLC.

## Failure sequence

```mermaid
sequenceDiagram
    autonumber
    participant A
    participant B
    participant C
    participant D
    participant Chain as On-chain settlement

    A->>B: AddTlc
    B->>C: AddTlc
    C->>D: AddTlc

    B->>B: Force-close B-C
    Note over B,C: B-C = ShuttingDown<br/>WAITING_COMMITMENT_CONFIRMATION

    D-->>C: TLC expires / RemoveTlc
    C-->>B: RemoveTlc

    B->>B: Mark TLC as RemoteRemoved
    B->>B: removed_reason = Some(...)

    C--xB: CommitmentSigned is rejected
[...]

Issue #1611, opened 2026-08-07
Terminal 0xfe settlement can leave an upstream TLC unreconciled after #1583


gpBlockchain:
## Scenario

Topology: `A-B-C-D`, payment direction `A -> D`. The `B-C` channel is force-closed.

After the TLC expires, `C-D` completes `RemoveTlc` off-chain. The `B-C` TLC is settled on-chain, after which B is expected to propagate `RemoveTlc` to `A-B`.

The TLC on `B-C` is offered by B and received by C. After C claims its final-party balance, the remaining commitment-cell capacity, including the expired TLC, belongs to B. B can therefore complete settlement using terminal unlock `0xfe`.

## Intermittent failure

The result depends on B/C watchtower transaction ordering.

In the failed run:

1. C first sends final-party `0xff` transaction `0x43f2822701f2ea18141cd7cbc29bbf4737feedd57a809067b40a6a7f2cb342f4`.
2. B builds explicit TLC timeout transaction `0xbc365cd3b7ae9c083b40c59a232ae07a64e455406b042e2a25f2831acffec992` with `unlock_type=0`.
3. The transaction is rejected: `PoolRejectedRBF: current fee 869, expected >= 2172`.
4. B then consumes C's settlement output using terminal `0xfe` transaction `0xcbc47f741c565657c9d053e6bf82cc6d394db5def6f6d666d97e7b3e2969c8bc`.

The terminal `0xfe` transaction settles all remaining capacity, including B's expired TLC, so the TLC is settled without a preimage.

After #1583, however, the watchtower skips final-party unlocks:

```rust
if unlock.unlock_type >= 0xFE {
    continue;
}
```

No `SettledWithoutPreimage` resolution is recorded. The B-C channel repeatedly logs:

```text
on-chain settlement reconciliation incomplete; keeping ONCHAIN_SETTLEMENT_CONFIRMED
```

B does not propagate `RemoveTlc` to A-B, and A's offered TLC remains locked.

In the successful run, B's explicit `unlock_type=0` transaction enters the pool before C's final-party transaction. The TLC receives an explicit timeout resolution, B-C reconciliation completes, and B propagates `RemoveTlc` to A-B.

[...]

PR #1618, opened 2026-08-07
fix: complete on-chain TLC reconciliation for terminal sweeps and uncommitted removes (#1611, #1612)

jjyr:
## Summary

Two fixes for on-chain TLC reconciliation on force-closed channels, addressing two sister issues reported against the exact-TLC settlement model introduced in #1583.

### 1. Terminal final-party sweep leaves pending TLCs unreconciled (#1611)

After #1583 the watchtower skips every final-party unlock (`0xFE`/`0xFF`). When such an unlock consumes the last commitment cell (no successor commitment output), the terminal sweep carries every still-pending TLC off-chain without recording an exact settlement proof. The channel then keeps `ONCHAIN_SETTLEMENT_CONFIRMED` forever, reconciliation never completes, and `RemoveTlc` is never propagated upstream.

**Fix (watchtower):** detect a terminal settlement (no successor commitment output) in `process_watched_settlement_tx` and, when a final-party unlock is observed on such a transaction, persist an exact no-preimage settlement (`OnChainTlcSettlement { preimage: None }`) for every remaining tracked TLC. Intermediate final-party unlocks (with a successor commitment cell) keep the previous behavior: pending TLCs stay tracked.

### 2. Uncommitted `RemoteRemoved` TLC is skipped during on-chain timeout reconciliation (#1612)

When a peer `RemoveTlc` arrives while the channel is shutting down with `WAITING_COMMITMENT_CONFIRMATION`, `set_offered_tlc_removed` marks the TLC `RemoteRemoved` but the remove commitment handshake never completes, so `removed_confirmed_at` stays `None` and the upstream `RemoveTlc` is never propagated. On-chain timeout reconciliation skipped such TLCs because `collect_onchain_timeout_settled_tlcs` filtered on `removed_reason.is_none()`, while `get_expired_offered_tlcs` already scopes to unconfirmed removes via `removed_confirmed_at.is_none()`. The channel then finalized settlement and stopped, leaving the upstream offered TLC permanently present.

**Fix (fiber):** drop the redundant `removed_reason` filter so uncommitted removed TLCs are collected and relayed upstream once their on-chain timeout settlement is observed. Guard the `OriginPayer` path in both the live-actor and no-live-actor reconciliation flows against re-marking an already removed TLC (`set_offered_tlc_removed` asserts `Committed`).

## TDD

- `terminal_final_party_unlock_records_settled_without_preimage` fails on `e6cb7ac7` (0 settlement records) and passes with the watchtower fix; covers both `0xFE` and `0xFF`.
- `collect_timeout_settled_collects_uncommitted_removed` fails on `e6cb7ac7` (0 collected) and passes with the fiber fix.
- `collect_timeout_settled_skips_already_removed` now pins the confirmed-remove case (`removed_confirmed_at` set).

## Validation

- `cargo nextest run -p fnn -p fiber-bin --no-fail-fast` — 1200 passed, 8 skipped

## Notes

A symmetric fulfill variant of #1612 (uncommitted removed TLC settled on-chain with a preimage is skipped by `collect_onchain_fulfilled_tlcs` via `can_reconcile_onchain_fulfillment`) is out of scope and can be followed up separately.

### DIFF
--- a/crates/fiber-lib/src/fiber/channel.rs
+++ b/crates/fiber-lib/src/fiber/channel.rs
@@ -3082,7 +3082,16 @@ where
                         ))
                         .expect(ASSUME_NETWORK_ACTOR_ALIVE);
                     if let TLCId::Offered(id) = tlc.tlc_id {
-                        state.tlc_state.set_offered_tlc_removed(id, reason);
+                        // A peer RemoveTlc may already have marked this TLC removed without the
+                        // commitment handshake completing (issue #1612). Only mark it again when
+                        // it was never removed: `set_offered_tlc_removed` asserts Committed.
+                        if state
+                            .tlc_state
+                            .get(&TLCId::Offered(id))
+                            .is_some_and(|t| t.removed_reason.is_none())
+                        {
+                            state.tlc_state.set_offered_tlc_removed(id, reason);
+                        }
                     }
                 }
             }
--- a/crates/fiber-lib/src/fiber/network.rs
+++ b/crates/fiber-lib/src/fiber/network.rs
@@ -3564,8 +3564,17 @@ where
                         ))
                         .expect(ASSUME_NETWORK_ACTOR_ALIVE);
                     if let TLCId::Offered(id) = tlc.tlc_id {
-                        actor_state.tlc_state.set_offered_tlc_removed(id, reason);
-                        actor_state_changed = true;
+                        // A peer RemoveTlc may already have marked this TLC removed without the
+                        // commitment handshake completing (issue #1612). Only mark it again when
+                        // it was never removed: `set_offered_tlc_removed` asserts Committed.
+                        if actor_state
+                            .tlc_state
+                            .get(&TLCId::Offered(id))
+                            .is_some_and(|t| t.removed_reason.is_none())
+                        {
+                            actor_state.tlc_state.set_offered_tlc_removed(id, reason);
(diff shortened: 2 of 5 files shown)


chainTe:
Security review (GPT-5.6 Codex) of
`e6cb7ac7770b1798a1ad5dfb9a8f4ae5db52036f..bac1dc6863643fe2eebc42b7f8dc24e2ab7d45b7`
found two low-severity crash-consistency gaps:

1. A forwarded TLC can reach terminal downstream persistence before its
   live-upstream removal is durably acknowledged. A process failure in that
   interval can lose automatic replay and leave upstream value locked until
   manual recovery or force close.
2. An origin-payer TLC can reach terminal channel persistence before the
   fire-and-forget payment failure event is durably applied, leaving stale
   `Inflight` payment/retry state after a crash.

Both paths require a narrow victim process-failure window that is not shown to
be peer-controlled. The robust fix is an acknowledged or persistent outbox
handoff before clearing terminal reconciliation state, with idempotent startup
replay.

Focused tests `collect_timeout_settled_collects_uncommitted_removed` and
`terminal_final_party_unlock_records_settled_without_preimage` passed with the
`sqlite` feature. Broad tests were skipped; CI is expected to cover them. No
deterministic crash-injection test was run.

gpBlockchain:
lgtm

PR #1629, opened 2026-08-19
fix: finalize on-chain settlement for TLCs with a non-matching preimage

quake:
## Problem

When a force-closed channel is settled on-chain, the watchtower records the settlement preimage for each consumed TLC. If that preimage does not hash to the TLC's full 32-byte payment hash, `resolve_onchain_tlc` classifies the exact settlement record as `Unknown`.

`Unknown` records are ignored by both the fulfillment and the timeout collectors, so the affected TLC is never reconciled. As a result the channel stays in `ONCHAIN_SETTLEMENT_CONFIRMED` and cannot complete its on-chain settlement.

## Fix

Treat an exact settlement record whose preimage fails the full-hash check as `SettledWithoutPreimage` rather than `Unknown`. The on-chain output has already been spent, so the TLC is settled; it simply can no longer be fulfilled upstream. This lets the timeout collector reconcile the TLC and finalize the channel.

## Test

Renamed `resolve_returns_unknown_when_preimage_mismatches` to `resolve_returns_settled_without_preimage_when_preimage_mismatches`.

```
cargo test -p fnn --features sqlite fiber::tests::onchain_tlc_reconcile::
```

passes (15/15).

### DIFF
--- a/crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs
+++ b/crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs
@@ -141,10 +141,10 @@ pub(crate) fn resolve_onchain_tlc(
                 return OnChainTlcResolution::Fulfilled(preimage);
             }
             warn!(
-                "Ignoring invalid on-chain preimage for channel {:?} tlc {:?} tx {:?}: derived hash {:?}, expected {:?}",
+                "On-chain preimage for channel {:?} tlc {:?} tx {:?} does not match the full payment hash (derived {:?}, expected {:?}); treating as settled without preimage",
                 channel_id, tlc_id, settlement.tx_hash, discovered_payment_hash, payment_hash
             );
-            OnChainTlcResolution::Unknown
+            OnChainTlcResolution::SettledWithoutPreimage
         }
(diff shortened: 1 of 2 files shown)


chainTe:
## Security Review

Result: 1 confirmed High-severity finding.

Reviewed range: `189b249a5739a0a54272680692012272e1e75d00..cda91e423614766f8f42f4a55e25f071e02068b4`

Review model: Codex (GPT-5)

Validation:

- Both changed files were reviewed in full, with the watchtower producer, channel-state consumers, and pinned commitment-lock semantics traced.
- The focused mismatch-classification and shared-prefix exact-index tests passed.
- Broad existing test suites were intentionally skipped under the PR-diff review policy; CI is expected to run them.

The changed branch treats a preimage-bearing on-chain observation that fails the complete payment-hash check as `SettledWithoutPreimage`. That value is affirmative input to the timeout/failure collectors, so it can propagate payment failure even though the on-chain unlock used the preimage path. Please retain `Unknown` or introduce a distinct anomaly state that cannot enter timeout/failure consumers, and add a source-realistic shared-prefix, `with_preimage = true`, post-mutation regression test.

PR #1630, opened 2026-08-19
fix: include offered LocalAnnounced TLCs in on-chain reconciliation

jjyr:
## Summary

Offered TLCs in `OutboundTlcStatus::LocalAnnounced` are already included in the **remote** signed commitment (`build_settlement_data(for_remote=true)` / `commitment_signed_tlcs`: `LocalAnnounced => for_remote`). A peer can withhold `RevokeAndAck`, broadcast that commitment, and spend the TLC on-chain.

On-chain reconciliation previously treated offered TLCs as reconcilable only in `Committed`:

- `can_reconcile_onchain_fulfillment` fed fulfillment, received-timeout, and `has_unresolved_onchain_tlcs`
- `collect_onchain_timeout_settled_tlcs` reused `get_expired_offered_tlcs`, which also skips `LocalAnnounced`

The on-chain outcome was therefore dropped. An origin payer (or a closed upstream) never learned it. If that TLC was the only pending one, `has_unresolved_onchain_tlcs` returned false and the channel could finish `WAITING_ONCHAIN_SETTLEMENT` too early.

Inbound already covers the matching wait-ack state (`AnnounceWaitAck | Committed`). Outbound has no separate wait-ack variant after `CommitmentSigned` is sent, so it stays `LocalAnnounced`. This change does not add a new persisted enum variant.

## Fix

- **Collectors:** accept offered `LocalAnnounced | Committed`, symmetric with inbound `AnnounceWaitAck | Committed`. The real gate is still the watchtower exact settlement record.
- **Timeout collector:** stop using `get_expired_offered_tlcs`. Filter offered TLCs with `removed_confirmed_at.is_none()` and `expiry < expect_expiry` only, so `LocalAnnounced` is included when an on-chain timeout record exists.
- **`has_unresolved_onchain_tlcs`:** count offered `LocalAnnounced` as unresolved only on `Closed(UNCOOPERATIVE_REMOTE)`. A local force-close spends the local commitment, which does not contain these TLCs; treating them as unresolved would stall settlement forever.
- **`set_offered_tlc_removed`:** allow `LocalAnnounced | Committed`, matching `set_received_tlc_removed` (`AnnounceWaitAck | Committed`). Otherwise origin-payer apply panics.

`get_expired_offered_tlcs` is unchanged. It is still used by live-channel CheckChannels to force-close on expired offered TLCs that are already in the **local** commitment. Including `LocalAnnounced` there would force-close TLCs that are not in the local commitment.

## TDD

These tests fail on `c991b44e` and pass with the fix:

- `collect_fulfilled_includes_offered_local_announced`
- `collect_timeout_includes_expired_offered_local_announced`
- `collect_timeout_includes_forwarded_local_announced`
- `has_unresolved_keeps_local_announced_on_remote_force_close`
- `set_offered_tlc_removed_accepts_local_announced`

`has_unresolved_ignores_local_announced_on_local_force_close` already passed before the fix and stays as a lock so local force-close cannot hang.

The old `collect_skips_removed_and_uncommitted_tlcs` pin (which encoded the bug as expected) is split into:

- `collect_skips_removed_offered_tlcs`
- `collect_skips_inbound_remote_announced`

## Validation

- `cargo nextest run --no-fail-fast` — 1366 passed, 8 skipped
- `cargo fmt --all`

## Notes

Inbound `RemoteAnnounced` is the mirror case (present in the local commitment, still excluded from reconciliation). Out of scope here.

### DIFF
--- a/crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs
+++ b/crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs
@@ -5,7 +5,8 @@
 
 use crate::fiber::channel::{ChannelActorState, ChannelActorStateStore};
 use fiber_types::{
-    Hash256, HashAlgorithm, InboundTlcStatus, OutboundTlcStatus, RemoveTlcReason, TLCId, TlcInfo,
+    ChannelState, CloseFlags, Hash256, HashAlgorithm, InboundTlcStatus, OutboundTlcStatus,
+    RemoveTlcReason, TLCId, TlcInfo,
 };
 use serde::{Deserialize, Serialize};
 use tracing::warn;
@@ -253,9 +254,15 @@ pub(crate) fn collect_onchain_timeout_settled_tlcs(
     expect_expiry: u64,
 ) -> Vec<OnChainTimeoutSettledTlc> {
     let channel_id = state.get_id();
+    // Live-channel expiry handling still uses `get_expired_offered_tlcs`, which omits
+    // LocalAnnounced TLCs because they are not in the local commitment. On-chain timeout
+    // reconciliation must include them: a signed remote commitment already contains the TLC.
     state
         .tlc_state
-        .get_expired_offered_tlcs(expect_expiry)
+        .offered_tlcs
+        .tlcs
+        .iter()
+        .filter(|tlc| tlc.removed_confirmed_at.is_none() && tlc.expiry < expect_expiry)
         .filter_map(|tlc| {
             if !matches!(
                 resolve_onchain_tlc(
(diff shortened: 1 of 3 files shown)


gpBlockchain:
lgtm

chainTe:
## Security Review

Result: 1 confirmed finding (High)
Reviewed range: `c991b44e75994585a27eace794f37d2d64567df9..5e61edcf5c0c3cfc1e67d3e71baa169a4be9ac65`
Review model: GPT-5 Codex

Validation:

- `cargo test -p fnn --features sqlite --lib onchain_tlc_reconcile`: 23 passed, 0 failed, 1201 filtered out.

Notes:

- `has_unresolved_onchain_tlcs` treats every offered `LocalAnnounced` TLC as present after any remote force-close. Before `RevokeAndAck`, however, the peer can still publish the preceding unrevoked remote commitment, whose immutable settlement snapshot omits that TLC. The watchtower then correctly produces no exact settlement record, while both live and restored finalization retain `WAITING_ONCHAIN_SETTLEMENT` indefinitely. This can wedge the closed channel and linked origin/forwarding state until manual recovery.
- Please bind reconciliation eligibility to the exact snapshot selected for the observed commitment number, and add a regression that distinguishes publication of the pending commitment from publication of the preceding commitment.
- An adjacent received `AnnounceWaitPrevAck` local-close case also lacks regression coverage, but its predicate is pre-existing and was not counted as a diff finding.

PR #1636, opened 2026-08-27
feat: add default config helper for fiber-js

joii2020:
## Summary

Add `getDefaultConfig` to make Fiber configuration easier for `fiber-js` developers. By building on the bundled defaults, applications can stay aligned with configuration changes introduced by Fiber updates and catch incompatible customizations earlier during development.

## Testing

This PR also introduces a Jest-based test framework for `fiber-js` and uses it to test default config generation for both mainnet and testnet.


quake:
@joii2020 how to setup a dev chain for local dev after this PR? could you add an example code in README?

PR #1643, opened 2026-09-01
feat: add liquidity management Loop In and Loop Out

### LINKED ISSUE #1541: Design: Fiber liquidity management milestones
> *This was generated by AI during triage.*

## Summary

Track the approved milestone design for adding lnd Loop-inspired liquidity management to Fiber.

## Product Direction

Fiber should support Loop In and Loop Out style liquidity movement between Fiber channels and CKB on-chain assets.

Key decisions:

- Support CKB and whitelisted UDT assets from the first version.
- Use an asset registry / whitelist instead of arbitrary user-supplied UDT type scripts.
- Add an independent `liquidity` module rather than merging this into `cch`.
- Support dual-role FNN nodes: client capability always available, provider mode optional and disabled by default.
- Do not change Fiber p2p messages in the first product milestones.
- Use RPC for swap negotiation, while Fiber p2p continues to carry normal payments/TLCs.
- Defer provider discovery, quote gossip, and liquidity-specific p2p messages until a later market-mode milestone.

### M0: Protocol And Product Spec

Define Loop In / Loop Out semantics, dual-role model, CKB + UDT whitelist asset model, quote schema, swap order state machine, fee model, timeout/refund/recovery rules, and initial RPC draft.

### M1: Asset Registry And On-Chain Swap Primitive

Implement `Asset::Ckb` / `Asset::Udt`, provider whitelist configuration, per-asset limits, CKB + UDT swap cell support, claim/refund transaction builders, and chain-level tests.

### M2: Swap Store And Recovery Framework

Persist swap orders and state transitions, validate transitions, recover non-terminal orders on startup, and expose swap history queries.

### M3: Manual Loop Out

Ship the first end-to-end product: Fiber channel balance to on-chain CKB/UDT, with quote/execution RPCs, payment integration, payout integration, and recovery handling.

### M4: Manual Loop In

Implement on-chain CKB/UDT to Fiber channel balance, including lock watching, provider Fiber payment, claim/refund, and restart-safe recovery.

### M5: Integrated Provider Mode

Add optional provider mode with per-asset capacity, quote/fee policy, concurrent order limits, risk budgets, order recovery, and metrics.

### M6: Client-Side Liquidity Planner

Add dry-run recommendations based on channel liquidity, rebalance-first planning, Loop In/Out cost estimation, and human-readable decision reasons.

### M7: Automated Liquidity Management

Add policy-driven execution with channel target ratios, max fee budgets, cooldowns, spend limits, automatic rebalance/Loop In/Loop Out, and audit logs.

### M8: Productization

Add CLI support, RPC docs, Biscuit auth rules, swap history, fee breakdowns, operator metrics, and clear errors.

### M9: Multi-Provider Network Mode

Explore provider discovery, quote comparison, reputation, capacity advertisement, anti-spam, privacy review, and possible p2p/gossip extensions.

## Recommended First Phase

Implement `M0 -> M1 -> M2 -> M3` first. Manual Loop Out should validate the shared asset model, on-chain primitive, Fiber payment integration, provider flow, and restart recovery before Loop In and automation are added.

## M0 Decisions Still Required

- Exact CKB and UDT swap cell script design.
- Exact preimage ownership and reveal sequence for Loop In and Loop Out.
- Confirmation policy by asset/network.
- Whether quote negotiation uses existing JSON-RPC only or also an HTTP API.
- How much CCH tracker/state-machine code should be reused.
- Storage key layout and migration requirements.

quake:
## Summary
- add persisted liquidity assets, quotes, swaps, provider controls, RPCs, and restart-safe actor orchestration
- implement CKB Loop In and Loop Out chain/payment flows with the deployed liquidity-lock artifact and exact outpoint spend tracing
- add real-process Bruno E2E coverage for CKB and UDT flows, provider-mode gating, malformed lock rejection, lock-time refund behavior, and provider restart recovery
- harden Loop Out payout authorization, fee caps, invoice recovery, tracer registration, dynamic CI ports, and failure diagnostics

## Test Plan
- [x] In-process CKB Loop Out, restart, and malicious payout E2Es
- [x] Bruno helper smoke suite: 7 requests, 10 tests, 20 assertions
- [x] Mutator workspace tests: 23 passed
- [x] `tests/deploy/udt-init` tests: 3 passed
- [x] Native liquidity/RPC suites and CKB Loop Out E2Es passed locally
- [x] `cargo clippy --all-targets --all-features -p fnn -p fiber-bin -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `make check-migrate`
- [x] `make check-dirty-rpc-doc`

## Follow-ups
- UDT Loop Out payment propagation remains blocked by the existing payment-layer limitation and is documented in the suite README.
- Refund `RefundPending` actor transition remains an M0 limitation and is documented in the refund suite README.

Closes #1541


chainTe:
## Security Review

Result: 2 confirmed finding(s)
Reviewed range: `c991b44e75994585a27eace794f37d2d64567df9..92ed7dd597234a6fbb657a5fd54b7737d6423e1d`
Review model: `grok-4.6`

Validation:
- Review was a local static source trace of the PR diff and the coupled `liquidity-lock` source at `fiber-scripts@4da5d299`.

Notes:
- **Critical:** Loop In quote/import copies the client invoice `payment_hash` into lock args without requiring `HashAlgorithm::CkbHash`. `liquidity-lock` claim always checks `blake2b_256(preimage)`. A Loop In client can present a Sha256 invoice, receive the Fiber payment, then refund the on-chain lock because the provider cannot claim. Bind Loop In invoices to `CkbHash` (including the omitted default) before persisting a quote.
- **High:** Client Loop Out Fiber payment hardcodes `udt_type_script: None` while UDT quotes and provider invoices carry the UDT type script. Until SendPayment uses the quoted asset, fail closed on UDT Loop Out execution.
- Quote envelopes and swap RPC records omit preimages. Observed payout live-cell checks match lock args, amount, and claimant/refund hashes. Missing Biscuit rules for liquidity methods fail closed when RPC auth is enabled.
- `liquidity-lock` is not on `fiber-scripts` main; this review used the commit recorded by `tests/deploy/contracts/README.md`.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+        if let Some(payout_tx) = payout_tx {
+            self.store
+                .clear_payout_validation_failure_context(&swap_id, &payout_tx.tx_hash, now_ms)
+                .map_err(map_store_error)?;
+        }
+        if authorization == ClientPayoutAuthorization::PayoutConfirmation {
+            mark_client_payout_locked(&self.store, swap_id, now_ms)?;
+        }
+        self.watched_payout_swaps.remove(&swap_id);
+        match self.start_client_payment_dispatch(authorized, now_ms, myself.clone()) {
```
chenyukang, line 1150: **[P1] Check the remaining refund window before sending payment**

`validate_observed_loop_out_payout` checks that the cell is live and matches the quote, but neither it nor this dispatch path checks whether `refund_after_lock_time` is already mature or leaves enough time to claim. Quote import also accepts a future `expires_at` with an already-mature refund time. A previously valid quote can reach the same state during confirmation delays or restart recovery.

The provider can therefore leave a refundable payout live until validation, then race a refund against Fiber payment settlement. If the refund confirms first, the client has paid but cannot claim the spent cell even after receiving the preimage.

Please validate the refund `since` and a sufficient payment-plus-claim safety window against chain time immediately before dispatch, including recovery, and bound the payment/TLC lifetime accordingly. Focused local tests confirmed acceptance of a mature refund time and payment dispatch on recovery; a full-chain loss scenario was not run.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+    }
+
+    async fn start_provider_loop_in_payment(
+        &mut self,
+        swap: LiquiditySwapRecord,
+        myself: ActorRef<LiquidityActorMessage>,
+    ) -> Result<(), LiquidityLoopOutError> {
+        let quote = self.quote_terms(&swap.quote_id)?;
+        ensure_loop_in_quote_terms(&quote)?;
+        let request = prepare_provider_loop_in_payment(&self.store, quote)?;
```
chenyukang, line 1983: **[P1] Revalidate the Loop In lock before sending the provider payment**

This path prepares and sends the payment without reloading the lock cell or checking its remaining refund window. The validation in `accept_provider_loop_in` can become stale: after the provider accepts the lock, it can go offline, the client can refund once the lock matures, and startup recovery can resume an `OnchainLocked` swap here. The provider then pays the still-valid invoice even though the collateral has already been spent, so the client can receive both the refund and the Fiber payment.

Please revalidate the exact live outpoint and its quote terms immediately before dispatch, including recovery, and require enough remaining time for payment and claim confirmation. A focused test using the real actor/store confirmed that recovery dispatches payment without invoking lock validation; the chain adapter was mocked, so this is not a full-chain loss reproduction.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+                        myself.clone(),
+                    )
+                    .await;
+            }
+            LiquidityActorMessage::PaymentSettled(swap_id, preimage) => {
+                if let Err(error) = state
+                    .handle_payment_settled(swap_id, preimage, myself.clone())
+                    .await
+                {
+                    tracing::warn!(?swap_id, %error, "ignoring loop out payment settled continuation");
```
chenyukang, line 388: **[P1] Keep nonterminal payment and claim recovery active**

A transient error while broadcasting the claim reaches this log-only branch after the Fiber payment has already settled, with no retry scheduled. There is a related gap in `reconcile_loop_out_payment`: after 60 polls it sends `PaymentRecoveryFinished` even if the payment remains in flight, and that handler only removes the active guard. A later settlement is therefore never observed and no claim is attempted. `ResumeNonTerminal` is only invoked at startup in production.

Either path can leave the paying party's funds unclaimed until the counterparty can refund. Please schedule durable retries or an ongoing recovery scan until an authoritative terminal result is known; continue querying uncertain payments rather than blindly resending them. Focused actor/store tests confirmed both the lack of claim retries over an hour of virtual time and the missed settlement after polling exhaustion, using mocked payment/chain boundaries.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+                        swap.swap_id,
+                        "payout recovery missing persisted outpoint".to_string(),
+                    );
+                    return Ok(false);
+                }
+                if self.watched_payout_swaps.contains(&swap.swap_id) {
+                    return Ok(false);
+                }
+                self.chain
+                    .watch_payout_lock(swap.swap_id, myself.clone())
```
chenyukang, line 1605: **[P2] Rebroadcast persisted Planned payouts during recovery**

The provider accept path persists the payout transaction identity, signed bytes, and swap outpoint before calling `broadcast_payout_lock`. If the process crashes after those writes but before broadcast, recovery reaches this branch and only installs a watcher. The transaction is still `Planned` and was never submitted, so it cannot confirm and the swap remains in `PayoutPending` indefinitely. Repeating provider accept also returns the existing swap without broadcasting it.

Please reconcile the transaction's chain status and, when appropriate, rebroadcast the already-persisted signed transaction before watching it. Do not fund a new payout for the same swap. A focused test using the real actor/store and a recording chain adapter confirmed that recovery of this state calls the watcher but never the broadcast method.

docs/superpowers/plans/2026-08-26-fiber-liquidity-loop-out-hardening.md
```diff
@@ -0,0 +1,401 @@
...
+cargo nextest run -p fnn --features rocksdb -E 'test(/loop_out_.*(principal|fee_cap)/)' --no-fail-fast
+cargo nextest run -p fnn --features rocksdb -E 'test(liquidity::payment)' --no-fail-fast
+cargo nextest run -p fnn --features rocksdb -E 'test(liquidity::actor)' --no-fail-fast
+cargo fmt --all -- --check
+git add crates/fiber-lib/src/liquidity/types.rs crates/fiber-lib/src/liquidity/payment.rs crates/fiber-lib/src/liquidity/actor.rs
+git commit -m "fix: separate Loop Out principal from routing budget"
+```
+
+---
+
```
chenyukang, line 221: superpowers document should not be checked in?

crates/fiber-lib/src/liquidity/chain.rs
```diff
@@ -0,0 +1,6115 @@
...
+    let payout_capacity = quote
+        .capacity_requirement_ckb
+        .checked_sub(quote.onchain_fee_estimate_ckb)
+        .filter(|capacity| {
+            quote.asset.udt_type_script.is_some() || u128::from(*capacity) >= quote.amount
+        })
+        .unwrap_or(quote.capacity_requirement_ckb);
+    let mut output = packed::CellOutput::new_builder()
+        .capacity(payout_capacity)
+        .lock(quote.claimant_lock.clone());
```
chenyukang, line 2184: **[P1] Do not pay cell occupancy capacity as extra CKB principal**

When the asset registry permits CKB swaps below the liquidity-lock occupied capacity, the quote funds `max(occupied_capacity, amount)` plus the transaction fee, but the Fiber principal remains `amount + provider_fee`. This output then sends all funded capacity less the estimated fee to the claimant. With a 61 CKB quote and zero provider fee, the actual quote and claim builders produce a 193 CKB claim for a 61 CKB Fiber payment, overpaying the client by 132 CKB. Loop In has the reverse loss: the client funds about 193 CKB and the provider claims 193 CKB for a 61 CKB invoice.

Please reject CKB amounts below the required lock capacity, or explicitly define and enforce how excess occupancy capacity is returned to its funder. Focused tests using the actual quote actor and transaction builders confirmed both amounts; these were not full-chain loss reproductions.

crates/fiber-lib/src/liquidity/chain.rs
```diff
@@ -0,0 +1,6115 @@
...
+            tx.clone()
+        )
+        .map_err(|error| {
+            let failure_reason = format!(
+                "send tx actor call failed for loop in lock liquidity tx {tx_hash} from funding_tx {funding_tx}: {error}"
+            );
+            let _ = self.store.update_liquidity_chain_tx_status(
+                &quote.quote_id,
+                LiquidityChainTxRole::Payout,
+                fiber_types::LiquidityChainTxStatus::Rejected,
```
chenyukang, line 1057: **[P2] Keep timed-out lock broadcasts recoverable**

A `SendTx` actor timeout does not prove that CKB rejected the transaction: the node may already have accepted it while its success response is delayed beyond the 8-second timeout. This path nevertheless persists `Rejected` and returns before installing a tracer. `resume_loop_in_swap` then turns that status into terminal `Failed` on restart. Repeating `loop_in` is rejected by the state check, and later recovery scans skip the order despite its funds still being locked on-chain.

Please preserve an unknown/recoverable outcome for transport failures and reconcile the persisted transaction hash before marking the swap terminal, retaining the refund recovery path. A focused test using the real chain adapter, actor, and store, with a mock CKB actor accepting the transaction and returning success after 9 seconds, reproduced `Rejected -> Failed`, no tracer, and refusal to resume the order.

crates/fiber-bin/src/main.rs
```diff
@@ -180,6 +182,111 @@ async fn run() -> Result<(), ExitMessage> {
     }
 }
 
+#[cfg(test)]
```
chenyukang, line 185: move tests related into tests directory?

crates/fiber-bin/Cargo.toml
```diff
@@ -38,5 +38,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 nix = { version = "0.29", features = ["fs"] }
 
 [dev-dependencies]
+ckb-types = "1"
```
jjyr, line 41: We should move tests out of fiber-bin. then, remove this dependency

crates/fiber-lib/src/ckb/outpoint_tracing_actor.rs
```diff
@@ -0,0 +1,1022 @@
...
+    lock_script: packed::Script,
+    confirmations: u64,
+    generation: u64,
+    #[cfg(test)]
+    discovery: Option<DiscoveryFn>,
+}
+
+impl PollTask {
+    fn spawn(self) -> JoinHandle<()> {
+        ractor::concurrency::spawn(async move {
```
jjyr, line 294: It seems this feature can be installed on CKB Actor instead of  tracing actor, to reduce an abstraction layer.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+            asset_type_script.as_ref(),
+        )?;
+        let capacity_requirement_ckb = match asset.kind {
+            LiquidityAssetKind::Ckb => occupied_capacity.max(
+                u64::try_from(params.amount)
+                    .map_err(|_| LiquidityLoopOutError::GrossAmountOverflow)?,
+            ),
+            LiquidityAssetKind::Udt => occupied_capacity,
+        }
+        .checked_add(1_000)
```
jjyr, line 618: Is this 1000 CKB represents `onchain_fee_estimate_ckb`? should use a constant to represent this value.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+            asset,
+            amount: params.amount,
+            provider_fee: validated.provider_fee,
+            routing_fee_limit: validated.routing_fee_limit,
+            onchain_fee_estimate_ckb: 1_000,
+            capacity_requirement_ckb,
+            payment_hash,
+            payment_preimage: Some(preimage),
+            expires_at: validated.expires_at,
+            payout_deadline: validated.expires_at.saturating_add(10_000),
```
jjyr, line 633: Why add another 10000, use a constant to reference this value

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+    ) {
+        let generation = self
+            .payout_validation_retries
+            .get(&swap_id)
+            .copied()
+            .unwrap_or_default()
+            .checked_add(1)
+            .unwrap_or(1);
+        self.payout_validation_retries.insert(swap_id, generation);
+        self.spawn_job(async move {
```
jjyr, line 1211: we can use send_after to replace the spawn + sleep + send_message combo

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+        }
+
+        if attempt + 1 == PROVIDER_LOOP_OUT_PAYMENT_RECONCILE_MAX_RELOAD_ATTEMPTS {
+            persist_loop_out_payment_failure_context(
+                &store,
+                swap_id,
+                "provider payment observation exhausted while status remained in flight"
+                    .to_string(),
+            );
+            send_payment_recovery_finished(&myself, swap_id);
```
jjyr, line 2149: [P1] Payment observation stops without automatically resuming.
Both payment observers stop after 60 polling attempts. PaymentRecoveryFinished only clears the active-task guard; it does not schedule another check. A payment that settles later may go unnoticed, leaving the swap in PaymentInFlight and potentially delaying the payer’s on-chain claim. Recovery requires another externally triggered scan.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+                        .await
+                        .map_err(|error| LiquidityLoopOutError::Chain(error.to_string()))?;
+                    self.watched_claim_swaps.insert(swap_id);
+                }
+                return Ok(());
+            }
+            LoopOutClaimPlan::validate_payment_preimage(swap.payment_hash, preimage)?;
+            persist_client_loop_out_payment_preimage(&self.store, swap_id, preimage, now_ms())?;
+            self.active_payment_swaps.remove(&swap_id);
+            claim_client_loop_out(
```
jjyr, line 1314: [P1] A transient claim failure leaves an already-paid swap without automatic continuation.
Claim execution persists ClaimPending before broadcasting. If broadcasting fails, the error is logged but no retry is scheduled. Replaying PaymentSettled does not repair this: it attempts to transition the order back into PaymentSettled, which is invalid from ClaimPending. The handler should resume from the persisted state and retry the remaining claim operation.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+        {
+            if existing.role != LiquiditySwapRole::Provider
+                || existing.swap_kind != LiquiditySwapKind::LoopOut
+            {
+                return Err(LiquidityLoopOutError::Store(format!(
+                    "existing swap conflicts with provider loop out accept for quote {quote_id:?}: expected provider loop out, found {:?} {:?}",
+                    existing.role, existing.swap_kind
+                )));
+            }
+            let quote = self.quote_terms(&quote_id)?;
```
jjyr, line 1415: Provider Loop Out acceptance cannot recover from a partially completed attempt.
The problem spans three code paths:
1. The order is persisted before payout preparation succeeds.
   In [actor.rs (line 2572)](/Users/jjy/Workspace/fiber/crates/fiber-lib/src/liquidity/actor.rs:2572), the swap transitions to PayoutPending before reserve_payout_lock_outpoint() runs. If that operation fails before producing an outpoint, the order remains persisted with onchain_outpoint = None.
2. Retrying acceptance skips the unfinished work.
   In [actor.rs (line 1415)](/Users/jjy/Workspace/fiber/crates/fiber-lib/src/liquidity/actor.rs:1415), the existing-order branch only registers the provider invoice and returns the current swap response. It does not resume payout preparation, broadcasting, or watch registration.
3. Restart recovery does not repair this state either.
   In [actor.rs (line 1581)](/Users/jjy/Workspace/fiber/crates/fiber-lib/src/liquidity/actor.rs:1581), recovery detects a provider swap with no outpoint, records "payout recovery missing persisted outpoint", and returns without progressing.
Consequently, a temporary preparation failure can leave the order stuck even after the caller retries acceptance or restarts the node.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+            }
+            LiquidityActorMessage::PayoutConfirmed(swap_id) => {
+                if state.payout_validation_retries.contains_key(&swap_id)
+                    || state
+                        .definitive_payout_validation_failures
+                        .contains(&swap_id)
+                {
+                    return Ok(());
+                }
+                if let Err(error) = state
```
jjyr, line 351: [P2] Provider payout-confirmation failures are routed into a client-only retry path.
Any error from handle_payout_confirmed schedules RetryPayoutValidation, including errors for provider orders. That retry invokes authorize_client_payout, which immediately rejects non-client roles. Consequently, the log says “retrying,” but the provider continuation is never retried.

crates/fiber-lib/src/liquidity/actor.rs
```diff
@@ -0,0 +1,10737 @@
...
+                    &self.store,
+                    swap.swap_id,
+                    "loop in lock transaction rejected before restart".to_string(),
+                    now_ms(),
+                )?;
+                return Ok(true);
+            }
+        }
+        if swap.state == LiquiditySwapState::RefundPending {
+            return self.resume_refund_pending_swap(swap, myself).await;
```
jjyr, line 1819: [P1] Refund execution exists, but the actor never initiates it.
The actor can resume an order already in RefundPending, broadcast its refund, and process confirmation. However, production code in this file never transitions an order into RefundPending. For example, a Client Loop In order whose provider never accepts remains OnchainLocked even after refund maturity. An expiry/reconciliation mechanism must safely initiate the refund transition.

PR #1649, opened 2026-09-03
fix: do not block settlement on LocalAnnounced TLC when remote force-closes with preceding commitment

jjyr:
### Problem

A remote force-close can publish either the preceding unrevoked commitment or the pending commitment. An offered `LocalAnnounced` TLC may exist only in the pending commitment: waiting for it after the preceding commitment is published leaves the channel stuck in `WAITING_ONCHAIN_SETTLEMENT` ([original report](https://github.com/nervosnetwork/fiber/pull/1630#issuecomment-5351769708)).

The waitlist must also exclude received `AnnounceWaitPrevAck` TLCs that are absent from the remote commitment. Conversely, TLCs actually included in the published commitment must continue blocking until reconciled.

Comparing a rebuilt pending transaction hash with the on-chain hash is insufficient: broadcast-time FundingLock/UDT `cell_deps` change the raw transaction hash. Rebuilding a historical snapshot from current channel state is also unreliable because that state may have advanced.

### Implementation

- Share Watchtower's commitment snapshot selection with channel reconciliation. Parse the commitment lock, validate its script and funding-key identity, select the local/preceding remote/pending remote snapshot by direction and commitment number, then verify its settlement witness hash against the lock.
- Persist the verified snapshot as `ShutdownSettlementRecord`, including the closing transaction hash, direction, commitment number and settlement data. Reconciliation reads it directly from Store and validates its transaction hash and direction against the channel. There is no additional runtime snapshot field or hydration step.
- Check unresolved TLCs against the actual snapshot, translating TLC IDs into the local channel's direction. A snapshot TLC missing from local state remains unresolved. A force-close without a valid snapshot also remains waiting.
- Keep snapshot recovery scheduled even when `ONCHAIN_SETTLEMENT_CONFIRMED` is already set. A recovery event wakes reconciliation for either a live ChannelActor or the NetworkActor's offline path; recovery alone does not mark settlement complete.

### Record lifecycle and recovery

`ShutdownSettlementRecord` is created when a confirmed remote close is detected and its snapshot can be verified, or by the background shutdown check when a matching record is missing. The background path also covers local force-close and restart recovery.

The database key is the record prefix plus `channel_id`, so there is at most one record per channel. Retries or corrected snapshots overwrite that key; normal payments and commitment updates do not append records.

Once on-chain settlement is confirmed and TLC reconciliation and the required payer effects are complete, both live and offline paths persist the completed channel state before deleting the record. This ordering avoids leaving a waiting channel without its recovery snapshot after a crash between the writes. Channel state and TLC settlement proofs are retained.

The closing transaction's lock contains a settlement witness hash, not the complete TLC snapshot. Recovery therefore needs locally stored Watchtower data as well as the chain transaction. Persisting the verified record avoids repeated RPC/verification and dependence on retaining the candidate historical snapshot throughout reconciliation.

A record with a mismatched transaction hash or direction is ignored and recovery is retried; a successfully verified replacement overwrites it. Invalid scripts, funding identity/input mismatches, or witness-hash mismatches are rejected. Transient RPC failures can recover on later checks. If the required historical snapshot is permanently unavailable, retries cannot reconstruct it from its hash: the channel remains waiting until valid data becomes available rather than finalizing without evidence.

### Validation

- 368 related channel, network, on-chain TLC, Watchtower and Store tests passed with `cargo nextest`.
- Regression coverage includes preceding/pending commitments, a broadcast transaction with `CellDep`, received `AnnounceWaitPrevAck`, missing/wrong-direction records, RPC retry, live/offline actor finalization, tenant isolation and a real database close/reopen.
- Additional assertions verify that Store record updates, hash mismatches and deletion affect reconciliation without refreshing channel state.
- `make clippy` passed, including native, SQLite without default features, and WASM checks.
- `make check-migrate`, formatting and diff checks passed. The removed runtime field was not serialized; no persisted channel-format change is required.


jjyr:
@gpBlockchain

crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs
```diff
@@ -331,21 +332,91 @@ pub(crate) fn collect_onchain_received_timeout_settled_tlcs(
...
+    if local_uncooperative_close {
+        // On local force-close, check against the local commitment snapshot which omits uncommitted offered TLCs.
+        if let Ok(local_snapshot) = state.build_settlement_data(false) {
+            return has_unresolved_onchain_tlcs_for_snapshot(state, &local_snapshot, false);
+        }
+    } else if remote_uncooperative_close {
+        // Deterministically derive the pending remote commitment tx hash to determine whether
+        // the remote party published the pending commitment (containing LocalAnnounced TLCs)
+        // or the preceding one (omitting them).
+        let pending_commitment_hash = state
```
chenyukang: [P1] Compare a normalized commitment transaction rather than this pre-broadcast hash

build_commitment_tx_and_settlement_data(true) returns a transaction without cell_deps, while get_latest_commitment_transaction injects the FundingLock/UDT deps before broadcasting. The CKB transaction hash covers raw cell_deps, so shutdown_transaction_hash obtained from the chain will differ even when the peer published this pending commitment. That makes a real pending commitment look like the preceding commitment and can let its LocalAnnounced TLC stop blocking before reconciliation. The existing test assigns pending_tx.hash() before adding deps, so it misses the production path. Please compare normalized transactions (for example, clear cell_deps on both sides) or select the settlement snapshot by commitment number. A focused test that adds a broadcast-time CellDep to the rebuilt pending transaction fails at this head.

crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs
```diff
@@ -331,21 +332,91 @@ pub(crate) fn collect_onchain_received_timeout_settled_tlcs(
...
+            {
+                match (&state.shutdown_transaction_hash, &pending_commitment_hash) {
+                    // Confirmed tx matches pending commitment: TLC is on-chain, must wait for settlement.
+                    // Different hash (e.g. preceding commitment broadcasted): TLC was never on-chain, do not block.
+                    (Some(confirmed_hash), Some(pending_hash)) => confirmed_hash == pending_hash,
+                    // Unknown shutdown tx hash: fall back to conservative waiting.
+                    _ => true,
+                }
+            } else {
+                // All other active TLCs on remote force-close must be settled on-chain.
```
chenyukang: [P1] Do not wait for received AnnounceWaitPrevAck on a remote commitment

commitment_signed_tlcs(true) and build_settlement_data(true) explicitly exclude a received AnnounceWaitPrevAck TLC because that state only belongs to the local commitment. This remote-force-close branch treats every non-LocalAnnounced reconcilable TLC as present, while can_reconcile_onchain_fulfillment now includes AnnounceWaitPrevAck. The channel will therefore wait forever for an on-chain settlement record that cannot exist. Please filter every TLC through the actual selected remote settlement snapshot, rather than special-casing only offered LocalAnnounced. A focused regression test passes at the PR base and fails at this head.

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `f9232d52254a5aa52195ecae296c896de7078887..afc07c3d02e48b915825e5ab02e3681ae30751c9`
Review model: `grok-4.6`

Validation:
- Static source trace of the new remote-close waitlist, commitment builder, and broadcast path.

Notes:
- High: `has_unresolved_onchain_tlcs` identifies a pending remote force-close by comparing a rebuilt skeleton commitment `tx.hash()` with `shutdown_transaction_hash`. Broadcast commitments inject FundingLock cell_deps, and CKB transaction hashes include those deps, so the equality does not hold in production. Offered `LocalAnnounced` TLCs are therefore treated as off-chain even when the pending remote commitment was published. The same function also waits on inbound `AnnounceWaitPrevAck` TLCs, which are not included in the remote commitment and can re-stick `WAITING_ONCHAIN_SETTLEMENT`.
- Recommended fix: bind the remote-close waitlist to the published commitment snapshot (lock-args version / settlement hash), the same identity watchtower already uses, via `has_unresolved_onchain_tlcs_for_snapshot`. Add a regression that uses a cell_deps-bearing broadcast hash rather than the skeleton hash.
- Local force-close snapshot path and the panic-to-`Result` changes did not produce additional findings.
- fiber-scripts and fiber-sphinx revisions are unchanged by this diff.

gpBlockchain:
This PR fixes the later “don’t block `WAITING_ONCHAIN_SETTLEMENT` on a `LocalAnnounced` TLC that is not in the published preceding commitment” path. It does **not** cover the earlier watchtower reconstruction failure on a freshly opened channel when RAA is withheld.

## Repro

Python case: `test_stale_pre_tlc_commitment_leaves_localannounced_stuck`

1. Channel is `ChannelReady` (no RAA has been exchanged yet, so `revocation_data` is `None`).
2. Victim sends `AddTlc + CommitmentSigned`; attacker withholds `RevokeAndAck`.
3. Attacker broadcasts the **pre-TLC** commitment (lock args commitment number `1`), not the latest TLC-containing one.
4. Victim watchtower `PeriodicCheck` then loops on:

```text
WARN  fnn::fiber::onchain_tlc_reconcile: Settlement snapshot hash does not match commitment lock for channel Hash256(...), commitment 1
ERROR fnn::watchtower::actor: Cannot reconstruct settlement TLC identities for channel Hash256(...); skipping on-chain TLC reconciliation and settlement construction
```

The commitment cell is never spent. The error repeats every watchtower interval.

## Why this PR misses it

`settlement_data_for_commitment()` only selects `remote_settlement_data` when `revocation.commitment_number == commitment_number - 1`. With `revocation_data == None` it **always** returns `pending_remote_settlement_data`.

On this attack:

- `pending_remote_settlement_data` was overwritten by the victim’s later `LocalCommitmentSigned` and **includes the LocalAnnounced TLC**.
- The on-chain tx is still the funding-era remote commitment **without** that TLC.
- `verify_and_select_settlement_data()` hashes the heuristic pick, fails, and returns `None`.
- Watchtower then skips settlement entirely.

The matching snapshot is still stored in `remote_settlement_data` (the CreateChannel / pre-TLC copy). The selector never tries it.

Existing unit coverage (`revocation.commitment_number == 10` → preceding 11 / pending 12) assumes a revocation already exists. The withheld-RAA fresh-channel case is untested.

## Suggested fix

In `verify_and_select_settlement_data()`, if the heuristic snapshot’s witness hash does not match the lock, try the other remote snapshot before refusing. `build_settlement_tx` should use that verified snapshot instead of calling the heuristic again.

Until that lands, this PR can still leave the original report stuck: snapshot verification rejects the close, so the “LocalAnnounced must not block preceding close” logic never runs.

gpBlockchain:
Independent security review of `f9232d52254a5aa52195ecae296c896de7078887..7e6f9bc7d89eba3c536b869738a2b51cb9024a63` (review model: **gpt-6-astra**, Codex desktop). Earlier reports were not consulted.

**[P2] Preserve shutdown snapshot recovery for standalone-watchtower deployments.**

The new unconditional missing-snapshot wait at [onchain_tlc_reconcile.rs:590–595](https://github.com/nervosnetwork/fiber/blob/7e6f9bc7d89eba3c536b869738a2b51cb9024a63/crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs#L590-L595) depends on `get_local_watch_channel`. Both snapshot writers require that local record. However, the supported configuration `disable_built_in_watchtower=true` with `standalone_watchtower_rpc_url` omits the local watcher and forwards events only remotely. The WASM/no-watchtower getter also returns `None`.

Even a never-used, zero-TLC channel therefore remains in `WAITING_ONCHAIN_SETTLEMENT` after its on-chain settlement completes. Live/offline cleanup stays blocked, retaining channel lifecycle resources and retries after restart.

Persist the verified immutable snapshot independently of optional built-in watcher execution, or add verified retrieval usable by standalone/WASM deployments; retain conservative handling for genuinely missing/invalid history.

Validation: exact base/head guard functions compiled with minimal typed stubs. For both close directions, base/no snapshot resolves; head/no snapshot waits; head/valid empty snapshot resolves. This is not a full-node/chain integration test. No theft or actual on-chain funds lock is asserted. Broad existing suites were not run; CI is expected to cover them.

gpBlockchain:
lgtm

PR #1656, opened 2026-09-11
fix: prevent prefix-only on-chain preimage settlement stranding

quake:
Summary: negotiate and persist the ONCHAIN_FULL_PAYMENT_HASH commitment feature per channel; support legacy and v1 settlement witness layouts in channel builders and watchtower parsing; treat exact prefix-valid/full-hash-invalid on-chain preimages as failed settlements and fail forward immediately; add settlement-size, watchtower, reconciliation, and compatibility regression coverage. Depends on the matching fiber-scripts draft PR for the upgraded commitment-lock contract. Verification: focused reconciliation tests 25/25 passed; watchtower tests 28 passed; make check passed; make check-migrate passed; clippy and formatting passed. Full nextest has 29 environment failures: Cannot assign requested address.


crates/fiber-types/src/channel.rs
```diff
@@ -1615,6 +1675,11 @@ pub struct ChannelActorData {
     /// Persisted state for an in-progress external funding flow.
     #[serde(default)]
     pub external_funding: Option<ExternalFundingPersistState>,
+
+    /// Which commitment-lock settlement witness layout this channel uses,
+    /// decided once at channel-open and never changed.
+    #[serde(default)]
+    pub commitment_contract_version: CommitmentContractVersion,
```
chenyukang, line 1682: I think this need a migration?

crates/fiber-lib/src/fiber/channel.rs
```diff
@@ -9851,19 +9898,25 @@ impl ChannelActorState {
...
                 &settlement_data,
                 for_remote,
+                self.commitment_contract_version,
                 local_settlement_key,
                 remote_settlement_key,
             ))
             .as_ref(),
         );
         commitment_lock_script_args.push(0x00);
+        if self.commitment_contract_version == CommitmentContractVersion::V1 {
```
chenyukang, line 9939: V1 requires 99 CKB, but the default reserve remains 99 CKB including fees. We should change to 100 CKB as reserve for V1?

see the function of occupied_capacity, seems we need to add different branch based on version.

crates/fiber-lib/src/watchtower/actor.rs
```diff
@@ -765,11 +779,96 @@ fn try_settle_commitment_tx<S: WatchtowerStore>(
...
+    channel_data: &ChannelData,
+    for_remote: bool,
+) -> Option<Vec<TrackedSettlementTlc>> {
+    let lock_args = commitment_lock.args().raw_data();
+    if lock_args.len() < 56 {
+        return None;
+    }
+    let commitment_number = u64::from_be_bytes(lock_args[28..36].try_into().ok()?);
+    let settlement_data =
+        settlement_data_for_commitment(channel_data, for_remote, commitment_number);
```
chenyukang: **[P1] Reuse the verified snapshot when reconstructing tracked TLCs**

This local helper replaces the imported `onchain_tlc_reconcile::tracked_settlement_tlcs`, but changes snapshot selection from matching the on-chain witness hash to a revocation-number heuristic. Before the first `RevokeAndAck`, `revocation_data` is `None` while `remote_settlement_data` can still be the initial snapshot S0 and `pending_remote_settlement_data` can already contain the first TLC in S1. If the peer publishes its still-valid S0 commitment, `try_settle_commitment_tx` correctly selects S0 with `verify_and_select_settlement_data`, but this helper picks S1, fails the hash check, and returns `None`. The caller then skips all reconciliation and settlement construction, including withdrawal of the balances in S0 even though it has no TLCs. This affects both Legacy and V1 channels. Please pass the already-verified `settlement_data` into this helper and only convert its TLCs, preserving an empty list as a valid result. A regression test should exercise snapshot selection and TLC reconstruction with `revocation_data = None`, S0 committed on-chain, and a different pending S1, rather than bypassing reconstruction by supplying an empty TLC list.

chenyukang: https://github.com/nervosnetwork/fiber/blob/2e6ce63a2698a1078f56b8e6b3cb8bf7e8454da0/crates/fiber-lib/src/fiber/onchain_tlc_reconcile.rs#L206-L241

do we need to remove this?

jjyr (commented):
Please remove .superpowers

